PR #3, opened 2023-07-11
[Feature] Local development Laravel Sail setup

Orrison:
Sets up the application with a basic Laravel Sail configuration and some light documentation for it. It can be expanded upon later, but it gives us a starting point to start working with efficiency locally.


joelicatajr (approved):
Looks good, thanks.

PR #4, opened 2023-07-11
[Feature] PHP CS Fixer Setup

Orrison:
Installs PHP CS Fixer and runs it across the codebase to style the repository based on the configuration.

I would recommend filtering to look at the first commit's changes first, as that will contain the PHP CS Fixer setup. The other file changes are the fixes across the entire codebase.

We should wait to merge until approval from all current team members, @dgoetzit and @joelicatajr, to ensure we are in agreement on the ruleset configuration.
Any ruleset addition or change can be discussed here if needed. Then included, if agreed upon, and ran again before merging.

All available rules can be found [here](https://cs.symfony.com/doc/rules/index.html).

As the team grows, we should include an open space for discussion of changes to the ruleset and planning for change incorporation. Possibly in retros or similar meetings.
All in an effort to ensure a baseline syntax is consistent across the code base and "bikeshedding" discussion is reduced in peer review.


joelicatajr (approved):
This makes sense, thanks @Orrison. I will hold off merging until @dgoetzit gets a chance to review.

dgoetzit (approved):

PR #7, opened 2023-07-12
[Feature] Prettier installation and setup

Orrison:
Sets up Prettier for CSS, JS, and Blade linting.

It may be best to filter out b55227c8331c78aa6eeacf440d35c3d53c9a03a1 to see the changes that are not related to me running it across the codebase.

Best to merge in after full team approval to mark agreement on the configuration.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,4 @@
+resources/views/vendor/mail/**/*.blade.php
+resources/views/mail/**/*.blade.php
+resources/lang/**/*.php
+**/vendor
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ yarn-error.log
 /.vscode
 public/css/
 public/js/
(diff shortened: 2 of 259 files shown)


.prettierignore
```diff
@@ -0,0 +1,4 @@
+resources/views/vendor/mail/**/*.blade.php
+resources/views/mail/**/*.blade.php
```
dgoetzit, line 2: Are we doing this preventatively in the event that we create markdown mail?

dgoetzit (approved):
Looks good! 1 non-blocking question.

.prettierignore
Orrison: Yeah, exactly, it causes a lot of issues before so just getting out ahead of it just in case it is ever added.

joelicatajr: This makes sense.

joelicatajr (approved):

PR #13, opened 2023-07-14
[Feature] Roles and Permissions

dgoetzit:
This PR introduces Roles and Permissions, facilitated in the app via the Spatie/Laravel-permission package. I've added some documentation to cover the high level convention, and there is some more work to be done here regarding:

1. Role Groups
2. One-off processes to migrate new roles/permissions without impacting existing data in an app
3. Determining all of the necessary permissions per model and custom permissions currently needed in the application


app/Models/Permission.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace App\Models;
+
+use App\Models\Concerns\DefinesPermissions;
+use Spatie\Permission\Models\Permission as SpatiePermission;
+
+class Permission extends SpatiePermission
+{
+    use DefinesPermissions;
```
Orrison, line 10: Is this perhaps one that should be overwritten? Since we would really only ever want the `index` and `show` permission for this?

app/Models/Concerns/DefinesPermissions.php
Orrison, line 1: How do you feel about `HasPermissions`? That seems to be the common convention for a lot of these traits.

config/permissions/api/custom.php
```diff
@@ -0,0 +1,65 @@
...
+    'permission.create',
+    'permission.edit',
+    'permission.show',
+    'permission.delete',
+    'permission.access',
+    'role.create',
+    'role.edit',
+    'role.show',
+    'role.delete',
+    'role.access',
```
Orrison, line 19: Curious if you think these could be deleted now, as they should be covered by the Model trait?

config/permissions/web/custom.php
```diff
@@ -0,0 +1,65 @@
...
+    'permission.create',
+    'permission.edit',
+    'permission.show',
+    'permission.delete',
+    'permission.access',
+    'role.create',
+    'role.edit',
+    'role.show',
+    'role.delete',
+    'role.access',
```
Orrison, line 19: Same questions as the `api` one.

config/roles/api/super_admin.php
Orrison, line 1: We may have talked about this already. Sorry if so.

Do we need this since the Super Admin's access in implicit?

docs/roles-and-permissions.md
```diff
@@ -0,0 +1,81 @@
+# Roles and Permissions
+Roles and permissions in the application have been setup in a flexible and maintainable manner, making it easy for application owners to make adjustments to configuration, and ensure that the custom roles and permissions are in place.
+
+This application uses the [Spatie Laravel-permission](https://spatie.be/docs/laravel-permission/v5/introduction) package in order to facilitate the usage of roles and permissions.
+
+The application uses a dedicated convention in order to define and poulate roles and permissions, leaving little in the way of thinking when it comes time to stand up the application locally or in a live environment.
```
Orrison, line 6:
```suggestion
The application uses a dedicated convention in order to define and populate roles and permissions, leaving little in the way of thinking when it comes time to stand up the application locally or in a live environment.
```

Orrison (changes requested):
Looks great! I just had a few questions. There are also some merge conflicts.

app/Models/Concerns/DefinesPermissions.php
dgoetzit: I only didn't do this as Spatie also has a `HasPermissions` trait, and didn't want to cause any confusion.

config/roles/api/super_admin.php
dgoetzit: We could do without it, but right now the convention in place will create a role for every configuration file that exists in the `roles` directory. So it's not necessary, but we'd need to make a minor adjustment to the `CreateRoles` action.

Orrison (approved):

PR #16, opened 2023-07-17
[Feature] Roles and Permissions Resources

dgoetzit:
This PR introduces simple resources for Roles and Permissions. 

Highlights:
- Doctrine/dbal install for table and form generation capability
- Roles and Permissions resources
- Roles and Permission policies
- New seeded users with roles assigned

This PR will be followed by changes to introduce `RoleGroup`, which will alter how Roles are assigned to users and ultimately managed in the application.


app/Models/Role.php
```diff
@@ -19,4 +30,14 @@ public function scopeWeb(Builder $query): void
...
+
+    public function scopeSuperAdmin(Builder $query): void
+    {
+        $query->where('name', 'super_admin');
+    }
+
+    public function scopeAdmin(Builder $query): void
+    {
+        $query->where('name', 'admin');
+    }
```
Orrison, line 42: Because roles are mutable, can we rely on these scopes?

I'll say, perhaps we make the Super Admin real immutable and required. But they **_could_** delete the Admin role, right?

app/Policies/RolePolicy.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Determine whether the user can view the model.
+     */
+    public function view(User $user, Role $role): bool
+    {
+        return $user->hasPermissionTo('role.*.view');
+    }
+
+    /**
```
Orrison, line 26: Are we setting up the create, edit, delete, etc. logic for roles later?

database/seeders/UsersTableSeeder.php
```diff
@@ -2,25 +2,34 @@
...
-                'emplid' => '',
-            ],
-        ];
-
-        User::insert($users);
+        /** Super Admin */
+        $superAdmin = User::factory()->create([
+            'name' => 'Super Admin',
+            'email' => 'superadmin@assist.com',
+            'password' => bcrypt('password'),
```
Orrison, line 17: We should use the `Hash::make()` helper here rather than calling `bcrypt`

Orrison (commented):
Looking great! I just had some questions.

app/Models/Role.php
dgoetzit: The plan is to only use these scopes in initial/local setup steps.

app/Filament/Resources/PermissionResource/Pages/CreatePermission.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace App\Filament\Resources\PermissionResource\Pages;
+
+use Filament\Resources\Pages\CreateRecord;
+use App\Filament\Resources\PermissionResource;
+
+class CreatePermission extends CreateRecord
```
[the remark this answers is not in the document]
dgoetzit: I've deleted the Create/Edit pages for now. There are currently no actions corresponding to any create/edit/delete for permissions, and all management of permissions (attaching) will happen through roles.

PR #17, opened 2023-07-18
[Feature] Assist Data Model Redshift Database POC

Orrison:
This sets up a POC for what it could look like to connect the Assist Data Model Models to the Redshift instance directly.

Exactly as Justin mentioned, everything works just fine. But there are issues when you attempt to use any of the `has` (`whereHas`, `whereDoesntHave`, etc.) functionality.
This is due to Eloquent creating it all as a single SQL query, which fails as the "RedShift" DB is in an entirely separate server. Many Laravel Core members (Mohammad and Taylor) have said that Laravel does not currently support this.

We don't "need" this functionality. But we would be losing functionality if we went with this choice.

There is the possibility that we could consider creating a package and/or core contribution to enable this functionality. (it should be possible to do and would be a great contribution)
But we can merge this in for now and discuss whether or not we want to go with this solution or a 3rd party sync to the local table.


joelicatajr (approved):
Changes look good, and it is great that this strategy checks out. The more I think about this, it solves for many of the challenges I have been trying to work around with how Colleges manage their data.

PR #18, opened 2023-07-21
[Feature] Postgres Foreign Data Wrapper POC

Orrison:
This is a new proposal for bringing the ASSIST Data Model Models within the CRM.

By using Postgres as the database of choice for the CRM itself, we can make use of a functionality known as [Foreign Data Wrappers (FDW)](https://wiki.postgresql.org/wiki/Foreign_data_wrappers).
FDW allows us to treat a table in a different database on a completely different server as if it was a local table.

This allows us to use all of the native Laravel/Eloquent functionality without much compromise other than needing to use Postgres as the CRM local database.
Postgres is a supported DB type for Laravel that works very much the same. It often can, in fact, work much better/faster. FDW is performant out of the box, with options to improve performance that we can look into along the way.

We currently believe that this is a better alternative to our original multi-connection proposal and may even win out against external DB syncing with how well it seems to be working initially.

One of the more interesting things about this route is the Postgres FDW has support for many external databases. We should be able to connect to Redshift to bring those views in.
But if at any time we, or an OSS User, don't want to/can't use Redshift, FDW can connect directly to their SIS database of choice. (logistics of that to be worked out at a different time, as it is required that the schema of your external table be known to the local and that their database be within the FDW-supported drivers listed in the link above)

This works well and is being tested by "replicating" an environment we expect. Because Redshift is based on Postgres, we simulate the "Redshift" instance by loading in a second Postgres server. Since most of what we are doing is basic queries, this can be mostly trusted as a mocked environment. But if this was decided as the option to go with, we may want to consider something like [LocalStack](https://localstack.cloud/) or the automatic spinning up of a Redshift cluster for CI/CD.


.github/workflows/ci.yml
```diff
@@ -98,7 +131,7 @@ jobs:
         run: php artisan route:list
 
       - name: Execute tests (Unit and Feature tests) via PHPUnit
-        run: php artisan test --parallel --configuration phpunit-ci.xml --testsuite=Feature,Unit
```
dgoetzit, line 101: Did we lose the ability to run tests in `parallel` with this configuration?

app/Models/Enrollment.php
```diff
@@ -0,0 +1,16 @@
...
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+/**
+ * @mixin IdeHelperEnrollment
+ */
+class Enrollment extends Model
+{
+    use HasFactory;
+
+    protected $connection = 'sis';
```
dgoetzit, line 15:
```suggestion
```

app/Models/Program.php
```diff
@@ -0,0 +1,16 @@
...
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+/**
+ * @mixin IdeHelperProgram
+ */
+class Program extends Model
+{
+    use HasFactory;
+
+    protected $connection = 'sis';
```
dgoetzit, line 15:
```suggestion
```

dgoetzit (approved):
This looks like a great solution for what we need to accomplish, without sacrificing any core Laravel functionality or increasing complexity on that end of things.

I just had a single question regarding runnings tests, but beyond that think we can move forward with this approach.

.github/workflows/ci.yml
Orrison: For now, yes, I am sure we can do it. But it gave me enough trouble to warrant not handling it in this POC.

joelicatajr (approved):
I really like this solution, nice work. I think we will see some performance improvements too going with AWS Aurora PostgreSQL Serverless on our back-end.

PR #19, opened 2023-07-25
[Feature] Role Groups

dgoetzit:
This PR is still a WIP, but some of the following remaining pieces may be extracted into their own PRs to prevent scope creep as much as possible.

Still remaining to do:
- [ ] Finish writing documentation on RoleGroups, Roles, Permissions, and how they exist within the system and interact with Users and each other
- [ ] Perform testing of all RoleGroup related action logic to ensure that functionality in Filament will work as expected when attaching/removing Roles/Users/RoleGroups
- [ ] Finish segmenting initial Roles/Permissions based on existing models
- [ ] Determine why we do not have access to the morphable type on the `RoleGroupPivot` deleted event, and figure out if we need to implement a workaround (simply don't make it a Polymorphic relationship and introduce pivots for `role_role_group` and `role_group_user`
- [ ] Add second dialog when removing a User from a RoleGroup to determine if they should keep all of the rest of their Roles from the RoleGroup "directly"
- [ ] Other small remaining cleanup items like simplifying existing code, extracting certain pieces of logic, finalize naming conventions, queueing processes that should run in the background, etc...


app/Actions/RolesAndPermissions/SyncRoleGroups.php
```diff
@@ -0,0 +1,33 @@
...
+use Illuminate\Support\Facades\File;
+
+class SyncRoleGroups
+{
+    public function handle(): void
+    {
+        collect(File::files(config_path('role_groups')))->each(function ($file) {
+            $config = config("role_groups.{$file->getFilenameWithoutExtension()}");
+
+            $roleGroup = RoleGroup::create([
```
Orrison, line 16: Should this be `firstOrCreate`?

app/Events/RoleGroupPivotDeleted.php
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Get the channels the event should broadcast on.
+     */
+    public function broadcastOn(): array
+    {
+        return [
+            new PrivateChannel('channel-name'),
+        ];
+    }
```
Orrison, line 33: Feel free to ignore cleanup tasks if your intention is to do so later. But I believe this can be removed if we are not working with broadcasting.

app/Events/RoleGroupPivotSaved.php
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Get the channels the event should broadcast on.
+     */
+    public function broadcastOn(): array
+    {
+        return [
+            new PrivateChannel('channel-name'),
+        ];
+    }
```
Orrison, line 33: Same as above

app/Filament/Resources/RoleGroupResource/RelationManagers/PermissionsRelationManager.php
```diff
@@ -0,0 +1,53 @@
...
+    public function form(Form $form): Form
+    {
+        return $form
+            ->schema([
+                Forms\Components\TextInput::make('name')
+                    ->required()
+                    ->maxLength(255),
+                Forms\Components\TextInput::make('guard_name')
+                    ->required()
+                    ->maxLength(255),
```
Orrison, line 26: This could be a select with web and API as the only options

app/Filament/Resources/RoleGroupResource/RelationManagers/RolesRelationManager.php
```diff
@@ -0,0 +1,55 @@
...
+    public function form(Form $form): Form
+    {
+        return $form
+            ->schema([
+                Forms\Components\TextInput::make('name')
+                    ->required()
+                    ->maxLength(255),
+                Forms\Components\TextInput::make('guard_name')
+                    ->required()
+                    ->maxLength(255),
```
Orrison, line 27: This could be a select with web and API as the only options

app/Models/Concerns/HasSlug.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace App\Models\Concerns;
+
+use Illuminate\Support\Str;
+use Illuminate\Database\Eloquent\Model;
+
+trait HasSlug
```
Orrison, line 8: Do we want to consider something like https://github.com/spatie/laravel-sluggable. Which has built-in support for slug collisions?

Orrison (approved):
This is looking great! Just a few questions. Feel free to merge once reviewed.

Let's just make sure we keep those remaining tasks somewhere to be adding into the project management software once we have that setup.

app/Models/Concerns/HasSlug.php
dgoetzit: Will pull in on separate PR.

app/Events/RoleGroupPivotDeleted.php
dgoetzit: Will be removed upon cleanup.

Orrison (changes requested):
Looks like tests are failing now.

Orrison (approved):

PR #23, opened 2023-07-26
[Feature] Initial Setup of Student Resource and Re-Ordering of Modules

Orrison:
Per our discussion earlier today @dgoetzit, I have rearranged the modules to fit our settled pattern.

I also setup the direct Model morph mapping for both the case and adm modules.

And finally created the resource for Student to continue work on it within future PRs.


app-modules/assist-data-model/routes/assist-data-model-module-routes.php
dgoetzit, line 1: Not a big deal but we should remove the `module` from the name here.

composer.json
```diff
@@ -12,7 +12,8 @@
     "require": {
         "php": "^8.2",
         "ext-pdo": "*",
-        "canyon-gbs/case-module": "*",
+        "canyon-gbs/assist-data-model": "^1.0",
+        "canyon-gbs/case": "^1.0",
```
dgoetzit, line 16: Should these be versioned at this point?

dgoetzit (approved):
Looks good. Just a couple of non-blocking questions/comments.

composer.json
Orrison: Ah no, that was a mistake with something I was doing.

dgoetzit (approved):

PR #29, opened 2023-07-31
[Feature] Finish adding Case Update Model and Resource

Orrison:


app-modules/case/database/factories/CaseUpdateFactory.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+class CaseUpdateFactory extends Factory
+{
+    public function definition(): array
+    {
+        return [
+            'case_id' => CaseItem::factory(),
+            'update' => $this->faker->sentence(),
+            'internal' => $this->faker->boolean(),
+            'direction' => $this->faker->randomElement(CaseUpdateDirection::cases())->value,
```
dgoetzit, line 21: Does there need to be any correlation between `internal` and `direction`? I might be a bit confused on the terminology, but would an `internal` update also always be "inbound"?

app-modules/case/database/migrations/2023_03_06_000029_create_case_update_items_table.php
```diff
@@ -6,15 +6,18 @@
...
-            $table->string('update');
-            $table->string('internal');
+            $table->unsignedBigInteger('case_id')->nullable();
+            $table->text('update');
+            $table->boolean('internal');
             $table->string('direction');
             $table->timestamps();
             $table->softDeletes();
+
+            $table->foreign('case_id', 'case_fk_8136752')->references('id')->on('case_items');
```
dgoetzit, line 20: Do we need to specify the foreign key name here? Can we just do the following and let the index be generated for us?

```suggestion
            $table->foreignId('case_id')->references('id')->on('case_items');
```

composer.json
```diff
@@ -15,7 +15,7 @@
     "canyon-gbs/authorization": "*",
     "canyon-gbs/assist-data-model": "*",
     "canyon-gbs/case": "*",
-    "filament/filament": "v3.0.0-beta11",
+    "filament/filament": "^3.0.0",
```
dgoetzit, line 18: 🎉

dgoetzit (commented):
This looks great - I love the Infolist usage. Just a couple of questions.

app-modules/case/database/factories/CaseUpdateFactory.php
Orrison: That's a good point. Theoretically, I guess `internal` is neither inbound nor outbound. But inbound does make the most sense.

composer.json
Orrison: I believe I may have done this already in a recent commit I pushed up, I'll check though.

dgoetzit (approved):

PR #41, opened 2023-08-15
[Feature] Init auditing functionality

Orrison:


app-modules/case/database/migrations/2023_03_06_000008_create_case_items_table.php
```diff
@@ -6,21 +6,21 @@
...
             $table->string('respondent_id')->nullable();
             $table->longText('close_details')->nullable();
             $table->longText('res_details')->nullable();
-            $table->unsignedBigInteger('institution_id')->nullable();
-            $table->unsignedBigInteger('status_id')->nullable();
-            $table->unsignedBigInteger('type_id')->nullable();
-            $table->unsignedBigInteger('priority_id')->nullable();
-            $table->unsignedBigInteger('assigned_to_id')->nullable();
-            $table->unsignedBigInteger('created_by_id')->nullable();
+            $table->uuid('institution_id')->nullable();
```
dgoetzit, line 18: We can probably do this in another PR, but makes sense to use `foreignUuid()` here and remove the extra migration to define the foreign key constraints. Will just need to re-order migrations.

app-modules/case/src/Models/CaseItem.php
```diff
@@ -71,10 +75,12 @@
  *
  * @mixin Eloquent
  */
-class CaseItem extends BaseModel
+class CaseItem extends BaseModel implements Auditable
 {
     use SoftDeletes;
     use PowerJoins;
+    use AuditableTrait;
+    use HasUuids;
```
dgoetzit, line 83: We should probably just add the `HasUuids` trait to the `BaseModel`.

composer.json
```diff
@@ -30,6 +31,7 @@
         "laravel/ui": "^4.2",
         "league/flysystem-aws-s3-v3": "^3.0",
         "livewire/livewire": "^3.0@beta",
+        "owen-it/laravel-auditing": "^13.5",
```
dgoetzit, line 34: Should this be installed on the module instead of the root?

app-modules/prospect/src/Models/ProspectSource.php
```diff
@@ -38,10 +42,11 @@
  *
  * @mixin Eloquent
  */
-class ProspectSource extends BaseModel
+class ProspectSource extends BaseModel implements Auditable
 {
     use HasUuids;
     use SoftDeletes;
+    use AuditableTrait;
```
dgoetzit, line 49: We should probably add this trait to the `BaseModel` as well.

dgoetzit (approved):
This looks good. I left a few comments/suggestions/questions, but feel free to merge as they are non-blocking and can be addressed in a later PR.

app-modules/prospect/src/Models/ProspectSource.php
Orrison: I don't think we can at this point. Because the ADM Models extend the `BaseModel` and they don't have this trait

app-modules/case/database/migrations/2023_03_06_000008_create_case_items_table.php
Orrison: Yeah I can rearrange these at a later PR

composer.json
Orrison: Hmm, that should probably work and probably be how it works. Let me try

dgoetzit (approved):

PR #42, opened 2023-08-15
[Feature] Implement initial notifications

dgoetzit:


app-modules/notifications/composer.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "canyon-gbs/notifications",
+  "description": "",
+  "type": "library",
+  "version": "1.0",
+  "license": "proprietary",
+  "require": {
+    "filament/filament": "^3.0.0"
```
Orrison, line 8: Per Dan's recommendation, let's change this to `^3.0-stable` for now. I have done so for the other Modules in a branch I will submit a PR for today.

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -67,7 +67,9 @@ public function panel(Panel $panel): Panel
             ])
             ->authMiddleware([
                 Authenticate::class,
-            ]);
+            ])
+            ->databaseNotifications()
+            ->databaseNotificationsPolling('10s');
```
Orrison, line 72: We should probably move this to do this configuration within the `notifications` module. See how we add `->discoverResources` in the other Modules like `case`.

We can do this in a different PR if you would prefer though.

Orrison (changes requested):

Orrison (approved):

PR #43, opened 2023-08-17
[Feature] Knowledge Base HTML Output

dgoetzit:
This PR adds the ability to render HTML via a `ViewEntry` in our Filament infolists. This uses a Blade view under the hood which still provides the field label and provides some degree of separation from other fields.

Screenshot 2023-08-16 at 10 52 44 PM


Orrison (changes requested):
Amazing! One minor thing, could we center the prose element or the container? Or perhaps make it the full width of the container? On larger screens the content on the view page ends up looking sort of left-aligned since the element doesn't fill the entire container.

Screenshot 2023-08-17 at 9 01 33 AM

Orrison (approved):

PR #44, opened 2023-08-17
[Feature] Audit enrollment and configuration

Orrison:
Setup for Audit Enrollment and additional configuration.

Now all relationship changes are audited by default without us having to call special helpers.

Sets up CRUD minus create, update, and delete for Audit records.


app-modules/audit/src/Providers/AuditServiceProvider.php
```diff
@@ -11,12 +14,19 @@ class AuditServiceProvider extends ServiceProvider
 {
     public function register()
     {
+        Panel::configureUsing(fn (Panel $panel) => $panel->plugin(new AuditPlugin()));
+
         app('config')->set('audit', require base_path('app-modules/audit/config/audit.php'));
+        app('config')->set('settings', require base_path('app-modules/audit/config/settings.php'));
```
dgoetzit, line 20: We might eventually run into some complications here, as I would think some other modules will potentially need to use Settings as well.

app-modules/audit/tests/AuditTrainUsageTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
dgoetzit, line 1: Small typo in file name - "Trait"

app-modules/audit/src/Filament/Actions/AuditAttachAction.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Assist\Audit\Filament\Actions;
+
+use OwenIt\Auditing\Contracts\Auditable;
+use Filament\Tables\Actions\AttachAction;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
+class AuditAttachAction extends AttachAction
+{
+    protected function setUp(): void
```
dgoetzit, line 11: Are we planning to use the custom Attach and Detach actions? Or is this covered by the other work done in setting this up behind the scenes?

dgoetzit (changes requested):
This looks awesome! Just a very minor file name fix and a couple of questions.

app-modules/audit/src/Providers/AuditServiceProvider.php
Orrison: Yeah, my plan was to possibly move this to a general Module at some point soon. Do you think anything would prevent us from moving it later?

app-modules/audit/src/Filament/Actions/AuditAttachAction.php
Orrison: This was covered by the behind-the-scenes work. I just decide to keep them around for a little bit just in case we end up needing them. As this behind-the-scenes work is still a little green feeling to me.

app-modules/audit/database/migrations/2022_12_14_083707_create_settings_table.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('settings', function (Blueprint $table): void {
+            $table->id();
```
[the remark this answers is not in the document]
Orrison: Changed to use UUIDs

app-modules/audit/src/Providers/AuditServiceProvider.php
dgoetzit: I don't think so - just wanted to keep it on our radar as it may benefit us to pivot to the other handling we looked into (hierarchical merging) for this case.

dgoetzit (approved):

PR #45, opened 2023-08-17
[Feature] Audit Purging

Orrison:
Add automated Audit purging based on admin configurable retention period

### DIFF
--- a/app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
+++ b/app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
@@ -6,7 +6,7 @@ use Spatie\LaravelSettings\Migrations\SettingsMigration;
 return new class () extends SettingsMigration {
     public function up(): void
     {
-        $this->migrator->add('audit.retention_duration', 90);
+        $this->migrator->add('audit.retention_duration_in_days', 90);
         $this->migrator->add('audit.audited_models', AuditableModels::all()->keys());
     }
 };
--- a/app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
+++ b/app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
@@ -6,7 +6,7 @@ use Spatie\LaravelSettings\Migrations\SettingsMigration;
 return new class () extends SettingsMigration {
     public function up(): void
     {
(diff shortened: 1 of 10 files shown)


app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
```diff
@@ -6,7 +6,7 @@
 return new class () extends SettingsMigration {
     public function up(): void
     {
-        $this->migrator->add('audit.retention_duration', 129600);
+        $this->migrator->add('audit.retention_duration', 90);
```
dgoetzit, line 9: I'm wondering if we also need to define a unit, or make this setting more explicit in the code that it's in days.

app-modules/engagement/src/Models/EngagementFile.php
```diff
@@ -73,7 +73,7 @@ public function students(): MorphToMany
             table: 'engagement_file_entities',
             foreignPivotKey: 'engagement_file_id',
             relatedPivotKey: 'entity_id',
-            relation: 'student',
+            relation: 'engagementFiles',
```
dgoetzit, line 76: I think this should remain `student`

dgoetzit (changes requested):
This looks good - just have a couple of suggestions.

app-modules/engagement/src/Models/EngagementFile.php
Orrison: It seems to break when I set it as `student`

app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
Orrison: Hmm, like perhaps a comment within the settings class for this?

app-modules/engagement/src/Models/EngagementFile.php
dgoetzit: That's odd 🤔

Orrison: I'll continue to look into it more. But this change has it working now

app-modules/audit/database/migrations/2023_08_16_025823_create_audit_settings.php
Orrison: Fixed, went with option #1 for now

PR #46, opened 2023-08-21
[Feature] Socialite setup and Azure Provider Setup

Orrison:
Sets up Socialite and the Azure provider for Socialite.

If the correct `.env` credentials are set, a "Login with Azure SSO" button should now be on the login page.

This will allow a User to authenticate with Azure and get access to the application. Currently, it sets no roles or anything. Logout will also detect whether or not the User logged in with Socialite. If they did, we will log them out via the SSO URL.
We may give the User an option to log out of just the app or the SSO provider entirely (like it does not) later on.


dgoetzit (approved):
Looks great!

PR #49, opened 2023-08-23
[Feature] Engagement CRUD

dgoetzit:
This PR introduces the basic CRUD needed for Engagements, in addition to:

- A dead simple notification for Engagements that Engagement data is injected into
- A job that will pick up undelivered Engagements every minute and deliver them via the correct channel (email, sms)

There is still more testing necessary here, as well as some other small touch ups and improvements, but most of that will be done in a separate PR when introducing the ability to send texts to Students/Prospects.

I decided to remove the `BaseModel` extension from Student, Enrollment, and Program as they are the lone exceptions to our UUID paradigm. Let me know if you want to discuss this further.


app-modules/engagement/src/Console/Commands/DeliverEngagements.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Assist\Engagement\Console\Commands;
+
+use Illuminate\Console\Command;
+use Assist\Engagement\Actions\DeliverEngagements as DeliverEngagementsJob;
+
+class DeliverEngagements extends Command
```
dgoetzit, line 8: This command is mostly just useful for local testing.

app-modules/engagement/src/Actions/EngagementEmailChannelDelivery.php
```diff
@@ -0,0 +1,24 @@
...
+            ->deliverable
+            ->engagement
+            ->recipient
+            ->notify(new EngagementNotification($this->deliverable));
+
+        // TODO We need to ensure that the notification was successfully delivered
+        // Before marking this as such
+        // This can probably be extracted into the QueuedEngagementDelivery,
+        // as we can figure out how to return the result of the `deliver()`
+        // Method and process the results there...
```
Orrison, line 21: Perhaps we can hook into https://laravel.com/docs/10.x/notifications#notification-events

app-modules/engagement/src/Models/EngagementDeliverable.php
```diff
@@ -0,0 +1,96 @@
...
+        ]);
+    }
+
+    public function deliver(): void
+    {
+        match ($this->channel) {
+            EngagementDeliveryMethod::EMAIL => EngagementEmailChannelDelivery::dispatch($this),
+            EngagementDeliveryMethod::SMS => EngagementSmsChannelDelivery::dispatch($this),
+            default => throw new UnknownDeliveryMethodException("Delivery channel '{$this->channel}' is not supported."),
+        };
```
Orrison, line 94: It may be nice to move this into the channel ENUM itself. That way, it can be like `$this->channel->deliver()`

Orrison (approved):
Looks great! Just a couple of non-blocking comments

Orrison:
@dgoetzit feel free to merge this in once you have had a chance to review my comments and mark the tasks as complete

app-modules/engagement/src/Actions/EngagementEmailChannelDelivery.php
dgoetzit: Yup! Will definitely do so as this integration grows.

app-modules/engagement/src/Models/EngagementDeliverable.php
dgoetzit: My only hesitation with this is it then requires the enum method to also accept the deliverable instance.

PR #52, opened 2023-08-23
[Feature] Subscribe to student

Orrison:
Sets up Student Subscription handling the following items:

https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/191/
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/193/
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/195/

### DIFF
--- a/app-modules/notifications/src/Models/Contracts/Subscribable.php
+++ b/app-modules/notifications/src/Models/Contracts/Subscribable.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Assist\Notifications\Models\Contracts;
+
+use Illuminate\Database\Eloquent\Model;
+
+/**
+ * Interface Subscribable
+ *
+ * @property string $id
+ *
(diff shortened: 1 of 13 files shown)


app-modules/notifications/src/Models/Contracts/Subscribable.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
dgoetzit, line 1: Planning to add the `subscriptions()` relation requirement here?

dgoetzit (approved):
Looks great! Left one non-blocking question, but I think it's probably taken care of in subsequent work.

app-modules/notifications/src/Models/Contracts/Subscribable.php
Orrison: Ah yeah I can add that in later

PR #54, opened 2023-08-24
[Feature] Subscription list filtering | Subscription on Student List View | Auto Subscribe arch on case

Orrison:
- Adds the ability to filter the Student List based on whether or not you are subscribed
- Can subscribe/unsubscribe via Student Liste view
- Auto subscribe for CaseItem and EngagementFile


app-modules/notifications/src/Models/Contracts/Subscribable.php
```diff
@@ -11,6 +11,4 @@
  *
  * @mixin Model
  */
-interface Subscribable
-{
-}
+interface Subscribable {}
```
dgoetzit, line 14: Do we still need this with the introduction of `CanTriggerAutoSubscription`?

app-modules/notifications/src/Listeners/NotifyUserOfSubscriptionCreated.php
```diff
@@ -0,0 +1,18 @@
...
+use Assist\Notifications\Events\SubscriptionCreated;
+
+class NotifyUserOfSubscriptionCreated implements ShouldQueue
+{
+    public function handle(SubscriptionCreated $event): void
+    {
+        Notification::make()
+            ->status('success')
+            ->title("You have been subscribed to {$event->subscription->subscribable->full}")
+            ->sendToDatabase($event->subscription->user);
```
dgoetzit, line 16: Do you think we might also want to deliver these via mail? 

An idea to consider for some flexibility: Standard Laravel notifications via `mail()` and `toArray()` channels. Send the DB notification immediately, delay the mail notification. If the user has read the DB notification, don't send the mail when we run a check after the delay has elapsed.

app-modules/notifications/src/Models/Contracts/CanTriggerAutoSubscription.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Assist\Notifications\Models\Contracts;
+
+interface CanTriggerAutoSubscription
```
dgoetzit, line 5: I wonder if because this is a contract, the name should be `TriggersAutoSubscription`, just so it doesn't sound like implementing is an option as opposed to a requirement.

app-modules/notifications/src/Actions/SubscriptionCreate.php
```diff
@@ -0,0 +1,27 @@
...
+
+class SubscriptionCreate
+{
+    public function handle(User $user, Subscribable $subscribable, bool $throwUniqueException = true): void
+    {
+        try {
+            $user->subscriptions()->create([
+                'subscribable_id' => $subscribable->getKey(),
+                'subscribable_type' => $subscribable->getMorphClass(),
+            ]);
```
dgoetzit, line 18: Should we try to first find an existing but deleted instance of this relationship? Or is the intention with the subscription soft delete to have a more fine grained historical trace (beyond the audit) of exactly when a user might have subscribed/unsubscribed from a Student/Prospect?

dgoetzit (approved):
Great work on this @Orrison!

I have a few questions/suggestions, but nothing that should prevent this from being merged in. We can address any feedback in subsequent PRs.

app-modules/notifications/src/Models/Contracts/Subscribable.php
Orrison: We do; this is the interface for the Subscribable Model (i.e. `Student` and `Prospect`) whereas `CanTriggerAutoSubscription` is for the related Models that can trigger an Auto Subscription like `CaseItem`

app-modules/notifications/src/Listeners/NotifyUserOfSubscriptionCreated.php
Orrison: Hmm, interesting. We could want to do that at some point.

Is it possible to wrap the logic into a Laravel Notification that could trigger either an email or Filament Notification? That way we have notifications in a central place.

app-modules/notifications/src/Actions/SubscriptionCreate.php
Orrison: I debated this in my head for a while, and I landed on deleting them and creating new records because I think data may soon be stored on a Subscription.

For example, the `via` concept we discussed where we could set "why" they were subscribed (i.e., direct subscription, touching a related item, etc.)

And I figured that we would want to retain that data rather than overwrite it.

app-modules/notifications/src/Listeners/NotifyUserOfSubscriptionCreated.php
dgoetzit: Yes - I think we'd be able to do that by just adding the `toDatabase()` method in our Laravel notification, and dispatching the Filament notification within that method.

app-modules/notifications/src/Actions/SubscriptionCreate.php
dgoetzit: Ahh yeah - that's a great point.

app-modules/notifications/src/Models/Contracts/CanTriggerAutoSubscription.php
Orrison: That's a good point, I am going to continue work on adding tests to this so I can change the name then

PR #56, opened 2023-08-29
[Feature] Add ability to receive texts

dgoetzit:
This PR introduces:

- The basic setup of consuming incoming webhooks
- The ability to receive texts from a webhook endpoint
- The ability to create an Engagement from the Student resource

TODOS and more tests will be addressed in the subsequent bulk PR work in order to not bloat this PR even more.


app-modules/assist-data-model/src/Filament/Resources/StudentResource/RelationManagers/EngagementsRelationManager.php
```diff
@@ -0,0 +1,127 @@
...
+                    }),
+            ]);
+    }
+
+    public function afterCreate(Engagement $engagement, array $deliveryMethods): void
+    {
+        $createDeliverablesForEngagement = resolve(CreateDeliverablesForEngagement::class);
+
+        $createDeliverablesForEngagement($engagement, $deliveryMethods);
+    }
```
Orrison, line 126: Just curious about your thoughts on whether or not this is something that should be handled by `created` event/listeners for `Engagement`?

app-modules/engagement/src/Actions/FindEngagementResponseSender.php
```diff
@@ -0,0 +1,25 @@
...
+use Assist\Prospect\Models\Prospect;
+use Assist\AssistDataModel\Models\Student;
+use Assist\Engagement\Exceptions\UnknownEngagementSenderException;
+
+class FindEngagementResponseSender
+{
+    public function __invoke(string $phoneNumber): Student|Prospect
+    {
+        // Student currently takes priority, but determine if we potentially want to store this response
+        // For *all* potential matches instead of just a singular result.
```
Orrison, line 14:
```suggestion
        // For *all* potential matches instead of just a singular result.
        // TODO: Make use of shared Student/Prospect implementation
```

I am adding this TODO here as I would like to discuss with you adding an interface, called `Entity` or `StudentEntity`.
It would enforce methods such as `getName()` and `getMobile()` so we can ensure that everywhere that we need to get the same value from either Model we know that no changes in the system would require changing what that column name is.

Orrison (approved):
Looks great! Just a couple of comments, but nothing blocking.

Feel free to merge in.

app-modules/assist-data-model/src/Filament/Resources/StudentResource/RelationManagers/EngagementsRelationManager.php
dgoetzit: I think it certainly can. Either way, we need to hook into functionality after the model is created in Filament, because we are reliant on the extra form fields that provide this information. But that's probably a slightly better way to handle this.

PR #72, opened 2023-09-08
[ASSIST-321] More robust Engagement Batch testing

dgoetzit:
This PR introduces more robust testing, and slightly more optimized dispatching of the Engagement Batch jobs.


app-modules/engagement/src/DataTransferObjects/EngagementBatchCreationData.php
```diff
@@ -0,0 +1,16 @@
...
+use App\Models\User;
+use Spatie\LaravelData\Data;
+use Illuminate\Support\Collection;
+
+class EngagementBatchCreationData extends Data
+{
+    public function __construct(
+        public User $user,
+        public Collection $records,
+        public array $data
```
Orrison, line 14: I think it may be preferred to use this DTO to ensure we don't have to pull items out of an array that we in the future, or static analysis, doesn't know exists post set.

Could we create properties for each item like `subject`, `delivery_methods`, etc. here and make items nullable and initially set to `null` that are optional, etc?

That way in `CreateEngagementBatch` we can call `$this->data->subject`.

dgoetzit: Sure - I'll make that adjustment.

Orrison (approved):

PR #74, opened 2023-09-08
[ASSIST-337] Remove ID presentation in list view of Service Request and Knowledge Base

dgoetzit:


Orrison (changes requested):
This looks good, but looks like a test needs to be adjusted.

Orrison (approved):

PR #78, opened 2023-09-12
[ASSIST-351, 352, 355, 357, 359]: Logo: User Story 1: Adding Theme Settings to the Settings Menu

danharrin:
The large line reduction is due to the fact that I removed an unused `composer.lock` file from the audit module while working.

https://github.com/canyongbs/assistforhighered-2/assets/41773797/ef0d1d41-8ef2-4695-ab15-6755b9f0991b


app-modules/audit/composer.json
```diff
@@ -6,9 +6,7 @@
   "license": "proprietary",
   "require": {
     "filament/filament": "^3.0",
-    "owen-it/laravel-auditing": "dev-feature/queued-auditing@dev",
-    "spatie/laravel-settings": "^3.2",
-    "filament/spatie-laravel-settings-plugin": "^3.0"
```
Orrison, line 11: It's fine that we moved these two to be dependencies in the root `composer.json` because we are using them in the root Laravel setup now as well.

But we may want to leave them here too. To keep continuity that this Module requires these items.

What do you think @dgoetzit and @danharrin?

app-modules/theme/src/Providers/ThemeServiceProvider.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Assist\Theme\Providers;
+
+use Filament\Panel;
+use Assist\Theme\ThemePlugin;
+use Illuminate\Support\ServiceProvider;
+
+class ThemeServiceProvider extends ServiceProvider
```
Orrison, line 9: If you could bring back the `boot()` with the empty morph map and the permission registry. Take a look at another Module for an example.

Those should be in all modules regardless and will be used a little bit later.

Orrison (changes requested):

app-modules/audit/composer.json
dgoetzit: Although we probably have missed a couple of places where this is the case, I think we should strive to add/keep dependencies where they would be necessary if the modules were standalone.

Orrison: I agree, though it is annoying that we will probably have to come into here and ensure that these numbers are kept up-to-date with root. I think it is important we keep track of the packages a module depends upon to some degree.

So @danharrin if you can just add these back here as well. You can keep them in root too though.

Orrison (approved):

Orrison (changes requested):
Looks good, just one merge conflict.

Orrison (approved):

PR #81, opened 2023-09-12
[ASSIST-299] AI Assistant Initial UI

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/299/


Orrison:
@danharrin, if you wouldn't mind giving this a look as well. Did some things in Livewire/Filament that I would appreciate your expertise on.

app-modules/assistant/resources/views/filament/pages/a-i-assistant.blade.php
```diff
@@ -0,0 +1,55 @@
...
+                    <path class="opacity-75" fill="currentColor" d="M4 12a8 8 0 018-8V0C5.373 0 0 5.373 0 12h4zm2 5.291A7.962 7.962 0 014 12H0c0 3.042 1.135 5.824 3 7.938l3-2.647z"></path>
+                </svg>
+            </div>
+        </form>
+        <script>
+            document.addEventListener('livewire:initialized', () => {
+                @this.on('current-message-saved', (event) => {
+                    @this.dispatch('ask');
+                });
+            });
```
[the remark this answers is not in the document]
Orrison: Yeah, it all has to do with the rendering.

I wanted the User-defined message to pop into the chat and clear the text box, then trigger the ask process.

This seems to be the only way to do so.

Other attempts like just using the Livewire default way of calling actions always resulted in the DOM waiting to re-render till all the requests completed. So the screen would sit there for a couple of seconds, then both the User message AND the response would pop into the chat. Which I felt was not very chat-like.

app-modules/assistant/resources/views/filament/pages/a-i-assistant.blade.php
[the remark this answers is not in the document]
Orrison: Done

app-modules/assistant/database/migrations/2023_09_11_214023_create_assistant_chat_messages_table.php
```diff
@@ -0,0 +1,18 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('assistant_chat_messages', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->foreignUuid('assistant_chat_id')->constrained('assistant_chats');
+            $table->longText('message');
+            $table->enum('from', ['user', 'assistant']);
```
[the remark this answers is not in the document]
Orrison: I believe this is different for Postgres. Postgres has much more advanced support for the enum column. It can be changed without deleting the column, etc.

I put this here to propose using it to add some data integrity to the DB side in some places.

I am also fine using a string-backed enum as well or completely instead.

What do you think?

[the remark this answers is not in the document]
Orrison: Ah you know what, nvm. I did some more research and apparently you can't delete enum values once created. So I will change this to a string.

app-modules/assistant/src/Filament/Pages/AIAssistant.php
danharrin, line 1: Adding this comment to suggest the use of response streaming

danharrin (changes requested):

app-modules/assistant/database/migrations/2023_09_11_214023_create_assistant_chat_messages_table.php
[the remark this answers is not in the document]
Orrison: Changed

app-modules/assistant/src/Filament/Pages/AIAssistant.php
Orrison: Created a ticket in the backlog to look into doing so. Which we will most likely do when the UI redesign comes for this.

app-modules/assistant/src/Filament/Pages/AIAssistant.php
```diff
@@ -0,0 +1,102 @@
...
+
+        $this->message = '';
+
+        $this->dispatch('current-message-saved');
+    }
+
+    #[On('ask')]
+    public function ask(): void
+    {
+        $ai = app(AIInterface::class);
```
danharrin, line 58: I believe you can resolve this dependency as a parameter to this method.

app-modules/assistant/src/Filament/Pages/AIAssistant.php
```diff
@@ -0,0 +1,102 @@
...
+
+            /** @var AssistantChat $assistantChat */
+            $assistantChat = $user->assistantChats()->create();
+
+            $this->chat->messages->each(function (ChatMessage $message) use ($assistantChat) {
+                $assistantChat->messages()->create($message->toArray());
+            });
+
+            $this->chat->id = $assistantChat->id;
+        }
```
danharrin, line 79: Can we make use of an early return here with `filled($this->chat->id)`?

dgoetzit (dismissed):

app-modules/assistant/src/Filament/Pages/AIAssistant.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    protected function setMessage(string $message, AIChatMessageFrom $from): void
+    {
+        if ($this->chat->id) {
```
danharrin, line 84:
```suggestion
        if (filled($this->chat->id)) {
```

This accounts for scenarios like `0`, `'0'`, `'null'`, `'false'`

app-modules/assistant/src/Models/AssistantChat.php
```diff
@@ -0,0 +1,46 @@
...
+ * @method static \Illuminate\Database\Eloquent\Builder|AssistantChat whereId($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|AssistantChat whereUpdatedAt($value)
+ * @method static \Illuminate\Database\Eloquent\Builder|AssistantChat whereUserId($value)
+ *
+ * @mixin \Eloquent
+ */
+class AssistantChat extends BaseModel
+{
+    protected $fillable = [
+        'user_id',
```
danharrin, line 34: I'd maybe push back on this and say, if you're going to use mass assignment protection instead of `$guarded`, this foreign key should not be fillable.

app-modules/assistant/src/Models/AssistantChatMessage.php
```diff
@@ -0,0 +1,51 @@
...
+ * @method static Builder|AssistantChatMessage whereId($value)
+ * @method static Builder|AssistantChatMessage whereMessage($value)
+ * @method static Builder|AssistantChatMessage whereUpdatedAt($value)
+ *
+ * @mixin Eloquent
+ */
+class AssistantChatMessage extends BaseModel
+{
+    protected $fillable = [
+        'assistant_chat_id',
```
danharrin, line 38: Also here, not sure about a fillable foreign key

app-modules/assistant/src/Providers/AssistantServiceProvider.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        Panel::configureUsing(fn (Panel $panel) => $panel->plugin(new AssistantPlugin()));
+
+        $this->app->bind(AIInterface::class, AzureOpenAI::class);
+    }
+
+    public function boot()
+    {
+        Relation::morphMap([
+            'assistant-chat' => AssistantChat::class,
```
danharrin, line 28: I believe other models are using `snake_case`, not `kebab-case` (as seen in `MorphServiceProvider`)

danharrin (changes requested):

app-modules/assistant/src/Filament/Pages/AIAssistant.php
Orrison: Updated

app-modules/assistant/src/Filament/Pages/AIAssistant.php
Orrison: Updated

app-modules/assistant/src/Providers/AssistantServiceProvider.php
Orrison: Ah, yes, good catch. Updated.

app-modules/assistant/src/Filament/Pages/AIAssistant.php
Orrison: Changed

dgoetzit (approved):

danharrin (dismissed):

dgoetzit (approved):

PR #87, opened 2023-09-14
Feature [ASSIST-297] [ASSIST-365] Azure Open AI Integration

dgoetzit:
Initial work for the setup of the Azure Open AI client, as well as the implementation and integration with the frontend AI Assistant.

There is still work to be done here in terms of testing, but want to ensure this doesn't become a blocker for others working on pieces surrounding this feature.


app-modules/integration-ai/config/settings.php
```diff
@@ -0,0 +1,88 @@
+<?php
```
Orrison, line 1: Don't we already have this file in root? Is there any issues with having this in two places?

app-modules/integration-ai/src/Client/BaseAIChatClient.php
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        return $response->choices[0]?->delta?->content ?: null;
+    }
+
+    protected function examineFinishReason(CreateStreamedResponse $response): void
+    {
+        match ($response->choices[0]->finishReason) {
+            'length' => throw new TokensExceededException('Your response was not successfully generated due to the max_tokens parameter or token limit being exceeded.'),
+            'content_filter' => throw new ContentFilterException('Your response was not successfully generated due to a flag from our content filters.'),
```
Orrison, line 106: Does this mean we can do what we discussed, where we could flag the audit record as like "concerning" or something like that?

app-modules/integration-ai/src/Client/BaseAIChatClient.php
```diff
@@ -0,0 +1,141 @@
...
+
+    protected function formatMessagesFromChat(Chat $chat): array
+    {
+        return [
+            ['role' => 'system', 'content' => $this->addContextToMessages()],
+            ...collect($chat->messages)->map(function (array $message) {
+                return [
+                    'role' => $message['from'],
+                    'content' => $message['message'],
+                ];
```
Orrison, line 119:
```suggestion
            ...collect($chat->messages)->map(function (ChatMessage $message) {
                return [
                    'role' => $message->from,
                    'content' => $message->message,
                ];
```

At this point, shouldn't the message be the `ChatMessage` DTO? Let's keep the references as such.

app-modules/integration-ai/src/DataTransferObjects/DynamicContext.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Assist\IntegrationAI\DataTransferObjects;
+
+use Spatie\LaravelData\Data;
+use Illuminate\Database\Eloquent\Model;
+
+class DynamicContext extends Data
+{
+    public Model $record;
```
Orrison, line 10: I believe you could type hint this to `ProvidesDynamicContext` then in the `ProvidesDynamicContext` give it a docblock `@mixin` to `Model` to have with auto completion.

Giving us protection that only `ProvidesDynamicContext` implementations can be used here but still giving us `Model` auto-complete.

config/openai.php
Orrison, line 1: Should this go into a Module?

config/services.php
Orrison, line 1: Should this go into a Module?

app-modules/integration-ai/src/Client/BaseAIChatClient.php
dgoetzit: It should be - but this was resulting in a type error when I originally had that. Will revisit.

config/openai.php
dgoetzit: Did we resolve issues with config in multiple places? I think we still need to dig a bit deeper into this, as typical package convention would still have config published into the core.

app-modules/integration-ai/src/Client/BaseAIChatClient.php
dgoetzit: Yes - we could add handling in here. From the docs, it seems like this is also/already done at the admin level based on settings in Azure. But, it could allow us to flag audits within the system as well with some sort of identifier.

app-modules/integration-ai/config/settings.php
dgoetzit: Hmmm - I thought there was initially a settings file in the `audit` module, so I had created one here as well. But, we probably want to keep these centralized.

Orrison: I believe @danharrin recently moved it to root since he used it in a different Module other than `audit`

config/openai.php
Orrison: Alright, we can leave it here for now then.

app-modules/integration-ai/src/DataTransferObjects/DynamicContext.php
dgoetzit: Should be resolved.

app-modules/integration-ai/src/Client/BaseAIChatClient.php
dgoetzit: Should be resolved.

app-modules/integration-ai/src/Client/BaseAIChatClient.php
dgoetzit: Should be resolved.

app-modules/integration-ai/config/settings.php
dgoetzit: Should be resolved.

app-modules/integration-ai/src/Client/BaseAIChatClient.php
```diff
@@ -0,0 +1,143 @@
...
+                    'role' => $message->from,
+                    'content' => $message->message,
+                ];
+            }),
+        ];
+    }
+
+    protected function addContextToMessages(): string
+    {
+        return $this->systemContext . ' ' . $this->dynamicContext;
```
danharrin, line 128:
```suggestion
        return "{$this->systemContext} {$this->dynamicContext}";
```

danharrin (changes requested):

app-modules/integration-ai/src/Client/BaseAIChatClient.php
dgoetzit: resolved.

Orrison (approved):

danharrin (approved):

PR #92, opened 2023-09-15
[ASSIST-311] Impersonate Users to Test Access for Administrators

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/311/

The only thing weird about this is a graphical bug:

image

This is caused by a style in the package and I'm not sure how to remove it.


app/Models/User.php
```diff
@@ -260,6 +263,16 @@ public function canAccessPanel(Panel $panel): bool
...
     }
 
+    public function canImpersonate(): bool
+    {
+        return $this->can('authorization.impersonate');
+    }
+
+    public function canBeImpersonated(): bool
+    {
+        return $this->cannot('authorization.super_admin');
```
Orrison, line 274: This should probably just be a role check for `super_admin` I am not sure if we will be keeping this permission.

Orrison (changes requested):
Just had one suggestion and I also pushed up a fix for the sidebar during impersonation.

Orrison (approved):

PR #99, opened 2023-09-18
[ASSIST-365] User Consent

dgoetzit:


app-modules/assistant/composer.lock
danharrin, line 1: Can we remove lock files from packages, as we discussed?

dgoetzit: Done.

app-modules/consent/database/factories/ConsentAgreementFactory.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @extends \Assist\Consent\Models\ConsentAgreement>
+ */
+class ConsentAgreementFactory extends Factory
+{
+    public function definition(): array
+    {
+        return [
+            'type' => fake()->randomElement(ConsentAgreementType::cases()),
```
Orrison, line 16: Do you need the `->value` here at the end?

app-modules/consent/src/Filament/Resources/ConsentAgreementResource.php
```diff
@@ -0,0 +1,87 @@
...
+                TextColumn::make('title'),
+            ])
+            ->filters([
+            ])
+            ->actions([
+                Tables\Actions\EditAction::make(),
+            ])
+            ->bulkActions([])
+            ->emptyStateActions([]);
+    }
```
Orrison, line 70: Could we move these to the individual pages like we have done for other resources?

app-modules/consent/database/factories/ConsentAgreementFactory.php
dgoetzit: Good catch - fixed.

app-modules/consent/src/Filament/Resources/ConsentAgreementResource.php
dgoetzit: This is resolved.

Orrison (dismissed):

app-modules/assistant/src/Filament/Pages/AIAssistant.php
```diff
@@ -39,11 +42,29 @@ class AIAssistant extends Page
...
+    }
+
     public function mount(): void
     {
         /** @var User $user */
         $user = auth()->user();
 
+        $this->authorize('assistant.access');
+
+        $this->consentAgreement = ConsentAgreement::where('type', ConsentAgreementType::AZURE_OPEN_AI)->first();
```
danharrin, line 66:
```suggestion
        $this->consentAgreement = ConsentAgreement::query()
            ->where('type', ConsentAgreementType::AZURE_OPEN_AI)
            ->first();
```

app-modules/consent/src/Enums/ConsentAgreementType.php
danharrin, line 1: Just realised that enums throughout the app have a mixture of UPPERCASE and PascalCase case names. Might be something we want to address in the future, but do we have a preference for this PR? Personally I prefer PascalCase.

app-modules/consent/src/Filament/Resources/ConsentAgreementResource/Pages/EditConsentAgreement.php
```diff
@@ -0,0 +1,50 @@
...
+class EditConsentAgreement extends EditRecord
+{
+    protected static string $resource = ConsentAgreementResource::class;
+
+    public function form(Form $form): Form
+    {
+        return parent::form($form)
+            ->schema([
+                TextInput::make('type')
+                    ->readonly()
```
danharrin, line 21: What is the desired behaviour here? I think you might want `disabled()`, so the field is not editable via JavaScript. `readonly()` is essentially just frontend styling and does not protect the field in the backend.

app-modules/consent/src/Filament/Resources/ConsentAgreementResource/Pages/EditConsentAgreement.php
```diff
@@ -0,0 +1,50 @@
...
+                        Textarea::make('body')
+                            ->required()
+                            ->rows(5)
+                            ->columnSpan([
+                                'sm' => 2,
+                                'xl' => 3,
+                                '2xl' => 4,
+                            ]),
+                    ]),
+            ]);
```
danharrin, line 43: Does the Fieldset itself need a fixed number of columns, since the fields inside it can span 2, 3 or 4?

danharrin (changes requested):

app-modules/assistant/src/Filament/Pages/AIAssistant.php
Orrison: Is there a point in adding `query` here? With our IDE helper configuration, `where` should be auto-completable and static analysis should be aware of it.

app-modules/consent/src/Enums/ConsentAgreementType.php
Orrison: I would prefer PascalCase. What do you think @dgoetzit? We can add a ticket in the backlog once we have a consensus

dgoetzit: I agree we should standardize this. I don't necessarily have a strong preference towards one or the other. I think PascalCase is more visually appealing, but UPPERCASE is a more conventional approach for constants.

app-modules/consent/src/Filament/Resources/ConsentAgreementResource/Pages/EditConsentAgreement.php
dgoetzit: The desired behavior is disabling - updated.

app-modules/consent/src/Enums/ConsentAgreementType.php
dgoetzit: Yup that's fine with me - this has been updated.

app-modules/assistant/src/Filament/Pages/AIAssistant.php
danharrin: Personally, I think the chaining looks cleaner when the magic static calls are avoided. Your call, but it's also nice to be consistent and always start a query with `::query()`.

Orrison: Hmm, I see your point about the cleanliness of the line. But I always felt like it was unnecessary, especially if you have the IDE helper.

@dgoetzit do you have a preference?

Orrison (approved):

Orrison:
@dgoetzit looks like there is just a merge conflict blocking this now.

Also @danharrin if you can resolve your open comment if you feel it is resolved for now? As @dgoetzit said, we can have a larger discussion there.

danharrin (approved):

danharrin:
Do we have a ticket created to deal with the rest of the enums?

Orrison (approved):
Merge c

PR #101, opened 2023-09-18
[ASSIST-413]: Implement badge for active open alerts for a given student or prospect. 

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/413/


dgoetzit (changes requested):
This looks good - I would just recommend updating the `.env.example` as this now requires a Cache Driver other than file.

Orrison:
> This looks good - I would just recommend updating the `.env.example` as this now requires a Cache Driver other than file.

Ah, good point. Change made.

app-modules/alert/src/Filament/RelationManagers/AlertsRelationManager.php
```diff
@@ -23,6 +27,17 @@ class AlertsRelationManager extends RelationManager
 {
     protected static string $relationship = 'alerts';
 
+    public static function getBadge(Model $ownerRecord, string $pageClass): ?string
+    {
+        /** @var Student|Prospect $ownerRecord */
+        $alertCount = Cache::tags('alert-count')->remember('alert-count-' . $ownerRecord->id, now()->addMinutes(5), function () use ($ownerRecord) {
+            // TODO: When it is decided how alerts are "resolved" this will need to take that into account to only display unresolved alerts
+            return $ownerRecord->alerts()->count();
+        });
```
danharrin, line 36:
```suggestion
        $alertCount = Cache::tags('alert-count')
            ->remember(
                'alert-count-' . $ownerRecord->id,
                now()->addMinutes(5),
                function () use ($ownerRecord): int {
                    // TODO: When it is decided how alerts are "resolved" this will need to take that into account to only display unresolved alerts
                    return $ownerRecord->alerts()->count();
                });
```

Also, do we prefer the usage of Facades or helpers() in this app?

app-modules/alert/src/Observers/AlertObserver.php
```diff
@@ -17,4 +18,9 @@ public function created(Alert $alert): void
 
         AlertCreated::dispatch($alert);
     }
+
+    public function saved(Alert $alert): void
```
danharrin, line 22: Do we also want to listen for the `deleted` event?

danharrin (changes requested):

app-modules/alert/src/Observers/AlertObserver.php
Orrison: Yeah that is a good point, added.

danharrin (approved):

dgoetzit (approved):

PR #102, opened 2023-09-18
[ASSIST-391] 403 on Azure login

dgoetzit:
This resolves an issue with the Azure login, which stemmed from users not being seeded and internal users not being able to authenticate via Azure.


app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -36,7 +36,7 @@ public function callback(SocialiteProvider $provider)
             ->where('email', $socialiteUser->getEmail())
             ->first();
 
-        if (! $user?->is_external) {
+        if ($user?->is_external) {
```
Orrison, line 39: I think the `!` needs to stay doesn't it?

`is_external` represents that their login is external, i.e. they log in via SSO.

Orrison (changes requested):
@dgoetzit I think perhaps you may be thinking that `is_external` represents Users that are theoretically external to the organization.

But what I really means is whether or not they login "externally" or not, i.e. via SSO. So I think some changes here don't coincide with that.

app-modules/authorization/src/Http/Controllers/SocialiteController.php
dgoetzit: Ahh yeah I think I misunderstood that field, and the nature of the external login. I thought the external abilities were for vendors who might not be part of an organization, and therefore may not be able to authenticate through OAuth?

Orrison: Yeah. Do you think there is a case for renaming `is_external`?

dgoetzit: Confirmed via external conversation - added a ticket to the backlog to address this.

Orrison (approved):

PR #104, opened 2023-09-19
[ASSIST-273]: Record Importer MVP for Prospects

danharrin:
Currently, these files are living in the app. Once I've found time to write some documentation, I'll move them into the Filament core and refactor this app. It will also give us a good period for thorough testing and API feedback.

New prospects can be imported, and existing prospects can be updated. The `email` and `email_2` columns are used to match. If *either* email is found in *either* column, the prospect is "matched".

I might need some guidance as to whether or not I implemented authorization correctly.


app-modules/authorization/config/permissions/web/model.php
```diff
@@ -3,6 +3,7 @@
 return [
     'view-any',
     'create',
+    'import',
```
Orrison, line 6: @dgoetzit in your opinion does this belong here in the Model permissions, or should it be in `custom`?

app-modules/prospect/src/Imports/ProspectImporter.php
```diff
@@ -0,0 +1,154 @@
...
+
+    public static function getFailureNotificationBody(int $processedRows): string
+    {
+        return "Something went wrong after importing {$processedRows} rows of prospects.";
+    }
+
+    public function getValidationFailureNotificationBody(string $message): string
+    {
+        return "A validation error occurred while importing a prospect: {$message}. This skipped row belonged to the prospect with full name \"{$this->data['full_name']}\".";
+    }
```
Orrison, line 153: Does this mean that we import rows individually? So if one row fails, that doesn't prevent the other rows from importing?

app/Filament/Actions/ImportAction.php
```diff
@@ -0,0 +1,267 @@
...
+        parent::setUp();
+
+        $this->modalHeading(fn (ImportAction $action): string => "Import {$action->getPluralModelLabel()}");
+
+        $this->modalSubmitActionLabel('Start import');
+
+        $this->groupedIcon('heroicon-m-arrow-up-tray');
+
+        $this->form(fn (ImportAction $action): array => array_merge([
+            Forms\Components\FileUpload::make('file')
```
Orrison, line 42: Small nit but, we prefer the direct `use` statements throughout the app to this partial. So that it is just `FileUpload` here.

app/Jobs/ImportCsv.php
```diff
@@ -0,0 +1,79 @@
...
+        /** @var User $user */
+        $user = $this->import->user;
+
+        auth()->login($user);
+
+        foreach ($this->rows as $row) {
+            ($this->importer)($row);
+
+            $this->import->increment('processed_rows');
+        }
```
Orrison, line 64: If an individual row fails to import here, then this entire group stops processing, correct?

What does failure recovery look like for a User, and does that failure recovery prevent the import of duplicate records and/or prevent attempting to import records again?

Orrison (changes requested):
This looks amazing! Just some questions and suggestions.

Orrison:
A couple of other notes:

- Does this cover "This interface should offer guidance on the required CSV format, perhaps with a downloadable template or sample file." within the ticket? Or will that be done in a separate PR?
- Is it possible to not allow a single row failure from importing others within a job? And provide info on the row# / other details?

app-modules/prospect/src/Imports/ProspectImporter.php
danharrin: Correct, and we send a notification for each row that fails validation, with the specific validation error. As it can differ.

app/Jobs/ImportCsv.php
danharrin: Hmm, in what way would it fail? With an exception or something? We can catch that.

Imports should never be duplicated when retried, since we upsert, not insert.

danharrin:
1. It does not cover adding a template, we could maybe have a discussion on how useful that is, since the system is flexible enough to not require a specific format, everything can be mapped in the UI.
2. We do provide the full name of the prospect that fails. What would the advantage be of preventing other rows from getting imported when a validation error occurs? Technically, rows can be imported in any order.

app/Jobs/ImportCsv.php
Orrison: Yeah, what I mean is `$this->rows` is a group of rows, right?

If a single one of these rows failes/throws an exception, the others after it won't process right?

Orrison (changes requested):

Orrison:
> 1. It does not cover adding a template, we could maybe have a discussion on how useful that is, since the system is flexible enough to not require a specific format, everything can be mapped in the UI.
> 2. We do provide the full name of the prospect that fails. What would the advantage be of preventing other rows from getting imported when a validation error occurs? Technically, rows can be imported in any order.

1. I think we will still need it so that the end User has something to work off of as they get this data from the other systems 
2. No, sorry. I am saying that I think they way this is setup, a single row failure would prevent the import of other rows in one of the jobs. I commented in the comment in the job file

dgoetzit (changes requested):
This is looking great @danharrin! Just a few comments/suggestions/questions on my end.

app-modules/prospect/src/Imports/ProspectImporter.php
```diff
@@ -0,0 +1,154 @@
...
+
+                    if (! $source) {
+                        return;
+                    }
+
+                    $record->source()->associate($source);
+                })
+                ->requiredMapping(),
+            ImportColumn::make('description'),
+            ImportColumn::make('email')
```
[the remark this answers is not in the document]
danharrin: They are not required fields in the database or UI at the moment

danharrin:
@Orrison @dgoetzit have you discussed with Joe before if email is required? Because it currently isn't

app-modules/prospect/src/Policies/ProspectPolicy.php
```diff
@@ -29,6 +29,13 @@ public function create(User $user): Response
             : Response::deny('You do not have permission to create prospects.');
     }
 
+    public function import(User $user): Response
+    {
+        return $user->can('prospect.import')
+            ? Response::allow()
+            : Response::deny('You do not have permission to import prospects.');
```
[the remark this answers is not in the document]
danharrin: I just copied this from other policies, do we need to update it there too?

Orrison:
> @Orrison @dgoetzit have you discussed with Joe before if email is required? Because it currently isn't

We have not, you may need to check with Joe

danharrin:
Good spot, probably good to also confirm he wants this done at the database level and in the resource

app-modules/authorization/config/permissions/web/model.php
danharrin: Done

app/Filament/Actions/ImportAction.php
danharrin: Done

app/Jobs/ImportCsv.php
danharrin: Fixed

database/migrations/2023_09_18_151659_create_imports_table.php
```diff
@@ -0,0 +1,26 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        Schema::create('imports', function (Blueprint $table) {
+            $table->id();
```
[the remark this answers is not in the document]
danharrin: Done

app/Models/Import.php
```diff
@@ -0,0 +1,43 @@
...
+namespace App\Models;
+
+use App\Imports\Importer;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @mixin IdeHelperImport
+ */
+class Import extends Model
```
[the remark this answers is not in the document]
danharrin: Done

app-modules/prospect/src/Imports/ProspectImporter.php
```diff
@@ -0,0 +1,154 @@
...
+                    if ($status) {
+                        return;
+                    }
+
+                    $fail('The selected status is invalid.');
+                }])
+                ->guess(['status'])
+                ->fillRecordUsing(function (Prospect $record, string $state) {
+                    $status = ProspectStatus::query()
+                        ->where('name', $state)
```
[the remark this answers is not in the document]
danharrin: Done

app-modules/prospect/src/Imports/ProspectImporter.php
```diff
@@ -0,0 +1,154 @@
...
+                    if ($source) {
+                        return;
+                    }
+
+                    $fail('The selected source is invalid.');
+                }])
+                ->guess(['source'])
+                ->fillRecordUsing(function (Prospect $record, string $state) {
+                    $source = ProspectSource::query()
+                        ->where('name', $state)
```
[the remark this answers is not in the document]
danharrin: Done

app-modules/prospect/src/Imports/ProspectImporter.php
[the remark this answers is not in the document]
dgoetzit: Where did we end up landing on this?

app/Jobs/ImportCsv.php
```diff
@@ -0,0 +1,119 @@
...
+        return $this->importer->getJobTags();
+    }
+
+    protected function logFailedRow(array $data, ?string $validationError = null): void
+    {
+        $failedRow = new FailedImportRow();
+        $failedRow->import()->associate($this->import);
+        $failedRow->data = $data;
+        $failedRow->validation_error = $validationError;
+        $failedRow->save();
```
dgoetzit, line 104: Can we use mass assignment here? Or is there a reason not to on `FailedImportRow`?

```suggestion
        $this->import->failedRows()->create([
            'data' => $data,
            'validation_error' => $validationError,
        ]);
```

app/Console/Kernel.php
```diff
@@ -25,7 +26,7 @@ protected function schedule(Schedule $schedule): void
         $schedule->command(DispatchQueueCheckJobsCommand::class)->everyMinute();
 
         $schedule->command(PruneCommand::class, [
-            '--model' => [Audit::class, AssistantChatMessageLog::class],
+            '--model' => [Audit::class, AssistantChatMessageLog::class, FailedImportRow::class],
```
dgoetzit, line 29: Do we need to add a task to the backlog to make the pruning settings configurable?

dgoetzit (changes requested):

app/Jobs/ImportCsv.php
danharrin: This will be moved to Filament at some point, and I don't particularly like mass assignment so this code will stay the same

app/Console/Kernel.php
danharrin: Yeah, good idea to add to the backlog

app-modules/prospect/src/Imports/ProspectImporter.php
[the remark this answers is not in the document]
danharrin: Joe has not yet responded to my question

app/Jobs/ImportCsv.php
Orrison: I am interested to hear why you don't like mass assignments that much. The suggestion seems a little more readable to me, but perhaps this is not always the case.

app-modules/prospect/src/Imports/ProspectImporter.php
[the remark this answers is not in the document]
Orrison: If we decided to move forward with making email a required field, what is the level of effort in doing so / effect of that change?

app-modules/prospect/src/Imports/ProspectImporter.php
```diff
@@ -0,0 +1,154 @@
...
+        $email2 = $this->data['email_2'] ?? null;
+
+        $emails = array_filter(
+            [$email, $email2],
+            fn (mixed $state): bool => filled($state),
+        );
+
+        if (empty($emails)) {
+            return new Prospect();
+        }
```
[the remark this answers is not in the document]
Orrison: I agree, at this point I think we can move forward with email being required. We can make the change in the DB soon as well.

Without an email engagements and other functionality just doesn't work, it would not make sense to not have an email.

app/Console/Kernel.php
Orrison: Created the ticket for this: https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/451/

Orrison (changes requested):
Looking great! Just a few comments, as well as one last requirement for preventing CSVs bigger than 100,000 rows of data. (Probably 100,001 if we include the header in the count)

app/Jobs/ImportCsv.php
danharrin: It usually comes down to the fact that I am often associating multiple related models at once. With mass assignment, I need to then pass in `_id` columns into the array, which I don't like, as IMO the relationship should handle the specifics of the relationship columns with `associate()`.

I know that in this case we are only associating one relationship, but I like to keep things consistent, and only using mass assignment in some places and not others feels weird.

It's purely personal preference though.

danharrin:
Email is now required, and I've used `configureUsing()` to set the max import rows to 100,000.

Orrison (approved):

dgoetzit (approved):

PR #113, opened 2023-09-22
[ASSIST-455] Task Management Updates

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/455/

### DIFF
--- a/app-modules/task/src/Filament/Concerns/TaskViewActionInfoList.php
+++ b/app-modules/task/src/Filament/Concerns/TaskViewActionInfoList.php
@@ -19,47 +19,43 @@ trait TaskViewActionInfoList
     public function taskInfoList(): array
     {
         return [
-            Split::make(
-                [
-                    Grid::make()
-                        ->schema(
-                            [
-                                TextEntry::make('description')
-                                    ->columnSpanFull(),
-                                TextEntry::make('assignedTo.name')
-                                    ->label('Assigned To')
-                                    ->url(fn (Task $record) => $record->assignedTo ? UserResource::getUrl('view', ['record' => $record->assignedTo]) : null)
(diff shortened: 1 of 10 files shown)


Orrison (changes requested):
This looks great! If you could please update the Kanban view to display the task description in the card as well? Replacing the description dummy text I had there.

You will probably want to truncate that after a reasonable length.

app-modules/task/src/Filament/Concerns/TaskViewActionInfoList.php
```diff
@@ -19,47 +19,43 @@ trait TaskViewActionInfoList
...
+                            ->columnSpanFull(),
+                        TextEntry::make('description')
+                            ->columnSpanFull(),
+                        TextEntry::make('assignedTo.name')
+                            ->label('Assigned To')
+                            ->url(fn (Task $record) => $record->assignedTo ? UserResource::getUrl('view', ['record' => $record->assignedTo]) : null)
+                            ->default('Unassigned'),
+                        TextEntry::make('concern.display_name')
+                            ->label('Concern')
+                            ->getStateUsing(fn (Task $record) => $record->concern->{$record->concern::displayNameKey()})
```
dgoetzit, line 35: We'll need to account for a potentially `null` concern here.

dgoetzit (changes requested):
Looks good! Just one minor change on my end.

mxm1070:
> This looks great! If you could please update the Kanban view to display the task description in the card as well? Replacing the description dummy text I had there.
> 
> You will probably want to truncate that after a reasonable length.

Done, and I set the limit to 50 characters.

Orrison (approved):

PR #114, opened 2023-09-23
[Bugfix] Fix bug when concern is null for a Task

Orrison:


app-modules/task/src/Filament/Resources/TaskResource/Pages/ListTasks.php
```diff
@@ -51,7 +51,7 @@ public function table(Table $table): Table
                     ->url(fn (Task $record) => $record->assignedTo ? UserResource::getUrl('view', ['record' => $record->assignedTo]) : null),
                 TextColumn::make('concern.display_name')
                     ->label('Concern')
-                    ->getStateUsing(fn (Task $record) => $record->concern->{$record->concern::displayNameKey()})
+                    ->getStateUsing(fn (Task $record) => $record->concern?->{$record->concern::displayNameKey()})
```
danharrin, line 54: Adding a return type could have avoided this problem in the first place, if PHPStan is on a high enough level

```suggestion
                    ->getStateUsing(fn (Task $record): ?string => $record->concern?->{$record->concern::displayNameKey()})
```

danharrin (changes requested):

app-modules/task/src/Filament/Resources/TaskResource/Pages/ListTasks.php
Orrison: Done

danharrin (approved):

dgoetzit (approved):

PR #116, opened 2023-09-26
[ASSIST-467] Teams

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/467/


app-modules/task/src/Filament/Resources/TaskResource/Pages/ListTasks.php
```diff
@@ -66,9 +67,19 @@ public function table(Table $table): Table
                 Filter::make('my_tasks')
                     ->label('My Tasks')
                     ->query(
-                        fn ($query) => $query->where('assigned_to', auth()->id())
-                    )
-                    ->default(),
+                        fn (Builder $query) => $query->where('assigned_to', auth()->id())
+                    ),
```
Orrison, line 71: Was there a reason we removed `->default()` off of the `my_tasks` filter here?

I believe we do still want this to be the default on page load.

app-modules/team/database/migrations/2023_09_26_162558_add_team_id_to_users_table.php
Orrison, line 1: Because we are not in production yet I think it may be best for us to just move this to the `users` table create migration now.

app-modules/team/src/Filament/Resources/TeamResource.php
```diff
@@ -0,0 +1,33 @@
...
+            UsersRelationManager::make(),
+        ];
+    }
+
+    public static function getPages(): array
+    {
+        return [
+            'index' => ListTeams::route('/'),
+            'create' => CreateTeam::route('/create'),
+            'edit' => EditTeam::route('/{record}/edit'),
```
Orrison, line 30: Let's go ahead and add a View page as well. There will most likely need to be Users that can view a team and see info about it, but not edit.

app-modules/team/src/Models/Membership.php
Orrison, line 1: Is this Pivot Model left over from when we thought a User would be a part of multiple Teams? If so, we should delete it if it is no longer needed.

We could leave the Team/User relationship how it is and delete this, or if you wanted to use this pivot still but create a unique index on `user_id` to prevent a User from being on multiple teams that could get us the quick ability to allow multiple teams by quickly removing that index.
What do you think? I am fine either way.

Orrison (changes requested):
Looking great, just a few comments/questions.

Let's also remember not to resolve unauthored comments in PRs per our team decision a few weeks ago. The individual raising the concern should resolve. To track progress on whether or not you addressed a comment, feel free to react or comment.

app-modules/task/src/Filament/Resources/TaskResource/Pages/ListTasks.php
mxm1070: I removed it because it was causing weirdness with the my team's task filter. This is why I asked Dan about programmatically removing filters and I'm still not sure about it.

Orrison: Ah, I see. You mean if you have both selected at the same time.

Let's bring the default back for now on the `my_team` filter, and we can add a TODO and a ticket to address, making sure neither filter can be on at the same time. But for now, it's an acceptable issue with the requirement of the My Tasks page being the first loaded filter.

app-modules/team/database/migrations/2023_09_26_162558_add_team_id_to_users_table.php
Orrison: Ah, interesting, I see your point. With that in mind, I would possibly advocate for bringing the Membership pivot back (and making user_id within that table unique to only allow 1 Membership per User.

That would keep the files within the Module and keep the team membership separate from the `user` table and prevent pollination if the teams module was not installed.

If we don't want to go that far now, then I would just move this into the users create.

app-modules/team/src/Models/Membership.php
Orrison: See my comment above on the migration.

app-modules/task/src/Filament/Resources/TaskResource/Pages/ListTasks.php
mxm1070: Actually I got this to work after speaking to Dan.

app-modules/team/database/migrations/2023_09_26_162558_add_team_id_to_users_table.php
mxm1070: I don't really want to mess with the Membership model until we want to allow multiple teams so I moved the foreign key into the users table.

Orrison: Sounds good, I have created a ticket for us to explore it at a later time. https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/509/

Orrison (approved):

PR #117, opened 2023-09-26
[ASSIST-501] Bug: Notifications no longer showing name after change to full_name

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/501/

### DIFF
--- a/app-modules/alert/src/Notifications/AlertCreatedNotification.php
+++ b/app-modules/alert/src/Notifications/AlertCreatedNotification.php
@@ -24,7 +24,7 @@ class AlertCreatedNotification extends Notification implements ShouldQueue
     {
         return FilamentNotification::make()
             ->status('warning')
-            ->title("A {$this->alert->severity->value} severity alert has been created for {$this->alert->concern->full}")
+            ->title("A {$this->alert->severity->value} severity alert has been created for {$this->alert->getSubscribable()?->getSubscriptionDisplayName()}")
             ->toDatabase()
             ->data;
     }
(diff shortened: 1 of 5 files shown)


app-modules/alert/src/Notifications/AlertCreatedNotification.php
```diff
@@ -24,7 +24,7 @@ public function toDatabase(User $notifiable): array
     {
         return FilamentNotification::make()
             ->status('warning')
-            ->title("A {$this->alert->severity->value} severity alert has been created for {$this->alert->concern->full}")
+            ->title("A {$this->alert->severity->value} severity alert has been created for {$this->alert->getSubscribable()?->getSubscriptionDisplayName()}")
```
Orrison, line 27: I would stick with the `concern` here still. And rather than this new  `getSubscriptionDisplayName()` method you use here we actually should be using the Educatable method to get the displayName column or something like that. That makes sure that we get the correct data whether it is a Prospect or Student. Even though right now those columns/data is stored the same. It may not in the future

app-modules/notifications/src/Models/Concerns/HasSubscriptions.php
```diff
@@ -11,4 +11,9 @@ public function subscriptions(): MorphMany
     {
         return $this->morphMany(Subscription::class, 'subscribable');
     }
+
+    public function getSubscriptionDisplayName(): string
+    {
+        return $this->full_name;
+    }
```
Orrison, line 18: Same comment as above. I don't think we need to introduce something here since we have this with Educatable

Orrison (changes requested):

Orrison (approved):

PR #120, opened 2023-09-28
[ASSIST-331]: Dashboard: Loading Enhancements

danharrin:
This is the first stage of identifying a possible performance issue on the staging server. Debugbar will help us identify any slow database queries or Livewire components which could affect the dashboard.

To clarify: this does not introduce any loading enhancement fixes, the ticket should be moved from "blocked" to "in progress" when this is merged.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -122,3 +122,5 @@ AZURE_OPEN_AI_ENABLE_TEST_MODE=true
 
 # A comma separated list of emails that will be seeded as internal users
 DEMO_INTERNAL_USER_EMAILS=
+
+DEBUGBAR_ENABLED=false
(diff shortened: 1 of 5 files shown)


.env.example
```diff
@@ -122,3 +122,5 @@ AZURE_OPEN_AI_ENABLE_TEST_MODE=true
 
 # A comma separated list of emails that will be seeded as internal users
 DEMO_INTERNAL_USER_EMAILS=
+
+DEBUGBAR_ENABLED=false
```
danharrin, line 126: Added to ensure that development environments are not slowed when debugbar is not being used

config/debugbar.php
danharrin, line 1: This config file has been published in order to disable the `views` collector, which is notoriously slow

Orrison (approved):

PR #121, opened 2023-09-29
[ASSIST-463] Implement Interaction Importer

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/463/


app-modules/interaction/src/Imports/InteractionsImporter.php
```diff
@@ -0,0 +1,202 @@
...
+                        };
+                    },
+                )
+                ->requiredMapping()
+                ->rules(
+                    [
+                        'starts_with:prospect:,student:',
+                    ]
+                )
+                ->example(fn (): string => 'student:' . Student::query()->value('email') ?? fake()->safeEmail()),
```
mxm1070, line 59: I believe this will fail in production because `composer install --no-dev` will not install faker.

mxm1070 (approved):

PR #122, opened 2023-09-29
[ASSIST-473]: Implement Task Importer

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/473/

### DIFF
--- a/app-modules/task/src/Imports/TaskImporter.php
+++ b/app-modules/task/src/Imports/TaskImporter.php
@@ -39,7 +39,7 @@ class TaskImporter extends Importer
                 ->example('pending'),
             ImportColumn::make('due')
                 ->rules(['date'])
-                ->example(fake()->dateTimeBetween('+1 day', '+1 week')->format('Y-m-d H:i:s')),
+                ->example('1990-01-01 00:00:00'),
             ImportColumn::make('assignedTo')
                 ->relationship(
                     resolveUsing: fn (mixed $state) => User::query()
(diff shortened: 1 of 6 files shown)


app-modules/task/src/Imports/TaskImporter.php
```diff
@@ -0,0 +1,131 @@
...
+                        };
+                    },
+                )
+                ->requiredMapping()
+                ->rules(
+                    [
+                        'starts_with:prospect:,student:',
+                    ]
+                )
+                ->example(fn (): string => 'student:' . Student::query()->value('email') ?? fake()->safeEmail()),
```
mxm1070, line 85: I believe this will fail in production because `composer install --no-dev` will not install faker.

app-modules/task/src/Imports/TaskImporter.php
```diff
@@ -0,0 +1,131 @@
...
+            ImportColumn::make('description')
+                ->requiredMapping()
+                ->example('A description of the task.'),
+            ImportColumn::make('status')
+                ->requiredMapping()
+                ->rules([new Enum(TaskStatus::class)])
+                ->example('pending'),
+            ImportColumn::make('due')
+                ->rules(['date_format:Y-m-d H:i:s'])
+                ->example(fake()->dateTimeBetween('+1 day', '+1 week')->format('Y-m-d H:i:s')),
```
mxm1070, line 42: I believe this will fail in production because `composer install --no-dev` will not install faker.

PR #123, opened 2023-09-29
[ASSIST-459] Investigate Filament Plugin Form Builder UI

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/459/

~~also https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/507/ because my UI changes were merged and I was able to move the Teams menu item.~~ Moved to #125


app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
```diff
@@ -0,0 +1,105 @@
...
+                FilamentTextInput::make('name')
+                    ->required()
+                    ->string()
+                    ->maxLength(255)
+                    ->autocomplete(false)
+                    ->columnSpanFull(),
+                FilamentTextarea::make('description')
+                    ->string()
+                    ->columnSpanFull(),
+                Builder::make('content')
```
Orrison, line 38: Should this be called `form_fields` or something like that?

app-modules/form/src/Filament/Resources/FormResource/Pages/CreateForm.php
```diff
@@ -0,0 +1,30 @@
...
+            ->schema([
+                TextInput::make('name')
+                    ->required()
+                    ->string()
+                    ->maxLength(255)
+                    ->autocomplete(false)
+                    ->columnSpanFull(),
+                Textarea::make('description')
+                    ->string()
+                    ->columnSpanFull(),
```
Orrison, line 27: Why do we not have the form field builder on create?

app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
```diff
@@ -0,0 +1,105 @@
...
+            ->toArray();
+
+        return $data;
+    }
+
+    protected function handleRecordUpdate(Model $record, array $data): Model
+    {
+        $record = parent::handleRecordUpdate($record, $data);
+
+        $record->items()->delete();
```
Orrison, line 75: Let's add a TODO here to look into versioning of form/form field configuration.

app-modules/form/src/Providers/FormServiceProvider.php
```diff
@@ -0,0 +1,57 @@
...
+    public function register()
+    {
+        Panel::configureUsing(fn (Panel $panel) => $panel->plugin(new FormPlugin()));
+    }
+
+    public function boot()
+    {
+        Relation::morphMap([
+            'form' => Form::class,
+            'form_item' => FormItem::class,
```
Orrison, line 25: I believe we need one for submissions as well.

app-modules/form/routes/web.php
```diff
@@ -0,0 +1,21 @@
...
+
+Route::middleware('web')
+    ->prefix('forms/embed')
+    ->name('forms.embed.')
+    ->group(function () {
+        // Route::get('/{form}', function (Form $form) {
+        //     dd($form->loadMissing('items'));
+        // })->name('show');
+
+        Route::get('/{embed}', EmbedForm::class)
```
Orrison, line 15: We could change `embed` for `form` here to make clearer what should be passed.

We may want to change `EmbedForm` to something like `RenderedForm` but not necessary.

app/Livewire/EmbedForm.php
```diff
@@ -0,0 +1,70 @@
...
+
+    public function form(FilamentForm $form): FilamentForm
+    {
+        $fields = $this
+            ->embed
+            ->items
+            ->map(fn ($item) => match ($item->type) {
+                'text_input' => \Assist\Form\Filament\Blocks\TextInput::display($item),
+                'text_area' => \Assist\Form\Filament\Blocks\TextArea::display($item),
+                'select' => \Assist\Form\Filament\Blocks\Select::display($item),
```
Orrison, line 37: Let's import these into `use` statements.

resources/views/components/layouts/app.blade.php
Orrison, line 1: What was the purpose in expecting out the app layout here?

app/Livewire/EmbedForm.php
```diff
@@ -0,0 +1,70 @@
...
+            })
+            ->toArray();
+
+        return $form
+            ->schema($fields)
+            ->statePath('data')
+            ->model($this->embed);
+    }
+
+    public function create(): void
```
Orrison, line 47: `submit()` may make more sense here.

Orrison (changes requested):
Looks great! Some comments and feedback.

It also looks like there are some linting error.

Tests are failing from timeout. In other PRs I have increased the Github Action timeout to 15 minutes. If you want to make that change here as well it may prevent tests from failing again.

app-modules/form/routes/web.php
```diff
@@ -0,0 +1,21 @@
...
+            ->name('show');
+
+        Route::post('/{form}', function (Request $request, Form $form) {
+            ray($request, $form);
+        })->name('update');
```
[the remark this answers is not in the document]
mxm1070: It was from before I decided to just submit the form in the component so I'll remove it.

app-modules/form/src/Filament/Blocks/Select.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        parent::setUp();
+
+        $this->label('Select');
+    }
+
+    public static function make(string $name = 'select'): static
+    {
+        return parent::make($name);
+    }
```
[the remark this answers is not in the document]
mxm1070: I can't make this abstract because it's overriding the base class. Instead I made an abstract method called type, and moved the make call into the base class which calls type. I can't call it name or key because these are also overriding the base class.

app-modules/form/src/Filament/Blocks/Block.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Assist\Form\Filament\Blocks;
+
+use Assist\Form\Models\FormItem;
+use Filament\Forms\Components\Field;
+use Filament\Forms\Components\Checkbox;
+use Filament\Forms\Components\Builder\Block as FilamentBlock;
+use Filament\Forms\Components\TextInput as FilamentTextInput;
+
+abstract class Block extends FilamentBlock
```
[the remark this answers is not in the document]
mxm1070: I renamed them for now, but I'm not entirely sure it's fine because there could be multiple fields within a block.

app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
mxm1070: `form_fields` feels a bit wordy to me, so I just went with fields, and I moved the builder into a section to differentiate it more.

image

app-modules/form/routes/web.php
mxm1070: I went through and changed it. I didn't do this before because it conflicts with the default behavior of Filament. I got around this, but it might make it slightly more cumbersome to work with.

resources/views/components/layouts/app.blade.php
mxm1070: It seems like Livewire and Filament want that file because the Component is not part of the Panels and it doesn't work without it: https://filamentphp.com/docs/3.x/forms/installation#configuring-your-layout

app-modules/form/src/Filament/Resources/FormResource/Pages/CreateForm.php
mxm1070: I thought it was confusing to do it all at once on create, but I've added it now.

Also, we're going to need a way to check for unique on the fields. I tried using both the unique rule and the rules call, but I can't figure it out because it needs to take the key and form_id into account, both on edit and on creation.

resources/lang/en/global.php
```diff
@@ -259,7 +259,7 @@
     'sample_answer' => 'Sample answer',
     'sample_category' => 'Sample category',
     'sample_question' => 'Sample question',
-    'select_all' => 'Select all',
+    'select_all' => 'SelectFormFieldBlock all',
```
Orrison, line 262: Looks like this was caught by mass find replace.

Orrison (changes requested):

PR #126, opened 2023-10-02
[ASSIST-429]: UI/UX: Artificial Intelligence Render Formatted Text 

Orrison:
Set's up the process to display formatted AI text via markdown that is sanitized.

During this process, I have started the first implementation of how we should be setting up JS implementations that rely on NPM packages.
@danharrin let me know if you think there is a better way.
The important things are that it is compiled so that we can make use of NPM packages without having to load them directly onto the pages. It is easy to configure. And that the JS is only ever loaded on the page it is needed.

### DIFF
--- a/app-modules/assistant/resources/views/filament/pages/ai-assistant.blade.php
+++ b/app-modules/assistant/resources/views/filament/pages/ai-assistant.blade.php
@@ -29,11 +29,20 @@ use Assist\Assistant\Services\AIInterface\Enums\AIChatMessageFrom;
                         </div>
                     @endforeach
                     @if ($showCurrentResponse)
-                        <div class="my-4 w-3/4 rounded-lg bg-gray-500 p-4 sm:p-6 lg:px-8" x-data="currentResponseData">
+                        <div
+                            class="my-4 w-3/4 rounded-lg bg-gray-500 p-4 sm:p-6 lg:px-8"
+                            x-data="currentResponseData"
+                        >
                             <h1 class="mb-1 text-2xl">AI Assistant</h1>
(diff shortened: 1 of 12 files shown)


danharrin (approved):
Might be worth reformatting the Blade as there were some indentation changes

app-modules/assistant/resources/views/filament/pages/ai-assistant.blade.php
```diff
@@ -43,64 +47,64 @@
...
                     @if (!$chat->id)
                         <button
-                            class="inline-flex cursor-pointer justify-center rounded-lg p-2 text-gray-500 hover:bg-gray-100 hover:text-gray-900 dark:text-gray-400 dark:hover:bg-gray-600 dark:hover:text-white"
-                            type="button"
-                            wire:loading.attr="disabled"
-                            wire:click="save"
+                                class="inline-flex cursor-pointer justify-center rounded-lg p-2 text-gray-500 hover:bg-gray-100 hover:text-gray-900 dark:text-gray-400 dark:hover:bg-gray-600 dark:hover:text-white"
+                                type="button"
+                                wire:loading.attr="disabled"
+                                wire:click="save"
```
danharrin, line 59: Is this correct?

Orrison: Do you mean the formatting, or something else?

danharrin: These changes indent the attributes an extra level, I didn't think this is how HTML was formatted in this app

danharrin (approved):

PR #137, opened 2023-10-04
[ASSIST-477] & [ASSIST-479]: Message Center with Threaded Timeline View

dgoetzit:
This PR introduces the core functionality of the Message Center, with a threaded timeline view of content when selecting an Educatable from the left-hand inbox. The Message Center creates a space where an authenticated user can see all Educatables who have "engaged" in any way (either from being engaged via an Engagement, or engaging the system via an EngagementResponse).

The left-hand inbox provides a linear history of outgoing/incoming engagements, with the most recently engaged Educatables showing first. The inbox also provides searching and light filtering capabilities in an attempt to scope down the results. Pagination is also available, but there is a known bug that I have added an item to the backlog for https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/573/.

There are some remaining TODOs which will be addressed in https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/481/, or moved into the backlog if we deem they aren't necessarily critical.

This PR also removes lock files from modules that still had them (as this was messing with my GOTO file in VSCode), and fixed a Task related bug introduced when "title" support was added: https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/569/


app-modules/assist-data-model/src/Models/Student.php
```diff
@@ -115,4 +116,11 @@ public static function filamentResource(): string
     {
         return StudentResource::class;
     }
+
+    protected function displayName(): Attribute
+    {
+        return Attribute::make(
+            get: fn (?string $value, array $attributes) => $attributes['full_name'],
```
Orrison, line 123: Do we want to make use of `displayNameKey` here?

app-modules/prospect/src/Models/Prospect.php
```diff
@@ -148,4 +149,11 @@ protected function serializeDate(DateTimeInterface $date): string
     {
         return $date->format(config('project.datetime_format') ?? 'Y-m-d H:i:s');
     }
+
+    protected function displayName(): Attribute
+    {
+        return Attribute::make(
+            get: fn (?string $value, array $attributes) => $attributes['full_name'],
```
Orrison, line 156: Same as for Student above

app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -145,6 +146,13 @@ public function createdBy(): BelongsTo
         return $this->belongsTo(User::class);
     }
 
+    // TODO We need a way to correspond custom statuses to some normalized data set
+    // So that we can determine what an "open" status is in the context of a client's data
```
Orrison, line 150: This will be very important before we have a folks using this I think. Do you know if we have a ticket for this already?

database/seeders/DatabaseSeeder.php
```diff
@@ -64,6 +64,7 @@ public function run(): void
             EngagementSeeder::class,
             EngagementResponseSeeder::class,
             DemoEducatableEngagementSeeder::class,
+            SuperAdminSeeder::class,
```
Orrison, line 67: Let's add this to the Demo seeder as well for now, so folks on Demo can see this functionality easily.

app-modules/engagement/resources/views/components/svg/calendar.blade.php
Orrison, line 1: Is it possible to use the filament icon component with the heroicon calendar instead? Or is there a particular reason we are using this one?

Orrison (changes requested):
A few, mostly non-blocking, questions/comments. There is also currently a merge conflict.

app-modules/service-management/src/Models/ServiceRequest.php
dgoetzit: I do not think so - I'll add one to the backlog.

app-modules/engagement/resources/views/components/svg/calendar.blade.php
dgoetzit: Ahh yeah - I just defaulted to pulling the svg from the design 😅

Orrison (approved):

PR #139, opened 2023-10-04
[ASSIST-535]: Remove display of GUID 'ID' from service management record view, [ASSIST-545]: Remove display of GUID 'ID' from audit list view and record view.

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/535/
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/545/

As per our discussion I replaced all the existing Id table columns with a new custom default hidden toggleable IdColumn, and added it for most models. I also removed the Id field from all infolists and forms.

### DIFF
--- a/app/Filament/Columns/IdColumn.php
+++ b/app/Filament/Columns/IdColumn.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace App\Filament\Columns;
+
+use Filament\Tables\Columns\TextColumn;
+
+class IdColumn extends TextColumn
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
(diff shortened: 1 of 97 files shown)


app/Filament/Columns/IdColumn.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->label('ID');
+
+        $this->sortable();
+
+        $this->toggleable(isToggledHiddenByDefault: true);
```
Orrison, line 17: Can these be chained?

Orrison (approved):

PR #142, opened 2023-10-04
[ASSIST-279]: Knowledge Base Orphan Media Management 

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/279/


danharrin (changes requested):
It might be worth adding some comments at various places in the MediaEncoder to give an example of input and output of each method, just so that the next person who works on this knows what's going on a bit better.

Other than that, looks good. But you do have some merge conflicts.

PR #150, opened 2023-10-05
[ASSIST-361]: Add PR templates to GitHub

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/361/

This adds the first version of a PR template for this repository. I can evolve over time as the project grows, but this should paint a clear picture of what we are already doing here.

@danharrin, @dgoetzit, @mxm1070, and @joelicatajr I think it would be best for everyone to review and approve if they are happy with this first pass.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,8 @@
+## Before contributing and submitting this PR, make sure you have:
+* [ ] Read the [contributing guidelines](https://github.com/canyongbs/assistbycanyongbs/blob/main/README.md#contributing).
+* [ ] Title the PR with the ticket/issue number and a short description of the changes made. Or if no ticket/issue exists, title the PR with a short description of the changes made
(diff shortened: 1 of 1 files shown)


dgoetzit (approved):
Looks good! One thing I'd like to see with this approach (but don't know if it needs to be documented in the template just yet) is the removal of all checkboxes that don't apply. 

This way, there aren't "unfinished" tasks in the PR when it's opened. There's really also no reason to see what the other options were.

Orrison:
> Looks good! One thing I'd like to see with this approach (but don't know if it needs to be documented in the template just yet) is the removal of all checkboxes that don't apply.
> 
> This way, there aren't "unfinished" tasks in the PR when it's opened. There's really also no reason to see what the other options were.

Ah yeah okay that is a good point, we can make sure we as a team do that now. And if it because an issue we can add a note in the template.

PR #155, opened 2023-10-06
[ASSIST-625]: Create new navigation item "Meeting Center" under Productivity Tools.

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/625/

### DIFF
--- a/app/Filament/Pages/Events.php
+++ b/app/Filament/Pages/Events.php
@@ -6,7 +6,7 @@ use Filament\Pages\Page;
 
 class Events extends Page
 {
-    protected static ?string $navigationIcon = 'heroicon-o-document-text';
+    protected static ?string $navigationIcon = 'heroicon-o-bell-alert';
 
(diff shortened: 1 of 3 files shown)


app/Filament/Pages/Events.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace App\Filament\Pages;
+
+use Filament\Pages\Page;
+
+class Events extends Page
+{
+    protected static ?string $navigationIcon = 'heroicon-o-document-text';
```
dgoetzit, line 9: This will most likely change with feedback from product, but for the time being, can we make this 
```suggestion
    protected static ?string $navigationIcon = 'heroicon-o-bell-alert';
```

dgoetzit (changes requested):
Looks good - just one minor requested change. Open to other ideas for the icon as well.

Orrison (approved):

PR #162, opened 2023-10-10
[ASSIST-623]: Audit gating of permissions

mxm1070:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/623/


app-modules/assistant/config/roles/web/assistant_log_management.php
dgoetzit, line 1: @Orrison What are your thoughts on a role surrounding chats themselves? Currently, this would effectively be accomplished by restricting access to the Assistant entirely.

app-modules/webhook/config/roles/web/webhook_management.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+return [
+    'model' => [
+        'inbound_webhook' => [
+            '*',
```
dgoetzit, line 6: This will ultimately come down to a product decision, but I don't think it should be possible to create/update data that normally originates from an external source. This role should probably be scoped down.

app-modules/timeline/src/Filament/Pages/Timeline.php
```diff
@@ -89,15 +90,21 @@ public function getSubNavigation(): array
 
     public static function shouldRegisterNavigation(array $parameters = []): bool
     {
+        if (auth()->user()->cannot('engagement.view-any')) {
+            return false;
+        }
```
dgoetzit, line 95: We might want to introduce a custom permission for this because the timeline itself consists of many different types of records, and it's possible a user can view some records but not all.

app-modules/notifications/config/roles/web/notifications_management.php
dgoetzit, line 1: This should probably be `subscription_management`

dgoetzit (changes requested):
Great work on this @mxm1070! Just a couple of comments/suggestions.

app-modules/assistant/config/roles/web/assistant_log_management.php
Orrison: @dgoetzit yeah I think the chat role/permissions should just be around just restricting access to the Assistant at all.

This permission though is for the audit records of User sent chat messages I believe. So we would want to restrict this based on whether or not an Admin has the rights to view these.

app-modules/assistant/src/Models/AssistantChatMessageLog.php
```diff
@@ -8,13 +8,15 @@
...
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Assist\Authorization\Models\Concerns\DefinesPermissions;
 
 /**
  * @mixin IdeHelperAssistantChatMessageLog
  */
 class AssistantChatMessageLog extends BaseModel
 {
     use MassPrunable;
+    use DefinesPermissions;
```
Orrison, line 19: Because these are never things a user would modify, create, or delete themselves, let's only generate the view and view any permissions.

See Role.

app-modules/assist-data-model/src/Policies/StudentPolicy.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Assist\AssistDataModel\Policies;
+
+use App\Models\User;
+use Illuminate\Auth\Access\Response;
+use Assist\AssistDataModel\Models\Student;
+
+class StudentPolicy
```
Orrison, line 9: Because these are never things a user would modify, create, or delete themselves, let's only generate the view and view any permissions.

See Role.

app-modules/assist-data-model/src/Policies/ProgramPolicy.php
Orrison, line 1: Because these are never things a user would modify, create, or delete themselves, let's only generate the view and view any permissions.

See Role.

app-modules/assist-data-model/src/Policies/PerformancePolicy.php
Orrison, line 1: Because these are never things a user would modify, create, or delete themselves, let's only generate the view and view any permissions.

See Role.

app-modules/assist-data-model/src/Policies/EnrollmentPolicy.php
Orrison, line 1: Because these are never things a user would modify, create, or delete themselves, let's only generate the view and view any permissions.

See Role.

app-modules/authorization/src/Filament/Resources/RoleGroupResource/RelationManagers/RolesRelationManager.php
```diff
@@ -51,11 +51,16 @@ public function table(Table $table): Table
...
+                // To the end user they are also removing all of the roles from the user
+                // That have been assigned to them through the RoleGroup
+                DetachAction::make()->label(function () {
+                    return 'Remove from ' . $this->ownerRecord->name . ' Role Group';
+                }),
             ])
             ->bulkActions([
                 BulkActionGroup::make([
-                    DeleteBulkAction::make(),
+                    // DetachBulkAction::make(),
```
Orrison, line 63: What is the purpose of commenting this out here?

app-modules/engagement/config/roles/web/engagement_management.php
```diff
@@ -8,5 +8,14 @@
...
         ],
+        'engagement_batch' => [
+            '*',
+        ],
+        'engagement_response' => [
+            '*',
+        ],
+        'engagement_file' => [
+            '*',
+        ],
```
Orrison, line 19: The fact that I named this "EngagementFile" was kind of a mistake. This is a separate concept from Engagements and is actually just Document storage on an Educatable.

Let's move this to be managed on it's own.

app-modules/knowledge-base/config/roles/web/knowledge_base_management.php
```diff
@@ -0,0 +1,18 @@
...
+        ],
+        'knowledge_base_category' => [
+            '*',
+        ],
+        'knowledge_base_quality' => [
+            '*',
+        ],
+        'knowledge_base_status' => [
+            '*',
+        ],
```
Orrison, line 16: These field property permissions should actually be separated away from the base Model CRUD access. As we will have Users that can access the KB for example, but they are not admins that should be able to change these field settings.

@dgoetzit what do you think about the naming for this separation? `knowledge_base_management` and `knowledge_base_field_settings_management`?

We will need to do this also for others like Prospects and Service Requests.

app-modules/notifications/config/roles/web/notifications_management.php
Orrison: I agree

app-modules/service-management/config/roles/web/service_request_management.php
```diff
@@ -0,0 +1,21 @@
...
+        ],
+        'service_request_status' => [
+            '*',
+        ],
+        'service_request_type' => [
+            '*',
+        ],
+        'service_request_update' => [
+            '*',
+        ],
```
Orrison, line 19: Same note as the KB one above.

app-modules/webhook/config/roles/web/webhook_management.php
Orrison: I agree, much like the others I mentioned above, this probably only needs view and view-any

Orrison (approved):

Orrison:
@dgoetzit if you can give this one last look when you can and resolve your comments.

dgoetzit (approved):

PR #164, opened 2023-10-10
[ASSIST-633]: Bug: Resolve view of student/prospect subscriptions 

Orrison:
https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/633/

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -14,6 +14,7 @@ use Assist\Authorization\Models\Role;
 use Illuminate\Notifications\Notifiable;
 use OwenIt\Auditing\Contracts\Auditable;
 use Assist\Assistant\Models\AssistantChat;
+use Assist\AssistDataModel\Models\Student;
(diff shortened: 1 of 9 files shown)


app/Models/User.php
```diff
@@ -102,6 +105,26 @@ public function subscriptions(): HasMany
...
     }
 
+    public function prospectSubscriptions(): MorphToMany
+    {
+        return $this->morphedByMany(
+            related: Prospect::class,
+            name: 'subscribable',
+            table: 'subscriptions'
+        )
+            ->using(Subscription::class);
```
dgoetzit, line 115: We don't necessarily need it at this point, but I'd recommend adding the `withPivot()` and `withTimestamps()` to these inverse relationships as well.

dgoetzit (approved):
Looks good! Left a non-blocking suggestion.

dgoetzit (approved):

PR #166, opened 2023-10-10
[ASSIST-619]: Set new primary color usage

dgoetzit:
This PR changes the primary color for light and dark mode, using the `trout` and `bright-sun` palettes. This currently utilizes the approach of passing our palettes into Filament, and not overriding classes via CSS Hooks.


resources/css/filament/admin/theme.css
```diff
@@ -7,3 +7,106 @@
...
+    --primary-400: 254, 195, 33;
+    --primary-500: 248, 162, 8;
+    --primary-600: 219, 122, 4;
+    --primary-700: 182, 86, 7;
+    --primary-800: 148, 66, 12;
+    --primary-900: 121, 55, 14;
+    --primary-950: 70, 27, 2;
+}
+
+/* Style Overrides */
```
Orrison, line 26: Do we want to keep this commented out code for now?

dgoetzit: The overrides? Yes. I've brought up the conversation that we had yesterday with Product in the ticket, and we'll see which way we need to go. 

With the approach taken, the overrides are mostly redundant, but I'm leaving them here in the event that we do want to introduce overrides in specific cases.

Orrison (approved):

PR #241, opened 2023-11-02
fix: Campaign action column

danharrin:
Tests on #240 are failing because I think this column has the wrong name?


app-modules/campaign/src/Models/CampaignAction.php
```diff
@@ -81,7 +81,7 @@ public function scopeCampaignEnabled(Builder $query): void
 
     public function hasBeenExecuted(): bool
     {
-        return (bool) ! is_null($this->executed_at);
```
danharrin, line 84: so it doesnt get lost in the diff... this one

dgoetzit (approved):

PR #268, opened 2023-11-13
[ASSIST-685]: Make API call to fetch Azure profile picture on SSO

mxm1070:
### Ticket(s) or GitHub Issue

https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/685/

### Technical Description

Make API call to fetch Azure profile picture on SSO.

### Types of changes

- [x] No

### DIFF
--- a/app-modules/authorization/src/Http/Controllers/SocialiteController.php
+++ b/app-modules/authorization/src/Http/Controllers/SocialiteController.php
@@ -3,10 +3,12 @@
 namespace Assist\Authorization\Http\Controllers;
 
 use App\Models\User;
+use Illuminate\Support\Str;
 use Illuminate\Http\Request;
 use Filament\Facades\Filament;
 use App\Http\Controllers\Controller;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\Http;
 use Filament\Notifications\Notification;
 use Assist\Authorization\Enums\SocialiteProvider;
 
@@ -33,6 +35,7 @@ class SocialiteController extends Controller
             ->setConfig($provider->config())
             ->user();
 
+        /** @var User $user */
         $user = User::query()
             ->where('email', $socialiteUser->getEmail())
             ->first();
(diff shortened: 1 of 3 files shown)


app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -46,9 +49,18 @@ public function callback(SocialiteProvider $provider)
...
 
+        if ($provider === SocialiteProvider::Azure) {
+            $request = Http::withToken($socialiteUser->token)
+                ->contentType('image/jpeg')
+                ->get('https://graph.microsoft.com/v1.0/me/photo/$value');
+
+            $user->addMediaFromString($request->body())->usingFileName(Str::uuid() . '.jpg')->toMediaCollection('avatar');
+        } else {
+            $user->addMediaFromUrl($socialiteUser->getAvatar())->toMediaCollection('avatar');
+        }
```
Orrison, line 60: After thinking about this a little more, I think if we DO have the image as a URL we should use the URL rather than download the image to save on storage space in our application.

We could change this here into a match statement and if it Azure then save to the media collection, if it is Google then we can save the avatar URL in the users table column.

Then add something like a "getAvatar" method on the User model that returns either the URL saved in the DB or, if that is null, a generated URL for the media item.

That way we are taking advantage of external hosting of the asset when we can, saving some resource cost on our end.

Orrison: `default` of the match could just return `null` to prevent errors just in case.

Orrison (changes requested):
Realized we could handle this in a more resource-saving way. Let me know if you have any questions or thoughts on it.

app-modules/authorization/src/Http/Controllers/SocialiteController.php
mxm1070: I reintroduced the url behavior. There's no reason to have the match with just the one option here because I doubt we will have other SSO providers that do not return a url. Just checking the property is a trivial operation and trying to split the update will just result in more boilerplate.

Orrison (approved):

PR #287, opened 2023-11-20
[ASSIST-893]: New Campaign Type to Subscribe 1 or more Users to Prospect or Students

mxm1070:
### Ticket(s) or GitHub Issue

https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/893/

### Technical Description

New Campaign action to subscribe one or more Users to Prospect or Students.

### Types of changes

- [x] Documentation

- [x] No


app-modules/notifications/src/Models/Subscription.php
```diff
@@ -40,6 +45,35 @@ public function subscribable(): MorphTo
...
 
+    public static function executeFromCampaignAction(CampaignAction $action): bool|string
+    {
+        try {
+            DB::beginTransaction();
+
+            $action->campaign->caseload->retrieveRecords()->each(function (Subscribable $subscribable) use ($action) {
+                if ($action->data['remove_prior']) {
+                    $subscribable->subscriptions()->delete();
+                }
```
Orrison, line 56: I asked a question to Product in this ticket: https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/893/?IFRAME=Y&IFRAME_TYPE=SIDE_SLIDER&MID=3349#com3349

I think it may be that this setting is supposed to clear the Subscriptions of the selected Users, not the Subscribables in the Caseload.

Orrison (commented):
I think this looks pretty good. But I would like @dgoetzit's opinion on the refactor here.

@dgoetzit if you can approve or request changes.

The implementation of the execute for Subscriptions looks good for now. But in the fault tolerance ticket we have coming up we should refactor all of these to make sure they are batched up into jobs.

mxm1070:
@dgoetzit If you approve of this refactor, I would like to do one further change now. I want to put
```php
public static function blocks(): array
{
    return [
        EngagementBatchBlock::make(),
        ServiceRequestBlock::make(),
        ProactiveAlertBlock::make(),
        InteractionBlock::make(),
        CareTeamBlock::make(),
        TaskBlock::make(),
        SubscriptionBlock::make(),
    ];
}
```
into the enum as it's the only piece left that requires you to touch a separate file.

resources/views/filament/forms/components/campaigns/actions/subscription.blade.php
```diff
@@ -0,0 +1,35 @@
+@php
+    use Carbon\Carbon;
+    use App\Models\User;
+    use Assist\Division\Models\Division;
+    use Assist\Interaction\Models\InteractionType;
+    use Assist\Interaction\Models\InteractionStatus;
+    use Assist\Interaction\Models\InteractionDriver;
+    use Assist\Interaction\Models\InteractionOutcome;
+    use Assist\Interaction\Models\InteractionCampaign;
+    use Assist\Interaction\Models\InteractionRelation;
```
dgoetzit, line 10:
```suggestion
    
```

dgoetzit (commented):
@mxm1070 This looks great - please just remove the unused imports from the specified file.

As far as the extraction of `blocks` into the CampaignActionType, that sounds good to me in order to maintain this pattern now.

app-modules/notifications/src/Models/Subscription.php
Orrison: Product has confirmed it is indeed supposed to clear the Subscribers of the population.

mxm1070:
I'm wrapping this up but I have a couple notes for future refactors like the fault tolerance ticket.

Right now a lot of the of the settings in the enum should probably references properties or functions on the block instead of being hardcoded in the enum. For example to change label on a block you must currently change it in 3 places: the enum getLabel, the block setUp, and the block summary blade file. This should be centralized to the getLabel function of the block. The getModel enum function should probably reference the model property of the block. The getStepSummaryView enum function should reference a summaryView property on the block. Lastly the executeAction enum function should be moved into the block or a standalone action class. In this way the block files themselves become the source of truth for all of their data.

All blocks also share an `execute_at` datetime field so this should probably be moved up into `CampaignActionBlock` class so that all blocks share the field and devs do not have to remember to add the field.

dgoetzit (approved):

PR #293, opened 2023-11-22
[ASSIST-913]: Outlook Calendar Synchronization 

Orrison:
### Ticket(s) or GitHub Issue

- https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/913/

### Technical Description

Adds Microsoft Outlook calendar syncing.

### Types of changes

- [x] Yes, please specify

You will need to make sure your `.env` `AZURE_...` credentials are set.


app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
```diff
@@ -31,65 +31,281 @@
...
+            try {
+                $response = $request->execute();
+            } catch (ClientException $exception) {
+                if ($exception->getCode() === 401) {
+                    $calendar = $this->refreshToken($calendar);
 
-        return [];
+                    $request->setAccessToken($calendar->oauth_token);
+
+                    $response = $request->execute();
```
mxm1070, line 93: You do this check throughout this file but is it necessary? In makeClient you already check and refresh the token so if it 401s again why try to refresh a second time?

If you do keep it what about `Symfony\Component\HttpFoundation\Response::HTTP_UNAUTHORIZED` instead of 401?

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
```diff
@@ -31,65 +31,281 @@
...
+                'client_id' => config('services.azure_calendar.client_id'),
+                'client_secret' => config('services.azure_calendar.client_secret'),
+                'grant_type' => 'refresh_token',
+                'scope' => ['Calendars.ReadWrite', 'User.Read', 'offline_access'],
+                'refresh_token' => $calendar->oauth_refresh_token,
+            ]
+        );
+
+        if ($response->clientError() || $response->serverError()) {
+            if ($response->status() === 401) {
```
mxm1070, line 248: What about `Symfony\Component\HttpFoundation\Response::HTTP_UNAUTHORIZED` instead of 401?

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
```diff
@@ -31,65 +31,281 @@
...
+                if ($userEvent->isDirty()) {
+                    $userEvent->saveQuietly();
+                }
+            });
+
+        $calendar->events()
+            ->whereNull('provider_id')
+            ->each(fn ($event) => $this->createEvent($event));
+
+        $calendar->events()
```
mxm1070, line 222: I can't quite tell, will this delete past events like the problem we had with Google?

mxm1070 (changes requested):

app-modules/meeting-center/src/Filament/Resources/CalendarEventResource/Pages/ListCalendarEvents.php
```diff
@@ -156,7 +159,13 @@ public function table(Table $table): Table
...
                 ]),
             ])
-            ->modifyQueryUsing(fn (Builder $query) => $query->orderBy('starts_at'));
+            ->modifyQueryUsing(function (Builder $query): Builder {
+                /** @var User $user */
+                $user = auth()->user();
+
+                return $query->whereRelation('calendar', 'user_id', $user->id);
+            })
+            ->defaultSort('starts_at', 'desc');
```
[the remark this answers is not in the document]
dgoetzit: Would this have previously been scoped to the authenticated user's calendar events?

app-modules/meeting-center/src/Http/Controllers/OutlookCalendarController.php
```diff
@@ -31,18 +31,51 @@
...
+        /** @var User $socialiteUser */
+        $socialiteUser = $driver
+            ->setConfig(SocialiteProvider::AzureCalendar->config())
+            ->user();
+
+        $user = auth()->user();
+
+        $calendar = $user->calendar ?: new Calendar();
+
+        ray($socialiteUser);
```
dgoetzit, line 69: This can be removed

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
Orrison: It will not since we return all events within the time span we set. So it will get the events in the past as well here and not delete them.

app-modules/meeting-center/src/Filament/Resources/CalendarEventResource/Pages/ListCalendarEvents.php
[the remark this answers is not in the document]
Orrison: So the change to `modifyQueryUsing` now makes it so you only see your events, which I believe is what Joe expects.

I moved the sorting stuff to `defaultSort` because, though what you had works, it would ALWAYS make that sorting happing, so clicking the other sorting buttons did nothing to the list.

This makes it so starts_at desc is the default, but a user COULD change it if they wanted.

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
dgoetzit: It seems possible that the oauth token could expire after instantiating the client initially, so it probably makes sense to do this at some level. But, it might make sense for us to abstract this so it's done behind the scenes per request execution. That way we don't need to remember to do it in every instance.

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
```diff
@@ -31,65 +31,281 @@
...
 
     public function getEvents(Calendar $calendar, ?Datetime $start = null, ?Datetime $end = null, ?int $perPage = null): array
     {
-        // https://learn.microsoft.com/en-us/graph/api/user-list-calendars?view=graph-rest-1.0&tabs=http
-        // https://learn.microsoft.com/en-us/graph/api/user-list-events?view=graph-rest-1.0&tabs=http
+        $client = $this->makeClient($calendar);
+
+        $start = $start ?? now()->subYear()->startOfDay();
+
+        $end = $end ?? now()->addYear()->endOfDay();
```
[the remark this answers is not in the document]
Orrison: I did change my mind here while working on this one, I'll change Google to a year as well.

Since I was testing this with my real Canyon calendar, two years brings in a massive amount of events into the app and takes a little longer to sync.

dgoetzit (changes requested):
I think this looks great - just a `ray()` call to remove.

app-modules/meeting-center/src/Managers/OutlookCalendarManager.php
Orrison: The check of whether or not we think the `access_token` is expired based on the time we stored it does happen in `makeClient`, but it is possible for something to go wrong with that instance of the access_token or Microsoft expires it for some reason. So these checks ensure that if we get a failure, we try and refresh.

Orrison:
> It seems possible that the oauth token could expire after instantiating the client initially, so it probably makes sense to do this at some level. But, it might make sense for us to abstract this so it's done behind the scenes per request execution. That way we don't need to remember to do it in every instance.

@dgoetzit, yeah I mulled it over in my head the best way to handle these without rebuilding the client. Because V2 of this SDK handles this for you we can wait till we use that. But unfortunately it is currently not at a place we can make use of it.

dgoetzit (approved):

mxm1070 (dismissed):

mxm1070 (approved):

PR #307, opened 2023-11-29
[ADVISINGAPP-977]: Introduce "Applications" Under Manage Admissions

Orrison:
### Ticket(s) or GitHub Issue

- https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/977/

### Technical Description

This adds Application as a new entity within the application, it is essentially a duplicate for Forms that has been split off and adhering to the new abstract classes that govern the implementation and generation of Forms.

The abstract classes provide and enforce certain columns by use of specifying getters via Laravel Attributes. After some back and forth I figured this was the best way to enforce these similar columns/properties while still keeping close to Laravel conventions. Creating `get___` methods resulted in there being two different ways one could retrieve the data, which I had a concern with.

Finally, you will notice that Applications are using the Form widget for now. This was intentional, and there is a separate ticket to consider splitting off a separate widget. But since this widget is currently in its infancy, especially when it comes to styling, and since the functionality as it stands now is almost identical, I held off on splitting until necessary.

### Types of changes

- [x] No


app-modules/application/src/Filament/Resources/ApplicationResource/Pages/Concerns/HasSharedFormConfiguration.php
```diff
@@ -0,0 +1,231 @@
...
+            $field->is_required = $isRequired ?? false;
+            $field->type = $componentAttributes['type'];
+            $field->config = $componentAttributes['data'];
+            $field->save();
+
+            $components[$componentKey]['attrs']['id'] = $field->id;
+        }
+
+        return $components;
+    }
```
danharrin, line 205: I think this could be shared between modules potentially

danharrin (changes requested):
Also merge conflicts

app-modules/application/src/Filament/Resources/ApplicationResource/Pages/Concerns/HasSharedFormConfiguration.php
Orrison: I think that the configuration of Form and Application will start to vary more and more. It already does so as Application does not have a toggle for "authenticated" because they are authenticated by default.

So leaving them as separate I feel, may be in our best interest.

PR #330, opened 2023-12-05
[ADVISING-931]: Request Workflow for Prospect or Student Form Submission

danharrin:
### Ticket(s) or GitHub Issue

- https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/931

### Technical Description

Exactly as described in the ticket. It seems that a UI for tracking the status of requests is outside of the scope.

When submitting a form after it was requested, the submission is attached to the request, which represents the request as "submitted".

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Content or styling update (Changes which don't affect functionality)
- [ ] DevOps
- [ ] Documentation

https://github.com/canyongbs/advisingapp/assets/41773797/285f0f16-85ca-4daf-99fc-4c6813bdcadf

- [ ] Yes, please specify
- [x] No


app-modules/form/src/Actions/DeliverFormRequestBySms.php
```diff
@@ -0,0 +1,63 @@
...
+                    'from' => config('services.twilio.from_number'),
+                    'body' => "You have been sent a request to complete {$this->request->form->name} by {$this->request->user->name}." .
+                        (filled($this->request->note) ? " {$this->request->note}" : '') .
+                        ' ' . route('forms.show', ['form' => $this->request->form]),
+                ],
+            );
+        } catch (TwilioException $e) {
+            dd($e);
+            // TODO Notify someone of the failure
+        }
```
Orrison, line 61: We possibly don't have to do this right now. But what are your thoughts on creating/installing a Notification channel for Twilio?

That way, we can have a single Notification class for this type of interaction that we just define the `via` for.

There is this: https://laravel-notification-channels.com/twilio/#contents
Or we could possibly roll our own if need be.

app-modules/form/src/Models/FormRequest.php
```diff
@@ -0,0 +1,92 @@
...
+use App\Models\BaseModel;
+use Illuminate\Database\Eloquent\Builder;
+use Assist\Form\Enums\FormRequestDeliveryMethod;
+use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @mixin IdeHelperFormRequest
+ */
+class FormRequest extends BaseModel
```
Orrison, line 49: This may seem redundant since we track the `user_id` that created this. But can we make this Model Auditable?

app-modules/form/src/Models/FormRequest.php
```diff
@@ -0,0 +1,92 @@
...
+
+    public function deliver(): void
+    {
+        $this->method->deliver($this);
+    }
+
+    public function scopeNotCanceled(Builder $query): Builder
+    {
+        return $query->whereNull('canceled_at');
+    }
```
Orrison, line 91: We may want to create something like an ENUM for `FormSubmissionStatus` that has `open`, `canceled`, and `submitted` as options with a method here or in the ENUM that does the logic determination of what the current status is.

Else, we could end up doing that check-in multiple places in the app.

app-modules/form/src/Notifications/FormRequestNotification.php
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Assist\Form\Notifications;
+
+use Illuminate\Bus\Queueable;
+use App\Notifications\MailMessage;
+use Assist\Form\Models\FormRequest;
+use Illuminate\Notifications\Notification;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class FormRequestNotification extends Notification implements ShouldQueue
```
Orrison, line 45: Let's add a resolveNotificationSetting method to this, similar to other Notifications.

Orrison (changes requested):
Looks great! Just left some feedback and questions.

app-modules/form/src/Actions/DeliverFormRequestBySms.php
danharrin: I agree with adding it to the backlog. Currently, I have copied how it is done for engagements.

For engagements, we are tracking the status of the message from Twilio, and I don't know if a channel allows the flexibility required to do that. I'm unsure how you would get the engagement object from the channel.

app-modules/form/src/Models/FormRequest.php
danharrin: Done

app-modules/form/src/Models/FormRequest.php
danharrin: To be honest, I disagree. Currently, the status is calculable based on the `canceled_at`, and the `submission_id`. Adding in an extra column would represent that same information again, requiring it to be kept up to date. Also, the checks for these (two `whereNull()` column checks) are so simple that I don't even think it would be inconvenient?

app-modules/form/src/Notifications/FormRequestNotification.php
danharrin: Done

app-modules/form/src/Models/FormRequest.php
Orrison: I don't think we should add a column. Simply a method that returns the status of the `FormRequest` as an ENUM based on the logic you described. So that if we need to check the status of a `FormRequest` there is a central place that is determined.

danharrin: Done

PR #335, opened 2023-12-06
[ADVISING-1083]: API Authorization

mxm1070:
### Ticket(s) or GitHub Issue

https://engage.canyongbs.com/workgroups/group/3/tasks/task/view/1083/

### Technical Description

API Authorization with permissions.

### Types of changes

- [x] No


app-modules/authorization/database/factories/RoleFactory.php
```diff
@@ -48,7 +48,7 @@ public function definition(): array
     {
         return [
             'name' => fake()->text(25),
-            'guard_name' => fake()->randomElement(['web', 'api']),
+            'guard_name' => 'web',
```
Orrison, line 51: Did you make this change because you need some seeder to be `web`?

If that is the case, let's leave this as the `randomElement` generation and create two factory states that set this as `web` or `api` and use that where you require it to be `web`

app-modules/authorization/database/factories/RoleGroupFactory.php
```diff
@@ -36,17 +36,19 @@
...
- * @extends \Illuminate\Database\Eloquent\Factories\Factory<\Assist\Authorization\Models\RoleGroup>
+ * @extends Factory<RoleGroup>
  */
 class RoleGroupFactory extends Factory
 {
     public function definition(): array
     {
         return [
-            'name' => fake()->catchPhrase,
+            'name' => fake()->word(),
```
Orrison, line 50: Why the change from `catchPhrase` to `word()` I feel like you and @dgoetzit keep swapping it each time y'all are in this file XD

app-modules/authorization/config/roles/api/super_admin.php
Orrison, line 1: Can this file just be deleted so that this role does not exist?

app/Filament/Resources/SystemUserResource/Pages/CreateSystemUser.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function afterCreate(): void
+    {
+        /** @var SystemUser $systemUser */
+        $systemUser = $this->getRecord();
+
+        $token = str($systemUser->createToken('api')->plainTextToken)->after('|')->toString();
+
+        session()->put('apiToken', $token);
```
Orrison, line 65: Giving this some more thought I worry about the security implications of having the API token be stored in plain text somewhere. Even if it is the session for split seconds.

If there is no more secure way to do this within the request lifecycle of the view/edit page, then we may just want to have generating it be an Action or something like that on the view/edit page.

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
```diff
@@ -0,0 +1,58 @@
...
+        Schema::create('personal_access_tokens', function (Blueprint $table) {
+            $table->id();
+
+            $table->uuidMorphs('tokenable');
+            $table->string('name');
+            $table->string('token', 64)->unique();
+            $table->text('abilities')->nullable();
+
+            $table->timestamp('last_used_at')->nullable();
+            $table->timestamp('expires_at')->nullable();
```
Orrison, line 53: What is the current default expire time for these? And should we possibly display that on the SystemUser view/edit page?

Orrison (changes requested):

app-modules/authorization/database/factories/RoleGroupFactory.php
mxm1070: I changed it because it kept telling me the method doesn't exist. I double checked the docs and it does, and it works, so I'll put it back.

app-modules/authorization/database/factories/RoleFactory.php
mxm1070: The guard can't be api anymore because of the `config/auth.php` change to the api guard to be provided by `system-users`. Otherwise you get things like this:

```
 FAILED  Appmodules\authorization\tests\Feature\Actions\AddNewRoleToExistingUsersInRoleGroupTest > it will not overwrite an existing role that was directly assigned to the user              GuardDoesNotMatch   
  The given role or permission should use guard `web` instead of `api`.
```

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
mxm1070: There is no default expires at.

PR #342, opened 2023-12-07
hotfix: Forms

danharrin:
As featured in the demo ✨


mxm1070 (approved):

Orrison:
@danharrin is this still needed at all? If not click the delete branch button if you could.

danharrin:
I made this PR so it could be fixed on staging, but it doesn't look like that's a priority so I'm just making these changes in 1113 instead.

PR #351, opened 2023-12-12
Exploration of Lighthouse API Implementation

Orrison:
### Ticket(s) or GitHub Issue

- N/A

### Technical Description

Small Implementation of the GraphQL API with Laravel Lighthouse.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Content or styling update (Changes which don't affect functionality)
- [ ] DevOps
- [ ] Documentation

- [x] Yes, please specify
- [ ] No

We will need to adjust the deployment scripts to call the command to generate the API docs.


app/GraphQL/Directives/AfeatureDirective.php
```diff
@@ -0,0 +1,88 @@
...
+  feature: [String!]
+) repeatable on FIELD_DEFINITION | OBJECT
+GRAPHQL;
+    }
+
+    public function manipulateTypeDefinition(DocumentAST &$documentAST, TypeDefinitionNode &$typeDefinition): void
+    {
+        $feature = $this->directiveArgValue('feature');
+        ray($documentAST);
+        ray($typeDefinition);
```
dgoetzit, line 70: These can be removed.

app/GraphQL/Directives/AfeatureDirective.php
```diff
@@ -0,0 +1,88 @@
...
+
+        unset($documentAST->types[$typeDefinition->name->value]);
+        //ASTHelper::addDirectiveToFields($this->directiveNode, $typeDefinition);
+    }
+
+    public function manipulateTypeExtension(DocumentAST &$documentAST, TypeExtensionNode &$typeExtension): void
+    {
+        ray($documentAST);
+        ray($typeExtension);
```
dgoetzit, line 79: These can be removed.

dgoetzit (commented):
Looks great as a POC!

Orrison:
@dgoetzit I have adjusted based on your feedback, finished the `@feature` directive, and improved the implementation with prettier and auto-importing from Modules.

mxm1070:
I like your auto-importer. If we wanted to import multiple schemas in a module would you prefer we just call `discoverSchema` multiple times or should we be able to pass it an array?

Although I wonder with the auto-importer do we need the `@feature` directive? Couldn't we just not discover the schema?

Orrison:
> I like your auto-importer. If we wanted to import multiple schemas in a module would you prefer we just call `discoverSchema` multiple times or should we be able to pass it an array?
> 
> Although I wonder with the auto-importer do we need the `@feature` directive? Couldn't we just not discover the schema?

You could do it one of two ways.
1. Call the import method multiple times
2. `#import` the other schemas within the main root schema named after the module. See the Notifications one.

Either is fine, but I like #2.

We could do that here. But the `@feature` directive works on types, type extensions, and fields. So we could disable a particular Query field based on a feature, while still having the rest of the file.

app-modules/knowledge-base/graphql/knowledge-base-item.graphql
```diff
@@ -0,0 +1,12 @@
+type KnowledgeBaseItem
+    @feature(feature: "knowledge-management")
+    @model(class: "AdvisingApp\\KnowledgeBase\\Models\\KnowledgeBaseItem") {
```
mxm1070, line 3: This seems to be the only `@model` that uses the FQN. Was there a reason for this or can it be switched to `knowledge_base_item`?

mxm1070 (changes requested):

app-modules/knowledge-base/graphql/knowledge-base-item.graphql
Orrison: For some reason it would not resolve the morphmap. Perhaps something with it being snake_case. Right now I think our best option is to define all the models with FQN. That is what I am doing in my current branch, and we can update the others to do the same.

Orrison: Perhaps we can figure out the morphmap thing later. But defining the FQN on the Model I think is sufficient for now. Spending time trying to get Morphmap to work would not be worth the effort atm.

mxm1070 (approved):

PR #417, opened 2024-01-04
[ADVAPP-134]: Implement Likert Scale Survey Question Type

mxm1070:
### Ticket(s) or GitHub Issue

https://canyongbs.atlassian.net/browse/ADVAPP-134

### Technical Description

Implement Likert scale survey question type.

### Types of changes

- [x] No


app-modules/form/src/Actions/GenerateFormKitSchema.php
```diff
@@ -58,21 +57,19 @@ public function __invoke(Submissible $submissible): array
         ];
     }
 
-    public function content(array $content, ?Collection $fields = null): array
+    public function content(array $blocks, array $content, ?Collection $fields = null): array
```
danharrin, line 60: Maybe the blocks should be passed after the content? Feels secondary to me

danharrin (changes requested):
Would it be much trouble to attach a screen recording of this feature in use, from adding it to an editor instance, to submitting a form with it, to viewing it in the infolist of a submission?

app-modules/form/src/Actions/GenerateFormKitSchema.php
mxm1070: I disagree because of the recursion. In my opinion the parameters at the front should be the static parameters that are used in each iteration, followed by the parameters that change each iteration, both with respect to decreasing importance.

mxm1070:
> Would it be much trouble to attach a screen recording of this feature in use, from adding it to an editor instance, to submitting a form with it, to viewing it in the infolist of a submission?

Sure, here you go!


https://github.com/canyongbs/advisingapp/assets/11588575/982f8cb9-5da0-426b-895b-733e68624655

Orrison (approved):

app-modules/form/src/Actions/GenerateFormKitSchema.php
danharrin: `$fields` is static, I think, and it's after content. But sure, lets just keep it like this for now, it's just a minor thing.

mxm1070: Actually you're right. I assumed because it was optional that it was changing somewhere.

danharrin: It's set on the first iteration if it is not present, and optional params need to be at the end of the list

PR #434, opened 2024-01-10
[ADVAPP-206]: Display device age as a calculated column in years and months in assets

mxm1070:
### Ticket(s) or GitHub Issue

https://canyongbs.atlassian.net/browse/ADVAPP-206

### Technical Description

Display device age as a calculated column in years and months for assets.

### Types of changes

- [x] No


app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ListAssets.php
```diff
@@ -65,8 +68,27 @@ public function table(Table $table): Table
...
+                        /** @var User $user */
+                        $user = auth()->user();
+
+                        $diff = $date
+                            ->roundMonth()
+                            ->setTimezone($user->timezone)
+                            ->diff();
+
+                        return $diff->y . ' ' . ($diff->y === 1 ? 'Year' : 'Years') . ' ' .
+                            $diff->m . ' ' . ($diff->m === 1 ? 'Month' : 'Months');
```
dgoetzit, line 89: Can we extract this functionality?

dgoetzit (changes requested):

app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ListAssets.php
mxm1070: Sure, to what? A field? A macro?

dgoetzit: This is probably one of the few cases that I'd say an attribute actually might make the most sense, unless we foresee this potentially being applicable to other models/fields in the future.

mxm1070: Hmm, would an attribute work? The actual field is still `purchase_date` because we need to sort on that so if I put that into an attribute would that mess with the sorting?

mxm1070: Actually I see you can pass the column to sort by so it might work.

dgoetzit: We might not need to do that though, I think you can make the accessor something like `formatted_purchase_date`, access the record inside of the closure in order to access this property. Sort would still be on db column.

dgoetzit (approved):

PR #437, opened 2024-01-11
Feature/quality of life changes

Orrison:


mxm1070:
If we're not going to do the Google cleanup anymore then maybe also remove 
```
"extra": {
        ...,
        "google/apiclient-services": [
            "Oauth2",
            "Calendar"
        ]
    },
```
because it won't actually do anything now.

danharrin (approved):

Orrison:
> If we're not going to do the Google cleanup anymore then maybe also remove
> 
> ```
> "extra": {
>         ...,
>         "google/apiclient-services": [
>             "Oauth2",
>             "Calendar"
>         ]
>     },
> ```
> 
> because it won't actually do anything now.

We are going to run the cleanup in deployment so I think these still need to be there.

PR #518, opened 2024-02-09
[ADVAPP-267]: Enhance delete or archive capabilities for service management

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-267

### Technical Description

Introduce archiving capability for Service Request Types, Service Request Statuses, Change Request Types, and Change Request Statuses.

### Types of changes

- [x] No

* [x] Rebased your branch PR on top of the latest upstream `main` branch.


app/Filament/Filters/ArchivedFilter.php
danharrin, line 1: I would advise putting this in `Filament\Tables\Filters` as per #519

danharrin (changes requested):
Refs to IdColumn will need to be updated if #519 is merged

Orrison (changes requested):
Hey @mxm1070, I am really sorry. This was on me for not updating the ticket based on the conversation during Sprint Planning. And without having done so I see how the ticket is confusing.

We don't want to add a new column or pattern for archiving something. The soft delete `deleted_at` column should be used for this.

I'll make sure going forward we have the tickets updated during/after or sprint/tech plannings.

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/EditServiceRequest.php
```diff
@@ -68,9 +71,10 @@ public function form(Form $form): Form
                 Select::make('status_id')
                     ->relationship('status', 'name')
                     ->label('Status')
-                    ->options(fn () => ServiceRequestStatus::optionsByClassification())
+                    ->options(fn () => ServiceRequestStatus::optionsByClassification(true))
```
danharrin, line 74: I like to use named parameters when the values aren't immediately obvious what you're passing in. What do you think?

danharrin (changes requested):

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/EditServiceRequest.php
mxm1070: I ended up removing the function.

app-modules/service-management/src/Observers/ChangeRequestTypeObserver.php
```diff
@@ -0,0 +1,49 @@
...
+use AdvisingApp\ServiceManagement\Models\ChangeRequestType;
+
+class ChangeRequestTypeObserver
+{
+    public function deleted(ChangeRequestType $changeRequestType): void
+    {
+        if ($changeRequestType->changeRequests()->doesntExist()) {
+            $changeRequestType->forceDelete();
+        }
+    }
```
Orrison, line 48: I am not sure about doing this here and similarly with the other observers.

I think it would be preferred that delete and forceDelete remain as separate actions in the system.

If we want to improve the DeleteAction to present the ability to trigger forceDelete that is fine. But delete should be possible without force delete regardless of the relationship.

mxm1070: I think I would normally agree, but the wording of the request from Product was:
```
If unused, introduce ability to delete.

If used, add ability to archive.
```
As we discussed earlier, I read this as if used soft delete, otherwise force delete. If we don't have this distinction then there's no meaningful difference between deleting a used or unused record. 

We could add a checkbox to the delete action form that switch to a soft delete, but that could still present the same flow problem, i.e., the user deletes a record, forgets about it, then tries to recreate it and runs into the unique constraint. We could have the query exclude deleted records, but then what's the point in having a bunch of hidden unused records? Also, unless we are going to introduce the custom checkbox action everywhere as our deletion standard then I don't think we should do it here.

The double action flow to force delete something later also isn't great because it depends on the user first remembering that there is something to delete, changing the filter to view it, going to the edit page, finally clicking force delete. The seems far less likely after the record is hidden.

Another option would be to pull the force delete action forward and swap the presented action based on whether the record is used or unused, but that feels like a harder pivot from the Filament paradigm than the other suggestions. Truth be told I don't know which way Product would lean. I can ask for more clarification unless you have a strong opinion?

Orrison: Right now, my gut is saying this doesn't belong in the observer. Nowhere else in the application do we force delete during a delete action, and I feel that this creates an unexpected and hidden backend consequence. While at the same time removing the ability to archive an unused record if you wanted to.

I would lean more towards modifying the front end to have a more clear distinction in action.

To modify the delete action, if the record is unused, add something like a check box to force delete. Otherwise, soft delete.
Then, you can also add/modify an action to force delete an unused trashed record later.

That gives the User much more control and information in the situation while not creating singular and unexpected results in the backend.

For the concern of then there being a hidden record. I would say we catch the UniqueConstraintException that would be thrown or check for this scenario and notify the User that there is a duplicate archived record.

@dgoetzit and @danharrin, what do y'all think?

Orrison (approved):

PR #529, opened 2024-02-13
[ADVAPP-301]: Align the experience for public profile features depending on assigned license

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-301

### Technical Description

This PR adds license checks for the authenticated user into public profile based settings.

### Types of changes

- [x] No

* [x] Rebased your branch PR on top of the latest upstream `main` branch.

### DIFF
--- a/app/Providers/FilamentServiceProvider.php
+++ b/app/Providers/FilamentServiceProvider.php
@@ -45,7 +45,6 @@ use Filament\Support\Colors\Color;
 use Filament\Forms\Components\Toggle;
 use Filament\Forms\Components\Checkbox;
 use Illuminate\Support\ServiceProvider;
-use Filament\Forms\Components\Component;
 use Filament\Support\Facades\FilamentView;
 use Filament\Support\Facades\FilamentColor;
 use Filament\Actions\Exports\Models\Export as BaseExport;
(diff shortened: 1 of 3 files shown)


app/Providers/FilamentServiceProvider.php
```diff
@@ -177,5 +180,19 @@ public function boot(): void
...
+                ->hintIcon(fn (Toggle $component) => $component->isDisabled() ? 'heroicon-m-lock-closed' : null)
+                ->hintIconTooltip('A CRM license is required for our public profile features.');
+        });
+
+        Checkbox::macro('lockedWithoutAnyLicenses', function (User $user, array $licenses) {
+            /** @var Checkbox $this */
+            return $this->disabled(! $user->hasAnyLicense($licenses))
+                ->hintIcon(fn (Checkbox $component) => $component->isDisabled() ? 'heroicon-m-lock-closed' : null)
+                ->hintIconTooltip('A CRM license is required for our public profile features.');
+        });
```
mxm1070, line 196: I'm not sure that I entirely agree with creating a macro for this. I would understand if these were intended to be used outside of the profile page, but they aren't, at least as written. I would lean towards creating a custom field and putting the method there before creating a macro. But honestly I would suggest just creating a private method on the profile page and returning the configured field: 
```php
private function configureLicensedField(Field $field): Field
{
    /** @var User $user */
    $user = auth()->user();

    return $field->disabled(! $user->hasAnyLicense([LicenseType::RetentionCrm, LicenseType::RecruitmentCrm]))
        ->hintIcon(fn (Field $component) => $component->isDisabled() ? 'heroicon-m-lock-closed' : null)
        ->hintIconTooltip('A CRM license is required for our public profile features.');
}
``` 
then
```php
$this->configureLicensedField(
    Toggle::make('has_enabled_public_profile')
        ->label('Enable public profile')
        ->live()
),
```
Thoughts?

mxm1070 (changes requested):

app/Providers/FilamentServiceProvider.php
Orrison: From what I can see, this could be used outside of the profile page. And we do already have some other checkboxes and toggles in the app that have similar functionality to this, don't we?

mxm1070: Maybe, we do a similar thing in `app/Filament/Forms/Components/Licenses.php` and `app-modules/portal/src/Filament/Pages/ManagePortalSettings.php`, but the `disabled` conditions and `hintIcon` message are different. And likely anywhere in the future we use this pattern will have different conditions and messages, so I think this macro either needs to be generalized and put on the `Field` component (because they all share these functions from the traits on the `Field` component), or moved into the `EditProfile` as a private method.

dgoetzit: The flexibility can be increased as needed, but I doubt this is the last time that we use this same pattern as pieces of the application are locked down appropriately. Adding a custom field adds a level of overhead that I don't think is worth introducing, as it requires the implementing developer to know that we've overridden a Filament class. This macro is discoverable via your IDE, and allows us to keep consistency with the patterns that someone using Filament would expect.

mxm1070 (approved):

Orrison (approved):

PR #552, opened 2024-02-20
[ADVAPP-361]: Extend user model with job title, [ADVAPP-360]: Enhance AI to be dynamically aware of the user creating prompts

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-361
- https://canyongbs.atlassian.net/browse/ADVAPP-360

### Technical Description

>

Added job title to User and configured the Assistant chat dynamic context to use this information if available.

>

### DIFF
--- a/app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
+++ b/app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
@@ -64,6 +64,7 @@ use AdvisingApp\Assistant\Jobs\ShareAssistantChatsJob;
 use AdvisingApp\Assistant\Enums\AssistantChatShareWith;
 use AdvisingApp\IntegrationAI\Client\Contracts\AIChatClient;
 use AdvisingApp\IntegrationAI\Exceptions\ContentFilterException;
+use AdvisingApp\IntegrationAI\DataTransferObjects\DynamicContext;
 use AdvisingApp\IntegrationAI\Exceptions\TokensExceededException;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use AdvisingApp\Assistant\Services\AIInterface\Enums\AIChatMessageFrom;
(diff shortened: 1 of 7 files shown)


app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
```diff
@@ -220,9 +221,14 @@ public function sendMessage(): void
...
-                $this->stream('currentResponse', nl2br($partial));
-            });
+            /** @var User $user */
+            $user = auth()->user();
+
+            $this->currentResponse = $ai
+                ->provideDynamicContext(new DynamicContext($user))
+                ->ask($this->chat, function (string $partial) {
+                    $this->stream('currentResponse', nl2br($partial));
+                });
```
Orrison, line 231: @dgoetzit and @mxm1070 the way this works does it include it as a system message once. Or is it included with ever prompt that is sent by the User?

mxm1070: I believe when I tested it it was always appended.

Orrison: @mxm1070 got it. Thank you!

@dgoetzit and @danharrin with y'all experience with the AI do you think we should do it this way? I am thinking not, as it increases the tokens per prompt doesn't it? And could result in strange messages.

Perhaps we want a way to have dynamic context included at the start of a conversation. Then every once an a while after.

We don't have to decide or change this now. Just want to discuss it for future items.

dgoetzit: This context should be included in every message.

dgoetzit (approved):

app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
dgoetzit: @Orrison In theory that makes a ton of sense, but I'm not sure how that would impact the response. I don't recall if the system context is preserved over multiple requests. But, if it is (even just for _x_ period of time), and if we could determine exactly how often we'd need to provide that context again without negatively impacting the results I'd be in favor of it, as it would reduce usage even if it is only a relatively small token savings per conversation.

Orrison (approved):

PR #560, opened 2024-02-22
[ADVAPP-363]: Introduce ability to log into knowledge portal

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-363

### Technical Description

> This PR adds the ability to authenticate via session or token inside of the Knowledge Management Portal

New env vars

`LANDLORD_APP_HOST `, `SANCTUM_STATEFUL_DOMAINS`, and `SESSION_DOMAIN` will need to be set


portals/knowledge-management/src/App.vue
```diff
@@ -39,12 +39,24 @@ import AppLoading from '@/Components/AppLoading.vue';
...
+import { useTokenStore } from '@/Stores/token.js';
 
 const errorLoading = ref(false);
 const loading = ref(true);
 const showMobileMenu = ref(false);
 
+const userIsAuthenticated = ref(false);
+
 onMounted(async () => {
+    await axios.get('/sanctum/csrf-cookie');
```
Orrison, line 54: Will this work in the embeded version of the portal?

resources/views/test-embed.blade.php
```diff
@@ -0,0 +1,47 @@
...
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+--}}
+<div>
+    <knowledge-management-portal-embed
+        url="http://test.advisingapp.local/api/portal/knowledge-management?signature=a9425eb754fc360f36635d850c3140013ff0391c2a80e1496c34ec41bfbc63f6"
```
Orrison, line 36: Is this a file we should delete later?

Orrison (changes requested):

resources/views/test-embed.blade.php
dgoetzit: Yes - I'll PR after demo

portals/knowledge-management/src/App.vue
dgoetzit: Nope - good catch. This is not the correct endpoint.

Orrison (approved):

PR #561, opened 2024-02-22
[ADVAPP-350]: Improve seeding process for demo/dev data

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-350

### Technical Description

Adjusted the seeding process to only seed the needed data when a new tenant is created. Added a process to generate a super admin user to be used temporarily for new tenants.

We will need to wipe the DBs as well as set `ALLOW_SUPER_ADMIN_CREATION` to `true` to create a new tenant.

### DIFF
--- a/app-modules/service-management/database/seeders/ChangeRequestTypeSeeder.php
+++ b/app-modules/service-management/database/seeders/ChangeRequestTypeSeeder.php
@@ -36,7 +36,6 @@
 
 namespace AdvisingApp\ServiceManagement\Database\Seeders;
 
-use App\Models\User;
 use Illuminate\Database\Seeder;
 use AdvisingApp\ServiceManagement\Models\ChangeRequestType;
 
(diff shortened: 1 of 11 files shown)


app-modules/service-management/database/seeders/ChangeRequestTypeSeeder.php
```diff
@@ -51,15 +50,11 @@ public function run(): void
...
-            $changeRequestType = ChangeRequestType::factory()
+            ChangeRequestType::factory()
                 ->create([
                     'name' => $type[0],
                     'number_of_required_approvals' => $type[1],
                 ]);
-
-            if (app()->environment('local', 'testing')) {
-                $changeRequestType->userApprovers()->attach(User::where('email', config('local_development.super_admin.email'))->first()->id);
-            }
```
dgoetzit, line 62: Do we need to get rid of this chunk? It's useful for local.

dgoetzit (approved):
Looks good - left a non-blocking question.

app-modules/service-management/database/seeders/ChangeRequestTypeSeeder.php
Orrison: For now yes, as we won't have a user when this runs by default. So we need a better way to handle this. I'll creat a ticket for it

PR #564, opened 2024-02-26
ADVAPP-393

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-393

### Technical Description

Fixes caching issues

>


app/Multitenancy/Tasks/PrefixCacheTask.php
```diff
@@ -0,0 +1,87 @@
...
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace Spatie\Multitenancy\Tasks;
```
dgoetzit, line 37: We should update the namespace and reference the new class within the config.

dgoetzit (changes requested):

app/Multitenancy/Tasks/PrefixCacheTask.php
Orrison: Ah shoot yes nice catch, thank you!

dgoetzit (approved):

PR #568, opened 2024-02-27
[ADVAPP-365]: Design process for data migration

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-365

### Technical Description

Creates a process for data migration in our application based on [One-Time Operations for Laravel](https://github.com/TimoKoerber/laravel-one-time-operations).

This allows us to have one-time commands that can be configured to run/dispatch during deployment that are tracked and run only once, similar to migrations.

The operations process commands will need to be added to deployment scripts. This will be done during CI/CD setup tickets.


app-modules/data-migration/database/migrations/2024_02_26_000000_create_one_time_operations_table.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __construct()
+    {
+        $this->name = OneTimeOperationManager::getTableName();
+    }
+
+    public function up(): void
+    {
+        Schema::create($this->name, function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->string('name');
```
dgoetzit, line 55: Since we are potentially looking for operations by name, should this column have a unique constraint?

database/landlord/2024_02_26_000000_create_one_time_operations_landlord_table.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __construct()
+    {
+        $this->name = OneTimeOperationManager::getTableName();
+    }
+
+    public function up(): void
+    {
+        Schema::create($this->name, function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->string('name');
```
dgoetzit, line 55: Since we are potentially looking for operations by name, should this column have a unique constraint?

app-modules/data-migration/database/migrations/2024_02_26_000000_create_one_time_operations_table.php
```diff
@@ -0,0 +1,61 @@
...
+            $table->string('dispatched');
+            $table->timestamp('processed_at')->nullable();
```
dgoetzit, line 57: I'm wondering if we should make these timestamps a little bit more robust to account for failure and multiple process attempts. What do you think about adding `first_processed_at` and `last_processed_at`? For more granularity, we can also add a count for `process_attempts`.

dgoetzit (changes requested):
This looks great - thanks for providing the docs as well. I left some thoughts/questions.

app-modules/data-migration/src/Models/Operation.php
```diff
@@ -0,0 +1,96 @@
...
+
+    public function getFilePathAttribute(): string
+    {
+        return OneTimeOperationManager::pathToFileByName($this->name);
+    }
+
+    protected static function newFactory(): OperationFactory
+    {
+        return new OperationFactory();
+    }
```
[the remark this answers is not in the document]
Orrison: Ah no, good catch.

app-modules/data-migration/src/Commands/OneTimeOperationsProcessCommand.php
```diff
@@ -0,0 +1,325 @@
...
+
+            $this->process();
+        });
+
+        return self::SUCCESS;
+    }
+
+    protected function typeArgumentIsValid(): bool
+    {
+        return in_array($this->argument('type'), ['landlord', 'tenant']);
```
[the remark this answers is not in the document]
Orrison: That is a good idea. Changed it to use the enum case throughout this file.

app-modules/data-migration/database/migrations/2024_02_26_000000_create_one_time_operations_table.php
Orrison: Yeah, that makes sense.

database/landlord/2024_02_26_000000_create_one_time_operations_landlord_table.php
Orrison: Yeah, that makes sense.

app-modules/data-migration/database/migrations/2024_02_26_000000_create_one_time_operations_table.php
Orrison: I think we would want to ensure these only ever run once. So, tracking that may not be needed. If the job that get's dispatch fails we can retry that job.

If we do ever need to re-run the command there is the option with this to pass the operation name directly. If you do so it runs it regardless of if it has run before.

dgoetzit (dismissed):

dgoetzit (approved):

PR #577, opened 2024-03-04
[ADVAPP-345]: Track views of knowledge base articles

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-345

### Technical Description

Rate limiter applied to the "view" page of the View Knowledge Base Item page for 1 attempt every minute. It only gets bumped when the page gets loaded and not when any Livewire update request gets made.

### DIFF
--- a/app-modules/knowledge-base/database/migrations/2024_03_04_155253_create_knowledge_base_item_views_table.php
+++ b/app-modules/knowledge-base/database/migrations/2024_03_04_155253_create_knowledge_base_item_views_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Advising App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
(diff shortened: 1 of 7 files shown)


app-modules/knowledge-base/database/migrations/2024_03_04_155253_create_knowledge_base_item_views_table.php
```diff
@@ -0,0 +1,52 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('knowledge_base_item_views', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->foreignUuid('knowledge_base_item_id')->constrained('knowledge_base_articles')->cascadeOnDelete();
+            $table->foreignUuid('user_id')->constrained()->cascadeOnDelete();
```
Orrison, line 47: Let's actually remove the `cascadeOnDelete` for the User here. I feel like even if the User record was deleted we wouldn't want to decrement views.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ViewKnowledgeBaseItem.php
```diff
@@ -56,6 +58,19 @@ public function getTitle(): string | Htmlable
         return $this->record->title;
     }
 
+    public function mount(int | string $record): void
```
Orrison, line 61: I think we should maybe do this on the edit page as well. What do you think?

Orrison (changes requested):

app-modules/knowledge-base/database/migrations/2024_03_04_155253_create_knowledge_base_item_views_table.php
danharrin: So it needs to be nullable and nullOnDelete?

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ViewKnowledgeBaseItem.php
danharrin: That wasn't part of the ticket, and it feels like editors aren't utilising the article so it shouldn't increment the views IMO

app-modules/knowledge-base/database/migrations/2024_03_04_155253_create_knowledge_base_item_views_table.php
danharrin: Done that, but LMK if you want it some other way

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ViewKnowledgeBaseItem.php
Orrison: Got it, makes sense.

Orrison (approved):

PR #581, opened 2024-03-06
[ADVAPP-277]: CI/CD to support production releases

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-277

### Technical Description

Updates the Docker image and the s6-overlay tasks to allow a more proper setup for deployment and CI/CD.

On startup, if the proper `.env` variables are set, the following will happen:

- Caching
    - Config
    - Event
    - Route
    - View
    - Icons
    - Filament Components
- Migrations for both tenant and landlord will be dispatched to run on the queue.

Separate tickets/PRs will be created to move the migrations to their own queue. As well as some improvements to Data Migrations as well as listening to and doing something about failure/success events. 

The new cache and migrate `.env` vars will need to be setup in dev/prod.


app/Console/Commands/DispatchLandlordMigrations.php
```diff
@@ -0,0 +1,75 @@
...
+    public function handle(): int
+    {
+        Bus::batch(
+            [
+                [
+                    new LandlordSchemaMigration(),
+                    new DispatchLandlordDataMigrations(),
+                ],
+            ]
+        )
```
dgoetzit, line 63: Should these be chained rather than batched? Are the data migrations dependent on the schema migrations running successfully?

app/Console/Commands/DispatchTenantMigrations.php
```diff
@@ -0,0 +1,89 @@
...
+        $tenant = Tenant::current();
+
+        Bus::batch(
+            [
+                [
+                    new TenantSchemaMigration(),
+                    new DispatchTenantDataMigrations(),
+                ],
+            ]
+        )
```
dgoetzit, line 70: Same question here regarding batching vs. chaining.

dgoetzit (changes requested):
This looks great - just had a question on the logistics of migrations.

app/Console/Commands/DispatchTenantMigrations.php
Orrison: These are chained and batched. The second array there makes sure that the two jobs are put onto the batch as a chain.

dgoetzit (dismissed):

dgoetzit (approved):

PR #617, opened 2024-03-22
[ADVAPP-389]: Update label for Interactions > Campaigns

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-389

### Technical Description

> This PR introduces InteractionInitiative as a replacement for InteractionCampaign, using a strategy that allows user experience to continue as normal as the transition occurs behind the scenes. Labels and references for InteractionCampaign have been modified to appear as Initiatives, as Campaigns will still be used until the data migration takes place, and the feature flag enables the use of InteractionInitiative as a concept. We'll introduce a subsequent PR to remove the remnants of InteractionCampaign, its table, and foreign key columns.

> We will need to ensure that the Roles/Permissions sync occurs, so that the new `interaction_initiative` permissions are created and applied to the appropriate role.


database/operations/.gitkeep
Orrison, line 1: Let's leave this here. We can delete operations over time so it is possible and valid for this folder to be empty.

Orrison (changes requested):
Just a minor fix and then this should be g2g. If you can get a PR created for the cleanup as well that would be awesome.

We will wait to merge the cleanup till next Sprint. But it's good to have it prepared.

Orrison (changes requested):
Looks like a test are currently failing

Orrison (approved):

PR #629, opened 2024-04-01
[ADVAPP-463]: Remove Onetime Operations and switch to only using migrations

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-463

### Technical Description

This removes all logic and references to the old One-Time Operations. It also updates the documentation to provide a guide on how to create data migrations in our new paradigm. Finally, it changes the s6-overlay tasks to run migrations within the task and not within jobs.

We will need to discuss this further as a team, but for now, I have decided to move all migrations out of the job/job batch pattern we originally had.
Because data migrations are now part of regular migrations, we cannot guarantee how long they will take to run. So, it is safer to have the s6-overlay task run until the migrations are completed for the landlord and all tenants.

>


Orrison:
> Should there be a new migration to drop the one time operations table/s?

Ah! Yes, there should be. Good catch. Working on that now.

danharrin (approved):

PR #632, opened 2024-04-02
[ADVAPP-462]: Resolve embed snippet for forms

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-462

### Technical Description

Fixes the issue with the embed snippet display in the UI. It seems there was some change in the Filament that now results in the HTML being sanitized. So, I removed marking it as `markdown` and formatted it manually.

@danharrin, can you confirm this is the best approach or at least a good enough one?

>

### DIFF
--- a/app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
+++ b/app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
@@ -36,11 +36,9 @@
 
 namespace AdvisingApp\Form\Filament\Resources\FormResource\Pages;
 
-use Illuminate\Support\Str;
 use Filament\Actions\Action;
 use AdvisingApp\Form\Models\Form;
 use Filament\Actions\DeleteAction;
-use Illuminate\Support\HtmlString;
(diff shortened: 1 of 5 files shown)


app-modules/form/src/Filament/Resources/FormResource/Pages/EditForm.php
```diff
@@ -76,13 +78,14 @@ protected function getHeaderActions(): array
...
                                 $code = resolve(GenerateSubmissibleEmbedCode::class)->handle($form);
 
-                                return <<<EOD
+                                $state = <<<EOD
                                 ```
                                 {$code}
                                 ```
                                 EOD;
+
+                                return new HtmlString(Str::markdown($state));
```
danharrin, line 87:
```suggestion
                                return str($state)->markdown()->toHtmlString();
```

IMO this looks cleaner, you can apply to the other files too if you want?

danharrin (changes requested):

danharrin (approved):

PR #634, opened 2024-04-03
[ADVAPP-383]: Delay loading of portal contents until after authentication

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-383

### Technical Description

A new API endpoint has been introduced to return categories, and they have been removed from the existing one. If authentication is required and has not been performed, the request to fetch categories is delayed until it happens successfully.


Orrison (approved):
This looks good to me, but @dgoetzit if you can review as well.

PR #653, opened 2024-04-09
[ADVAPP-482]: Introduce ability to import a list of users

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-482

### Technical Description

> This PR introduces the ability to import a list of users via Filament import functionality.

> Yes - we will need to ensure that we run the Roles and Permissions sync process so that the `user.import` permission is added and available. 

### DIFF
--- a/app/Filament/Imports/UserImporter.php
+++ b/app/Filament/Imports/UserImporter.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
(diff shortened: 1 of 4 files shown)


app/Filament/Imports/UserImporter.php
```diff
@@ -0,0 +1,106 @@
...
+                ->requiredMapping()
+                ->example('Jonathan Smith'),
+            ImportColumn::make('email')
+                ->rules(['required', 'email'])
+                ->requiredMapping()
+                ->example('johnsmith@gmail.com'),
+            ImportColumn::make('job_title')
+                ->rules(['required'])
+                ->requiredMapping()
+                ->example('12345'),
```
Orrison, line 63: A better example here might be like "Advisor"

Orrison (changes requested):
Looks great. Just one small nit.

Orrison (approved):

PR #655, opened 2024-04-09
[Chore] Check environment when syncing roles and permissions

dgoetzit:
### Ticket(s) or GitHub Issue

n/a

### Technical Description

> Because roles are not tested, we have no need to sync roles and permissions when running tests

> No


Orrison (approved):

Orrison (changes requested):
You may want to close this as I have fixed all test failures in my PR with this change.

PR #675, opened 2024-04-16
[ADVAPP-367]: AI chat history sometimes presents out of order

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-367

### Technical Description

> This PR preserves the initial message creation time in the ChatMessage DTO, so that the chat message order is correct upon saving.

> No

### DIFF
--- a/app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
+++ b/app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
@@ -844,7 +844,7 @@ class PersonalAssistant extends Page
             message_id: $messageId,
             message: $message,
             from: $from,
-            created_at: now()
+            created_at: now(),
         );
     }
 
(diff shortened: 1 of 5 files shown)


app-modules/assistant/src/Filament/Pages/PersonalAssistant.php
```diff
@@ -842,6 +844,7 @@ protected function setMessage(string $message, AIChatMessageFrom $from, string $
             message_id: $messageId,
             message: $message,
             from: $from,
+            created_at: now()
```
danharrin, line 847:
```suggestion
            created_at: now(),
```

danharrin: I wonder if we can enable trailing comma enforcement in Pint

dgoetzit: I would think the `trailing_comma_in_multiline` rule should enforce that since it mentions "parameters list" as a case where multiline expressions must have a trailing comma, but it doesn't seem to 🤷

danharrin (approved):
There are a few other missing commas but honestly not really something that we should spend time fixing, maybe just create a new ticket to investigate the lack of CS fixer action?

PR #696, opened 2024-04-30
[ADVAPP-511]: Introduce Creation of Custom Assistants

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-511

### Technical Description

Assistant chats are now associated with an assistant, if the feature has been migrated. If an assistant hasn't been associated, the default assistant is attached automatically for backwards compatibility.

When switching between chats, the current assistant is swapped so the AI understands how to respond based on the instructions.

When assistants are created and updated, those updates are reflected to the Azure API through an observer.


app-modules/assistant/database/migrations/2024_04_23_123415_create_ai_assistants_table.php
```diff
@@ -0,0 +1,84 @@
...
+
+            $table->timestamps();
+            $table->softDeletes();
+        });
+
+        Schema::table('assistant_chats', function (Blueprint $table) {
+            $table->foreignUuid('ai_assistant_id')->nullable()->constrained('ai_assistants');
+        });
+
+        Feature::activate('custom-ai-assistants');
```
Orrison, line 63: Schema migrations should only contain schema migrations. Any data migration including feature activation or purging should be moved to a separate data migration.

app-modules/assistant/src/Actions/CreateCustomAiAssistant.php
```diff
@@ -0,0 +1,63 @@
...
+    public function __construct(
+        private AiChatClient $ai
+    ) {}
+
+    public function create(AiAssistant $assistant): AiAssistant
+    {
+        $clientAssistantId = resolve(CreateAiAssistant::class)->from(
+            name: $assistant->name,
+            description: $assistant->description,
+            // Maximum length is 32768 characters - needs to be enforced
```
Orrison, line 53: Please create a ticket in the backlog for this and remove the comment.

app-modules/assistant/src/Actions/CreateDefaultInsitutionAiAssistant.php
```diff
@@ -0,0 +1,89 @@
...
+namespace AdvisingApp\Assistant\Actions;
+
+use App\Models\Tenant;
+use Laravel\Pennant\Feature;
+use AdvisingApp\Assistant\Models\AiAssistant;
+use AdvisingApp\Assistant\Enums\AiAssistantType;
+use AdvisingApp\IntegrationAI\Settings\AISettings;
+use AdvisingApp\IntegrationAI\Client\Contracts\AiChatClient;
+
+class CreateDefaultInsitutionAiAssistant
```
Orrison, line 46:
```suggestion
class CreateDefaultInstitutionAiAssistant
```

app-modules/assistant/src/Actions/CreateDefaultInsitutionAiAssistant.php
```diff
@@ -0,0 +1,89 @@
...
+        private AiChatClient $ai
+    ) {}
+
+    public function create()
+    {
+        $tenant = Tenant::current();
+
+        $settings = resolve(AISettings::class);
+
+        // TODO Update to reflect desired institutional defaults
```
Orrison, line 58: Can you explain this one a little more? Is this functionality we are losing or some functionality we want to suggest to Product?

Regardless we should move it to a backlog item.

app-modules/assistant/src/Actions/GetAiAssistantId.php
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+namespace AdvisingApp\Assistant\Actions;
+
+use App\Enums\Feature;
+use Illuminate\Support\Facades\Gate;
+use AdvisingApp\Assistant\Models\AiAssistant;
+use AdvisingApp\IntegrationAI\Settings\AISettings;
+
+class GetAiAssistantId
```
Orrison, line 44: So this always returns the tenant default AssistantId, right?

Is it expected to remain this way indefinitely? If so I propose changing the name to something like `GetDefaultAiAssistantId`

app-modules/assistant/src/Filament/Resources/AiAssistantResource/Forms/AiAssistantForm.php
```diff
@@ -0,0 +1,82 @@
...
+class AiAssistantForm
+{
+    public function form(Form $form): Form
+    {
+        return $form
+            ->schema([
+                SpatieMediaLibraryFileUpload::make('avatar')
+                    ->label('Avatar')
+                    ->disk('s3')
+                    ->collection('avatar')
```
Orrison, line 55: I don't see the `avatar` collection defined in the `AiAssistant` Model. Is this intentional / working?

app-modules/assistant/src/Observers/AiAssistantObserver.php
```diff
@@ -0,0 +1,63 @@
...
+use AdvisingApp\Assistant\Models\AiAssistant;
+use AdvisingApp\Assistant\Actions\UpdateAiAssistant;
+use AdvisingApp\Assistant\Actions\CreateCustomAiAssistant;
+use AdvisingApp\Assistant\DataTransferObjects\AiAssistantUpdateData;
+
+class AiAssistantObserver
+{
+    public function created(AiAssistant $assistant): void
+    {
+        resolve(CreateCustomAiAssistant::class)->create($assistant);
```
Orrison, line 48: Should we add a ticket to look into handling this if it fails?

As it stands, if this fails, we will be left with a created `AiAssitant` Model in our system, but not one that is created in external service

app-modules/assistant/src/Observers/AiAssistantObserver.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        resolve(UpdateAiAssistant::class)->from(
+            assistantId: $assistant->assistant_id,
+            data: new AiAssistantUpdateData(
+                name: $assistant->name,
+                description: $assistant->description,
+                instructions: "{$assistant->instructions} {$assistant->knowledge}",
+                model: $assistant->model,
+            )
+        );
```
Orrison, line 61: As above, if this fails, we will be left with an updated model in our system, but the assistant will not be updated in the external service.

Should we consider reversing the steps and doing these on updating and creating? To gracefully fail if it is not created or updated in the external service?

database/migrations/Concerns/CanModifySettings.php
```diff
@@ -0,0 +1,73 @@
...
+*/
+
+namespace Database\Migrations\Concerns;
+
+use Closure;
+use Illuminate\Support\Facades\DB;
+
+trait CanModifySettings
+{
+    public function updateSettings(string $group, string $name, Closure $modifyPayload, bool $isEncrypted = false): void
```
Orrison, line 44: What are your thoughts on how we could possibly make the `modifyPayload` argument safer for inferring the arguments passed?

Though this is a really simple example with only one argument and not something we need to tackle immediately, I would like to figure out an alternative or a better way to type hinting `Closure`.

Though it works, it feels like we should either add information in docs or a docblock or requires that the consumer read through the code of `updateSettings` to understand what arguments are expected to be provided. There is no type safety or help in defining the arguments of the Closure when using the API.

I don't think there is a native way to define `Closure` arguments in docblocks (which still isn't perfect as it can become out of date)
My other thought is to use something like an Interface or Abstract class instead of `Closure` but that just sounds like a mess having to define a whole new class every time you want to call a method.

Orrison (changes requested):
Great work, notes some items and it also looks like there are some small test failures.

Orrison (approved):

PR #702, opened 2024-05-02
[ADVAPP-288]: Ensure Twilio is enabled as a service before allowing related functionality

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-288

### Technical Description

>

Handle situations where the Twilio integration hasn't been configured or enabled.

>


app/GraphQL/Execution/IntegrationErrorHandler.php
Orrison, line 1: Is this because errors don't normally get reported in a GraphQL call?

config/local_development.php
```diff
@@ -38,4 +38,14 @@
...
         'email' => 'sampleadmin@advising.app',
     ],
+    'internal_users' => [
+        'emails' => env('DEMO_INTERNAL_USER_EMAILS') ? explode(',', env('DEMO_INTERNAL_USER_EMAILS')) : null,
+    ],
+    'twilio' => [
+        'account_sid' => env('TWILIO_ACCOUNT_SID'),
+        'auth_token' => env('TWILIO_AUTH_TOKEN'),
+        'from_number' => env('TWILIO_TEST_FROM_NUMBER', env('TWILIO_FROM_NUMBER', env('TWILIO_PHONE_NUMBER'))),
+        'to_number' => env('TWILIO_TEST_TO_NUMBER', env('TWILIO_TO_NUMBER')),
```
Orrison, line 48: What is the purpose of having these in a separate local development config? Do we have config for the twilio integration itself?

I expect that it make sense to not need to duplicate things like this. Local development settings really only need to be like `enable_test_sender` shouldn't it?

database/seeders/LocalDevelopmentSeeder.php
Orrison, line 1: Hmm, co-oping this InternalAdminUsersSeeder could be a little problematic. We would not really using it for local development in the past. But instead to seed internal Users in a new deployment so we could login and start setup.

app-modules/engagement/src/Enums/EngagementDeliveryMethod.php
```diff
@@ -46,8 +48,39 @@ enum EngagementDeliveryMethod: string implements HasLabel
...
+        };
+    }
+
+    public function getLabelForIntegrationState(): string
+    {
+        return $this->getCaseDisabled()
+            ? sprintf(
+                '%s (%s)',
+                $this->getLabel(),
+                IntegrationException::make(Integration::Twilio)->getMessage()
```
Orrison, line 75: Should we have a match statement here to future proof this? If we add another integration to `getCaseDisabled` in the future it would continue to display the Twilio error wouldn't it?

Orrison (changes requested):

app/GraphQL/Execution/IntegrationErrorHandler.php
mxm1070: So this is a bit of a weird one that comes down to a design decision from the Webonyx/Lighthouse team. 

If you have `APP_DEBUG` set to false, GraphQL will report a generic `Internal server error` to the client. This isn't great, because the user will likely think they could fix the problem somehow and keep trying. If you want the error to present you have to mark it as client safe. However, the default error handler assumes client safe errors are fixable by the user, even if it's not, and eats the error without reporting it to any of our internal logging.

I thought it would be useful to see if api users are triggering integration errors, because it might speak to a need to change how we display required settings if it happens a lot. If you don't think it's important I can remove the custom error handler and allow the default functionality to return.

config/local_development.php
mxm1070: The original three test settings were created awhile ago to help with development and were being used prior to this. 

I added `account_sid` and `auth_token` to speed up local development after I copied the keys 30+ times. I relocated the other configs to colocate them. I thought this was prudent given the rest of the configs in `services` are production configs and these only serve a purpose in testing, but they were left in the there after we moved the production configs into the UI. None of the

The `local_development` config was only used before to set the default super admin email, but I thought it would be a good place to centralize our other local development configs.

database/seeders/LocalDevelopmentSeeder.php
mxm1070: I co-opted it because as far as I can tell we stopped using it when we stopped doing the full seeding by default. The only other place I think it was used is `composer refresh-database`, but we haven't really supported that in awhile. 

I'm unaware of how the deployment works, so if you're saying it's used externally I'll revert it and make a new seeder. I'd like to keep the concept around to facilitate the `--local` option I added to the `CreateTenant` command, because it makes local development faster.

app/GraphQL/Execution/IntegrationErrorHandler.php
Orrison: That makes complete sense. But I would opt that we change this to a `Log::info`. `report()` will log an error into our error handling software and we want to try and minimize what gets put there to true errors for the time being.

config/local_development.php
Orrison: Got it, so these will act as an override the each tenant uses so you don't have to go put them in manually into the integration setting?

database/seeders/LocalDevelopmentSeeder.php
Orrison: After some more thought, it's fine. We shouldn't;t be using the internal seeder anymore to your point.

Orrison (changes requested):
Looking good, just one last actionable comment for me left.

config/local_development.php
mxm1070: Yeah, if you look at the `--local` command option you see I only call the seeder after everything is created like we've done with the other options: 

```php
if ($this->option('yes') || $this->option('local') || $this->confirm('Would you like to seed local development data?')) {
    Artisan::call(
        command: "tenants:artisan \"db:seed --database={$database} --class=LocalDevelopmentSeeder\" --tenant={$tenant->id}",
        outputBuffer: $this->output,
    );
}
```
My hope is that we can just keep adding things to the seeder and config and we don't have to setup the integrations every time for each tenant. And in general just makes it a faster setup.

Orrison (approved):

Orrison:
@mxm1070 merging this one in as it looks good to me now. But would you mind taking a screen recording and some screenshots of what this change looks like and the affect it has? And upload those to comments of the ticket.

This one is just slightly harder to "see" so it will be really helpful to have these references.

PR #710, opened 2024-05-06
[ADVAPP-534]: Refactor timezone handling throughout the application

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-534

### Technical Description

Filament now receives the users timezone / settings timezone / system timezone automatically and formats the date/time objects appropriately. It also adds tooltips everywhere to remind the user of the timezone being used.

If this approach does not present issues over the next few weeks/months, we can improve this DX in Filament itself.


Orrison (changes requested):
Test failures seeming to do with timezone settings changes.

danharrin:
@Orrison the failing tests were testing data migrations that happened a month ago. Since so much time has passed, I have deleted them, as that was the quickest option and I didn't think there was much point having them anymore. The error was because we were rolling back migrations to a certain point in order to test the data migration, and we didn't use a `down()` method on some of them, so they repeated their actions. This is an actual issue that we need to consider if we use this testing technique in the future, and may force us to write proper `down()` migrations for all upcoming features.

danharrin:
I have targeted this at ADVAPP-529 so the tests re-run quicker so please merge that first (#717)

database/migrations/2024_05_06_160433_data_activate_display_settings_feature.php
```diff
@@ -0,0 +1,45 @@
...
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use Laravel\Pennant\Feature;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
```
Orrison, line 41: This is missing a `down`

database/migrations/2024_05_06_161618_data_create_manage_display_settings_permission.php
```diff
@@ -0,0 +1,60 @@
...
+
+</COPYRIGHT>
+*/
+
+use Illuminate\Support\Str;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
```
Orrison, line 42: This is missing a down, and I believe we wanted to name these migrations with `seed_permissions_`as the prefix

database/settings/2024_05_06_155657_create_display_settings.php
```diff
@@ -0,0 +1,44 @@
...
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    public function up(): void
```
Orrison, line 40: This is missing a `down`

Orrison (changes requested):

database/migrations/2024_05_06_160433_data_activate_display_settings_feature.php
danharrin: Done

database/migrations/2024_05_06_161618_data_create_manage_display_settings_permission.php
danharrin: Done

database/settings/2024_05_06_155657_create_display_settings.php
danharrin: Done

Orrison (approved):

PR #712, opened 2024-05-07
[ADVAPP-488]: Prevent issues with brand new tenants throwing errors in scheduler

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-488

### Technical Description

>

Prevents the scheduler from running on tenants that haven't finished setting up. Introduces a data migration and feature flag to handle existing tenants.

>


database/landlord/2024_05_06_221309_data_fill_setup_complete_on_tenants_table.php
```diff
@@ -0,0 +1,52 @@
...
+    public function up(): void
+    {
+        if (Schema::hasColumn('tenants', 'setup_complete')) {
+            DB::table('tenants')->update(['setup_complete' => true]);
+        }
+    }
+
+    public function down(): void
+    {
+        // nop
```
Orrison, line 50: @dgoetzit and/or @danharrin do y'all see an issue with there not being a `down` here?

I feel like we talked about these types of scenarios before but forget where we landed. Obviously it is relatively impossible to revert this, since we can't know which items were updated. But will having nothing in down affect rollbacks?

Orrison (commented):
This looks good, but just opened a comment to discuss with the team.

database/landlord/2024_05_06_221309_data_fill_setup_complete_on_tenants_table.php
danharrin: Personally, I would prefer it if all data migrations had a `down()` method where possible, but I have previously been told that was not necessary so have not been doing it recently

Orrison: Yeah that does make sense. So because they all started out as false, it shouldn't hurt for the down here to just set all to false?

danharrin: The only reason it would hurt would be if a user had set it to true manually, but if we ever needed to revert then we'd probably have bigger problems

Orrison: True, I don't view the downs as a tool for disaster recovery. We would never use them in production. Only so we can rollback for migration testing.

So with that @mxm1070 if you can update the data migrations to include the `down`

Orrison (approved):

PR #714, opened 2024-05-07
[ADVAPP-539]: Investigate pasting of certain content into email results in fatal errors

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-539

### Technical Description

> Fixes an issue stemming from a missed match when trying to generate markdown from JSON content.

> No


mxm1070 (approved):

app-modules/engagement/src/Actions/GenerateEmailMarkdownContent.php
```diff
@@ -52,6 +51,8 @@ public function __invoke(array $content, array $mergeData = [], string $markdown
                 'mergeTag' => ' ' . $this->text($mergeData[$component['attrs']['id'] ?? null] ?? '', $component),
                 'orderedList' => PHP_EOL . PHP_EOL . $this->orderedList($component, $mergeData),
                 'text' => ' ' . $this->text($component['text'] ?? '', $component),
+                'hardBreak' => '  ' . PHP_EOL,
+                default => '',
```
Orrison, line 55: Right now this is throwing a bunch of errors in Production flooding Sentry and causing this bug. But I am a little worried to just put a `default` here resulting in no content.

Should we be at least logging something when this happens so we can catch types we have not added a match for?

@danharrin thoughts?

mxm1070: Instead of building the markdown by hand we could rely on the tiptap editor to convert the content for us. That's what I did for the sms/email api. The only thing you have to handle by hand are the custom nodes like mergeTags. Then we put the html through something like `league/html-to-markdown`.

```php
public function getBody(): string
{
    $converter = new HtmlConverter();

    $html = tiptap_converter()
        ->getEditor()
        ->setContent($this->body)
        ->descendants(function ($node) {
            if ($node->type !== 'mergeTag') {
                return;
            }

            $node->type = 'text';
            $node->text = $this->getMergeData()[$node->attrs->id ?? null] ?? '';
        })
        ->getHTML();

    return $converter->convert($html);
}
```

I've tested this in my local and it seems to work fine. For example this is some markdown I used to when making the api:
```markdown
Hello {{ student full name }} with email {{ student email }}\n\n* * *\n\n#### _We're_ **trying** to reach you about your car's extended warranty.\n\n* * *\n\n*   apple\n    \n*   melon\n    \n\n* * *\n\n1.  **red**\n    \n2.  _green_\n    \n3.  blue\n
```

Which creates the correct html:
```html
<p>Hello Idella Walker with email arnoldo.walsh@hotmail.com</p><hr><h4><em>We&#039;re</em> <strong>trying</strong> to reach you about your car&#039;s extended warranty.</h4><hr><ul><li><p>apple</p></li><li><p>melon</p></li></ul><hr><ol><li><p><strong>red</strong></p></li><li><p><em>green</em></p></li><li><p>blue</p></li></ol>
```

and then the correct markdown:
```markdown
Hello Idella Walker with email arnoldo.walsh@hotmail.com

---

#### *We're* **trying** to reach you about your car's extended warranty.

---

- apple
- melon

---

1. **red**
2. *green*
3. blue
```

I believe this should take care of all the built in nodes, and we only have to worry about future custom nodes which we would have to do anyway if we build the markdown by hand.

Orrison: > Instead of building the markdown by hand we could rely on the tiptap editor to convert the content for us. That's what I did for the sms/email api. The only thing you have to handle by hand are the custom nodes like mergeTags. Then we put the html through something like `league/html-to-markdown`.
> 
> ```
> public function getBody(): string
> {
>     $converter = new HtmlConverter();
> 
>     $html = tiptap_converter()
>         ->getEditor()
>         ->setContent($this->body)
>         ->descendants(function ($node) {
>             if ($node->type !== 'mergeTag') {
>                 return;
>             }
> 
>             $node->type = 'text';
>             $node->text = $this->getMergeData()[$node->attrs->id ?? null] ?? '';
>         })
>         ->getHTML();
> 
>     return $converter->convert($html);
> }
> ```
> 
> I've tested this in my local and it seems to work fine. For example this is some markdown I used to when making the api:
> 
> ```
> Hello {{ student full name }} with email {{ student email }}\n\n* * *\n\n#### _We're_ **trying** to reach you about your car's extended warranty.\n\n* * *\n\n*   apple\n    \n*   melon\n    \n\n* * *\n\n1.  **red**\n    \n2.  _green_\n    \n3.  blue\n
> ```
> 
> Which creates the correct html:
> 
> ```
> <p>Hello Idella Walker with email arnoldo.walsh@hotmail.com</p><hr><h4><em>We&#039;re</em> <strong>trying</strong> to reach you about your car&#039;s extended warranty.</h4><hr><ul><li><p>apple</p></li><li><p>melon</p></li></ul><hr><ol><li><p><strong>red</strong></p></li><li><p><em>green</em></p></li><li><p>blue</p></li></ol>
> ```
> 
> and then the correct markdown:
> 
> ```
> Hello Idella Walker with email arnoldo.walsh@hotmail.com
> 
> ---
> 
> #### *We're* **trying** to reach you about your car's extended warranty.
> 
> ---
> 
> - apple
> - melon
> 
> ---
> 
> 1. **red**
> 2. *green*
> 3. blue
> ```
> 
> I believe this should take care of all the built in nodes, and we only have to worry about future custom nodes which we would have to do anyway if we build the markdown by hand.

Hmm, very interesting. This would cover us for all existing and perhaps lay an easier pathway for custom nodes. What do you think @danharrin?

danharrin: Sorry if I'm missing something, but what is the block type that the default is catching here?

mxm1070: At the present, nothing. I think @Orrison just means that the original error here was `hardBreak` not being covered by the match. If we introduce a default it will prevent the pasting error from happening again, but it will also cover up the missing block and just not include it in the markdown. Which may, or may not, be desired, but we should at least consider adding a log so we know if a future block isn't covered.

danharrin: I think we should remove the default if we proceed with this method, yes.

I think we should go ahead and merge this PR and then add a spike to the backlog to investigate the new method, to investigate viability. By the sounds of it it sounds like a good idea though, I didn't know such a package existed by a trustworthy author.

mxm1070: Yeah, I didn't either until I was working on the api. I was hesitant at the time to introduce the package just for that feature, but if we can get some use here as well I think it's worth it.

Orrison (approved):

PR #722, opened 2024-05-09
[ADVAPP-559]: UI/UX enhancement on AI badge for assistant

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-559

### Technical Description

Updated the assistant name styling on the personal assistant page. Updated the default assistant name.

Screenshot 2024-05-10 at 12 39 32 PM

Screenshot 2024-05-10 at 12 39 24 PM


app-modules/assistant/src/Actions/CreateDefaultInstitutionAiAssistant.php
```diff
@@ -56,7 +56,7 @@ public function create()
         $settings = resolve(AISettings::class);
 
         // TODO Update to reflect desired institutional defaults
```
Orrison, line 58: This TDOD can be removed now

Orrison (changes requested):
This loos good but lets reduce the width of the tag to just a little more then the length of the text. That way it looks more like a tag/badge and we can update it to a button later on.

mxm1070:
I've updated the screenshots.

Orrison (changes requested):
Let's left-align the tag so that it matches what it looked like before, just in tag form.

Also I think if you rebase `main` into your branch it should fix the random test failure you have here.

Orrison (approved):

PR #764, opened 2024-05-29
[ADVAPP-377]: Scope the granted Knowledge Management API token appropriately to limit abuse

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-377

### Technical Description

Scoped the portal and graphql api tokens with specific permissions.


app-modules/knowledge-base/database/seeders/KnowledgeBaseArticleSeeder.php
```diff
@@ -46,5 +48,12 @@ public function run(): void
         KnowledgeBaseArticle::factory()
             ->count(25)
             ->create();
+
+        $settings = app(PortalSettings::class);
+        $settings->knowledge_management_portal_enabled = true;
+        $settings->knowledge_management_portal_primary_color = 'emerald';
+        $settings->knowledge_management_portal_rounding = Rounding::None->value;
+        $settings->knowledge_management_portal_service_management = true;
+        $settings->save();
```
Orrison, line 57: I get what you are doing here to try and save time. But let's not add these settings changes to the Seeder as enabling and configuring of the Portal does not really have to do with seeding KB articles.

app-modules/portal/src/Http/Middleware/AuthenticateIfRequiredByPortalDefinition.php
```diff
@@ -51,8 +54,14 @@ public function handle(Request $request, Closure $next): Response
             return $next($request);
         }
 
-        if (! auth('sanctum')->check()) {
-            abort(403);
+        $user = app(Guard::class)($request);
+
+        if (! $user instanceof Student && ! $user instanceof Prospect) {
```
Orrison, line 59: Shouldn't this be an `||`? Would this not fail since the `$user` would only ever been an instance of a `Student` or a `Prospect` not both?

Could we just check if is an instance of `Educatable`?

app-modules/portal/src/Http/Controllers/KnowledgeManagement/KnowledgeManagementPortalRequestAuthenticationController.php
Orrison, line 1: Though I understand this is a time saving change during the development of this feature and that we did something similar in AidingApp. I would prefer we hold off on doing this right now. I need to think more about whether or not I want us to do this.

config/local_development.php
Orrison, line 1: Though I understand this is a time saving change during the development of this feature and that we did something similar in AidingApp. I would prefer we hold off on doing this right now. I need to think more about whether or not I want us to do this.

database/seeders/LocalDevelopmentSeeder.php
Orrison, line 1: Though I understand this is a time saving change during the development of this feature and that we did something similar in AidingApp. I would prefer we hold off on doing this right now. I need to think more about whether or not I want us to do this.

app-modules/portal/src/Http/Middleware/AuthenticateIfRequiredByPortalDefinition.php
```diff
@@ -51,8 +54,14 @@ public function handle(Request $request, Closure $next): Response
             return $next($request);
         }
 
-        if (! auth('sanctum')->check()) {
-            abort(403);
+        $user = app(Guard::class)($request);
```
Orrison, line 57: Would `auth('sanctum')->user()` work here? What is the purpose of directly instantiating the guard class?

app-modules/portal/routes/api.php
```diff
@@ -52,19 +54,23 @@
...
+        Route::middleware(['auth:sanctum'])
+            ->group(function () {
+                Route::get('/user', function (Request $request) {
+                    $user = $request->user('student') ?? $request->user('prospect') ?? null;
+
+                    if ($user?->tokenCan('knowledge-management-portal')) {
+                        return $user;
+                    }
+
+                    abort(Response::HTTP_FORBIDDEN);
```
Orrison, line 71: Wouldn't applying the `abilities:knowledge-management-portal` middleware to this route do this logic for you?

Orrison (changes requested):

Orrison (approved):

PR #766, opened 2024-05-30
[ADVAPP-592] Footer sticky issue fix in assistant page

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-592

### Technical Description

> Removed the classes "h-screen max-h-[calc(100dvh-20rem)]" that creating issue for sticky footer in assistant page.

### DIFF
--- a/app-modules/ai/resources/views/components/assistant.blade.php
+++ b/app-modules/ai/resources/views/components/assistant.blade.php
@@ -42,7 +42,7 @@
             x-data="chats"
         >
             <div class="col-span-1 select-none">
-                <div class="flex h-screen max-h-[calc(100dvh-20rem)] flex-col gap-y-2">
+                <div class="flex flex-col gap-y-2">
                     <div
                         class="relative"
                         x-data="{ isSearchingAssistants: false }"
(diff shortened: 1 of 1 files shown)


Orrison (changes requested):
As we discussed in Sprint/Tech Planning, please ensure that this does not break that the AI chat window should vertically fill the page, even when no messages have been sent.

Also confirm whether or not this fixes the other display issues. Let me know here if so on either.

ankit-canyon:
> Member

Hi,

Yes it will look like this when there is no chat. 

image

Also it fixes the display issue described in [ADVAPP-611].

Thank you.

Orrison (changes requested):
@ankit-canyon I merged this together with 611. Let's go ahead and fix both issues here in this same PR.

This is looking good on mobile now, but removing the classes does now cause the chat to no longer be at the bottom of the page. And so as you added messages to the chat, the chat window grows in length and moves in position till it is off the page.

We need this fixed so the chat window always fills the size of the screen and does not move as the conversation grows. While also still looking correct on mobile.

Perhaps we need to bring the classes back but add a media query like `lg:` to the classes or something like that?

Orrison:
https://github.com/canyongbs/advisingapp/assets/6799341/f0645486-c0c6-45b4-b8b8-52d0618a3cf3

Orrison (approved):

PR #767, opened 2024-05-30
Keep toggle on when SSO settings are on

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-581

### Technical Description

> In Global settings when either of the one Google SSO or Azure SSO is enabled, then while creating a new user SSO login should be switched on by default and if user wants to switch it off then a popup will appear which will ask for confirmation.

### DIFF
--- a/app/Filament/Resources/UserResource/Pages/EditUser.php
+++ b/app/Filament/Resources/UserResource/Pages/EditUser.php
@@ -51,6 +51,8 @@ use App\Filament\Forms\Components\Licenses;
 use AdvisingApp\Authorization\Models\License;
 use App\Notifications\SetPasswordNotification;
 use STS\FilamentImpersonate\Pages\Actions\Impersonate;
+use AdvisingApp\Authorization\Settings\AzureSsoSettings;
+use AdvisingApp\Authorization\Settings\GoogleSsoSettings;
 
 class EditUser extends EditRecord
 {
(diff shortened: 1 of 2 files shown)


app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -76,7 +81,10 @@ public function form(Form $form): Form
                             ->string()
                             ->maxLength(255),
                         Toggle::make('is_external')
-                            ->label('User can only login via Single Sign-On (SSO)'),
+                            ->label('User can only login via Single Sign-On (SSO)')
+                            ->live()
+                            ->afterStateUpdated(fn (Toggle $component, $state) => $state ? null : (($azureSsoSettings || $googleSsoSettings) ? $component->state(true) && $this->mountAction('showSSOModal') : null))
+                            ->default(($azureSsoSettings || $googleSsoSettings)),
```
Orrison, line 87: There should be no need for the `default` call here since this will be populated by the actual value on the User Model

Orrison (changes requested):
Looks great, just one bit of feedback

app/Filament/Resources/UserResource/Pages/EditUser.php
jayushi-canyon: Done removed default()

Orrison (approved):

PR #768, opened 2024-05-30
Features/advapp 596 Restrict super admins from being placed in "Teams"

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-596

### Technical Description

> Added code for restrict super admin from being placed in "teams". 
  There is below three test cases.
      1. Non Super Admin Users can be added to a team via the UsersRelationManager for Teams.
      2. Super Admin Users cannot be added to a team via the UsersRelationManager for Teams.
      3. Super Admin Users do not show up in UsersRelationManager for Teams search results.
   so first two test cases working perfectly. but 3rd one is not working. I have tried with @mxm1070 @Orrison. 

>

### DIFF
--- a/app-modules/team/tests/Team/EditTeamTest.php
+++ b/app-modules/team/tests/Team/EditTeamTest.php
@@ -40,8 +40,11 @@ use AdvisingApp\Team\Models\Team;
 use function Pest\Laravel\actingAs;
 use function Pest\Livewire\livewire;
 
+use Filament\Forms\Components\Select;
+use Filament\Tables\Actions\AttachAction;
 use AdvisingApp\Team\Filament\Resources\TeamResource;
 use AdvisingApp\Team\Filament\Resources\TeamResource\Pages\EditTeam;
+use AdvisingApp\Team\Filament\Resources\TeamResource\RelationManagers\UsersRelationManager;
 
 // Permission Tests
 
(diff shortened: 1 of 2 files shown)


app-modules/team/tests/Team/EditTeamTest.php
```diff
@@ -40,14 +40,95 @@
...
+//     livewire(UsersRelationManager::class, [
+//         'ownerRecord' => $team,
+//         'pageClass' => EditTeam::class,
+//     ])
+//         ->callTableAction(
+//             AttachAction::class,
+//             data: ['recordId' => $user->getKey()]
+//         )->assertSuccessful();
+// });
+//
```
Orrison, line 129: Are these test commented out on purpose? I believe these should work and should be present.

app-modules/team/tests/Team/EditTeamTest.php
```diff
@@ -40,14 +40,95 @@
...
+test('Super Admin Users cannot be added to a team', function () {
```
Orrison, line 130: Though this test does seem correct in testing that the validation error is thrown when trying to add a User with the Super Admin role. The flow of getting there could be improved.

You should start with a `$user` to pass to `actingAs` and give it the proper permissions to access the Edit team page, like in the commented out tests (that we should bring back) above.
We don't need to test access to the EditTeam page in this test, because the above test should have already validated that.

Then create a second User, something like `$superAdmin`, give that Users the Super Admin role. And test trying to add that User and that you get the proper validation error.

Using the same $user to both login as, run the action, and test the validation feels like it could cause some pollution in the future.

Orrison (changes requested):

app-modules/team/tests/Team/EditTeamTest.php
ketan-canyon: Hi @Orrison 
I will remove that comment. Yesterday me and @mxm1070  checking issue so forgot to remove that comment. sorry for that.

app-modules/team/tests/Team/EditTeamTest.php
ketan-canyon: @Orrison 
I will improve this as you suggested. 
sorry, Because First Time I am writing test cases so this issues happen. next time I will keep this things  in mind.

ketan-canyon: Hi @Orrison 
All the changes done as per your suggestion and also 3rd test case is done.

Orrison (approved):

PR #771, opened 2024-05-31
ADVAPP-615 Decrease prominence of AI Assistant badge

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-615

### Technical Description

> Decrease prominence of AI Assistant badge. for that I have add xxs size to the extend section under theme.fontSize in tailwind config file.

>

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -68,7 +68,7 @@ export default {
                 gray: require('tailwindcss/colors').zinc,
             },
             fontSize: {
-                'xxs': ['.5rem','.75rem'],
+                'xxs': ['.5rem'],
             },
         },
         // Changes to colors also need to be reflected in FilamentServiceProvider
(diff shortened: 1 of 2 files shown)


tailwind.config.js
```diff
@@ -67,6 +67,9 @@ export default {
             colors: {
                 gray: require('tailwindcss/colors').zinc,
             },
+            fontSize: {
+                'xxs': ['.5rem','.75rem'],
+            },
```
Orrison, line 72: font-size and line-height are separate classes in Tailwind so I would prefer we keep them separate in any custom classes we created.

There already exists a `.75rem` line-height class `leading-3`. So let's use that, remove the line height from this new `xxs` class. Then this should be good.

Orrison (changes requested):

tailwind.config.js
ketan-canyon: Hi @Orrison 
Changes done as per your suggestion. I don't know about that leading-3 class. sorry for that.

Orrison: No problem at all!

Orrison (approved):

PR #775, opened 2024-06-03
[ADVAPP-589] Enhance the UI/UX for unfiled and filed chats using folders in the AI Personal Assistant

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-589

### Technical Description

> Folder chats will now show in container. Fixed chat container height.

image

### DIFF
--- a/app-modules/ai/resources/js/chat.js
+++ b/app-modules/ai/resources/js/chat.js
@@ -42,10 +42,6 @@ document.addEventListener('alpine:init', () => {
         users: [],
 
         init: async function () {
-            this.render();
-
-            setInterval(this.render.bind(this), 500);
-
             const showThreadResponse = await fetch(showThreadUrl, {
                 headers: {
                     Accept: 'application/json',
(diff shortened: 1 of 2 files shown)


ankit-canyon:
> As per our chat:
> 
> Screenshot 2024-06-03 at 11 11 12 AM
> And please take a look at the broken drag-and-drop.

HI,

This fixed in latest commit.

Thank you.

Orrison (changes requested):
Looks like you and @mxm1070 are looking into this further. Please re-request my review when it is ready to look at again.

Orrison:
Hey @ankit-canyon and @mxm1070, this is till not right and there are some more issues now.

It looks like the footer issue fixed in the other PR is now a little worse as it seems to be affecting my screen size now. (1920 x 1080) It looks like the footer issue is present on tablet sizes as well.

Also the position of the message/chat window is not quite right. On a brand new empty chat it looks fine. But as soon as your chat has enough messages to fill the window, it is nudged slightly off screen.

Also I am not sure the folders section needs to be the entire length of the screen. Perhaps it should just have a max height allowing it to grow to the length of the screen and then have a scrollbar. Which now it seems the folders section always has a blank scrollbar, which should be removed.

Screenshot 2024-06-04 at 8 49 01 AM
Screenshot 2024-06-04 at 8 48 14 AM

Orrison (changes requested):
This is looking better but there still seem to be some issues.

- When hovering over the folders, the icons changing to displaying slightly increases the width of the area, causing the whole screen to jostle.
- The empty scrollbar is still present on the folders area.
- On mobile the chat messages area takes up as much space as there are messages, so in long conversations you scroll forever
- Are we sure that calculating the amount of chats or folders is the correct way to handle styling? This seems brittle to screensizes, browsers, etc. there must be some CSS flexboxy way to handle this.

Looking at Dev now I am realizing that the left handle folder and chat area currently does have issues where it grows forever. So since this is a current issue addressed in this ticket I am fine if our changes leave it as is. The goal here across this ticket and the other that was merged into it [ADVAPP-592](https://canyongbs.atlassian.net/browse/ADVAPP-592) was to do two things:
- Fix the footer overlapping the chat and folder area on smaller screens
- Have the folders be in a single box rather than individual boxes per folder

Let's make sure we accomplish these. And if that means we need to go back to ground zero from main to do so then okay. Or if y'all believe we can resolve these final issues I listed above we can move forward with this.

@danharrin would you mind taking a look here and weighing in.

danharrin:
Replaced by #789

PR #777, opened 2024-06-03
[ADVAPP-594]: Prevent super admin email and text messaging from counting towards customer limits

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-594

### Technical Description

> This PR adds handling to ensure we don't count emails/sms sent to Users with the `authorization.super_admin` against organizational quota usage.

> No

### DIFF
--- a/app-modules/notification/src/Notifications/Channels/EmailChannel.php
+++ b/app-modules/notification/src/Notifications/Channels/EmailChannel.php
@@ -133,14 +133,26 @@ class EmailChannel extends MailChannel
         })->count();
     }
 
-    public function canSendWithinQuotaLimits(Notification $notification, object $notifiable): bool
+    public function canSendWithinQuotaLimits(BaseNotification $notification, object $notifiable): bool
     {
         if (! $notification instanceof EmailNotification) {
             throw new Exception('Invalid notification type.');
         }
 
+        $primaryRecipientUsage = 1;
+
+        if ($notification->getMetadata()['outbound_deliverable_id']) {
+            $deliverable = OutboundDeliverable::with('recipient')->find($notification->getMetadata()['outbound_deliverable_id']);
+
+            $recipient = $deliverable->recipient;
+
+            if ($recipient instanceof User && $recipient->hasRole('authorization.super_admin')) {
+                $primaryRecipientUsage = 0;
+            }
+        }
+
         // 1 for the primary recipient, plus the number of cc and bcc recipients
-        $estimatedQuotaUsage = 1 + count($notification->toMail($notifiable)->cc) + count($notification->toMail($notifiable)->bcc);
+        $estimatedQuotaUsage = $primaryRecipientUsage + count($notification->toMail($notifiable)->cc) + count($notification->toMail($notifiable)->bcc);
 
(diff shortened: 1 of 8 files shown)


mxm1070 (changes requested):
Do we need to handle checking in `canSendWithinQuotaLimits` to exclude super admins from the pre-send check also?

dgoetzit:
> Do we need to handle checking in `canSendWithinQuotaLimits` to exclude super admins from the pre-send check also?

> Do we need to handle checking in `canSendWithinQuotaLimits` to exclude super admins from the pre-send check also?

I don't know if that should matter. This solution will ensure that anyone who was a super admin at the time of sending a deliverable is not counted towards the quota. If someone wasn't a super admin at the time of sending, I'm not sure it would make sense to retroactively adjust the quota.

Orrison (changes requested):
@dgoetzit let me know that he still wants to do some work on here before review. @dgoetzit re-request my review when this is ready.

app-modules/notification/src/Notifications/Channels/EmailChannel.php
```diff
@@ -123,14 +124,35 @@ public static function afterSending(object $notifiable, OutboundDeliverable $del
         }
     }
 
-    public function canSendWithinQuotaLimits(Notification $notification, object $notifiable): bool
+    public static function determineQuotaUsage(array $recipients): int
+    {
+        return collect($recipients)->filter(function ($recipient) {
+            $user = User::where('email', $recipient->getAddress())->first();
```
Orrison, line 130: Should we add a `with('roles')` here to eager load the roles relationship for the Users?

app-modules/notification/src/Notifications/Channels/SmsChannel.php
```diff
@@ -137,10 +139,33 @@ public static function afterSending(object $notifiable, OutboundDeliverable $del
         }
     }
 
+    public static function determineQuotaUsage(SmsChannelResultData $result): int
+    {
+        if ($user = User::where('phone_number', $result->message->to)->first()) {
```
Orrison, line 144: Should we eager load `role` here?

tests/Unit/LicenseSettingsEmailQuotaTracking.php
```diff
@@ -96,3 +97,30 @@ function (MessageSent $event) use ($configurationSet) {
...
+    $licenseSettings = app(LicenseSettings::class);
+
+    $licenseSettings->data->limits->emails = 0;
+    $licenseSettings->save();
+
+    $notifiable = User::factory()->create();
+
+    $notifiable->assignRole('authorization.super_admin');
+
+    $notification = new Tests\Unit\TestEmailNotification();
```
Orrison, line 119: Let's import this class

Orrison (changes requested):
Looks good, just some small nits.

Orrison (approved):

Orrison:
@mxm1070 need you to review again since you requested changes.

mxm1070 (approved):

PR #782, opened 2024-06-04
[ADVAPP-622]: Update label for Report Center to Custom Reports

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-622

### Technical Description

> Update label for Report Center to Custom Reports as well as in breadcrumb.

### DIFF
--- a/app-modules/report/src/Filament/Resources/ReportResource.php
+++ b/app-modules/report/src/Filament/Resources/ReportResource.php
@@ -52,9 +52,9 @@ class ReportResource extends Resource
 
     protected static ?int $navigationSort = 10;
 
-    protected static ?string $navigationLabel = 'Report Center';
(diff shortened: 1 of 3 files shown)


mxm1070 (approved):

Orrison (changes requested):
This looks good!

Let's also change the reports list page title to match, saying "Custom Reports". And also check to see if we need to update the view, create, and edit pages to match that verbiage.

payal-canyon:
> This looks good!
> 
> Let's also change the reports list page title to match, saying "Custom Reports". And also check to see if we need to update the view, create, and edit pages to match that verbiage.

Done.

Orrison (approved):

PR #790, opened 2024-06-05
[ADVAPP-610]: Upgrade to Azure OpenAI 4o Model as general intelligence model for SaaS customers

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-610

### Technical Description

This adds a new model and allows selection of the model from the settings or assistant resource. When selecting a model on a different server, a sync process can be triggered.


Orrison (changes requested):
@danharrin I tried to finish this one up but I think the needed changes are a little more extensive then I can get to tonight. So we can pair on it tomorrow, and/or if you want to start working on it when you get in tomorrow here are the details:

- I `auth()->setUser()` the User during the Thread recreation job. That works fine. But there is no User relation to an Assistant that can be used during the Assistant recreation job. In fact, I don't think the Users dynamic content should be included in the Assistant Instructions, only a Thread. That will solve this issue and also be a better configuration overall
- We need to adjust `ReInitializeAiService`, if any of the Assistants fail to build even if I retry the batch and it succeeds Laravel is not intelligent enough to then kick off the next batch for the Threads in the chain. This also blocks other Threads for complete assistants from being recreated. Below is how I think it should work:
    - ReInitializeAiService becomes an Action that creates a Batch, this batch should be named something indicating it is a reinitialization for X model and also not cancel on a failure
    - The batch initially has a single job, that iterates over all assistants for the model and adds a chain onto the Batch. This chains first job should reinitialize the assistant, then the second job should populate the batch with jobs to reinitialize the threads for the assistant
- I am unsure if the WithoutOverlapping is the correct approach or if we want to use a RateLimiter. But regardless I think we need to change the jobs to apply `retryUntil`, as the jobs trying to process eventually trigger max attempts
- I also notice that if I try and chat with a Thread that has not had the jobs process yet, the assistant ID will remain `null`

danharrin:
> I am unsure if the WithoutOverlapping is the correct approach or if we want to use a RateLimiter.

I think this is the correct approach until we know the actual rate limits of the API.

danharrin:
> I also notice that if I try and chat with a Thread that has not had the jobs process yet, the assistant ID will remain null

I think this is because we create the assistant and populate the ID into the model instance, but not save the model instance. Fixing that.

Orrison (approved):

PR #791, opened 2024-06-06
[ADVAPP-590] Enhance the display of saved chats with last engaged hover details

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-590

### Technical Description

> Last engage time will display when hover over chat name on sidebar. Restore the missing responsive center chat buttons from [ADVAPP-592]

#### Before contributing and submitting this PR, make sure you have Read, agree, and are compliant with the [contributing guidelines](https://github.com/canyongbs/advisingapp/blob/main/README.md#contributing).

[ADVAPP-592]: https://canyongbs.atlassian.net/browse/ADVAPP-592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -89,6 +89,19 @@ class="flex flex-col gap-y-1 rounded-xl border border-gray-950/5 bg-white p-2 sh
...
+                                            const formattedDate = new Date(time).toLocaleDateString('en-US', { year: 'numeric', month: 'long', day: 'numeric' });
+                                            document.getElementById('chat-' + threadItemId).title = 'Last Engaged: ' + formattedDate;
+                                        },
+                                        listenForEvent(threadId) {
+                                            window.addEventListener(`message-sent-${threadId}`, (event) => {
+                                                this.updateTitle(event.detail.time, event.detail.threadItemId);
+                                            });
+                                        }
+                                }"
+                                x-init="listenForEvent('{{ $threadItem->id }}')"
```
Orrison, line 104: Now that I am looking at this all together I am realizing that instead of placing this `x-data` and `x-init` on each individual thread, you could place this way up the container for all threads. (in or out of a folder)

You can then just listen for one event "message-sent" and then do your `update-title` method.

I was wrong, we don't need to make the event specific per Thread, because you pass the thread ID to you updateTitle and then update all threads with the ID of that thread with the JS.

So I would move this to the container for all threads. And make it so you are just listening to the single event and then updating the proper thread title based on the ID.

Orrison (changes requested):
One note for improvement, and it looks like there are some merge conflicts from recent changes in this file.

app-modules/ai/resources/views/components/assistant.blade.php
danharrin: I think there's quite a lot of Alpine being underutilised here. The event listener could use `x-on` instead of being manually registered in `x-init`. Also, the title could be bound to with `x-bind:title` instead of manually being set through JavaScript. And then the last updated date can be stored in an Alpine property, which makes everything that listens to it reactive when it is updated.

I don't think the event should contain the current time, that could be simplified by getting the current time inside the listener.

I think that potentially, `x-tooltip` should be used instead of `x-bind:title` to make it more visible.

danharrin (changes requested):

app-modules/ai/src/Models/AiThread.php
```diff
@@ -93,4 +96,21 @@ public function user(): BelongsTo
...
+            get: function () {
+                $date = $this->messages()->latest()->where('user_id', '!=', null)->first();
+
+                if ($date) {
+                    $timezone = app(DisplaySettings::class)->getTimezone();
+
+                    return Carbon::parse($date->created_at)->setTimezone($timezone);
+                }
+
+                return null;
```
danharrin, line 112:
```suggestion
                $date = $this->messages()
                    ->latest()
                    ->whereNotNull('user_id')
                    ->value('created_at');

                if (! $date) {
                    return null;
                }

                return Carbon::parse($date)->setTimezone($timezone);
```

app-modules/ai/src/Models/AiThread.php
```diff
@@ -93,4 +96,21 @@ public function user(): BelongsTo
     {
         return $this->belongsTo(User::class);
     }
+
+    protected function lastEngagedAt(): Attribute
+    {
+        return Attribute::make(
+            get: function () {
```
danharrin, line 103:
```suggestion
            get: function (): ?CarbonInterface {
```

danharrin (changes requested):

app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -214,6 +226,16 @@ class="flex-1 truncate"
                                 @foreach ($folder->threads as $threadItem)
                                     <li
                                         id="chat-{{ $threadItem->id }}"
+                                        x-on:message-sent-{{$threadItem->id }}.window="updateTitle"
```
danharrin, line 229:
```suggestion
                                        x-on:message-sent-{{ $threadItem->id }}.window="updateTitle"
```

app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -89,6 +91,16 @@ class="flex flex-col gap-y-1 rounded-xl border border-gray-950/5 bg-white p-2 sh
                         @foreach ($this->threadsWithoutAFolder as $threadItem)
                             <li
                                 id="chat-{{ $threadItem->id }}"
+                                x-on:message-sent-{{$threadItem->id }}.window="updateTitle"
```
danharrin, line 94:
```suggestion
                                x-on:message-sent-{{ $threadItem->id }}.window="updateTitle"
```

danharrin (approved):

PR #793, opened 2024-06-06
[ADVAPP-579]: Form validation issue for certain input types

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-589

### Technical Description

So this one was a bit weird. It turns out that in a FormKit form, if you never enter a value for an input that input isn't included in the form and never submitted. But, if you enter a value and remove it the key is left in the form with an empty string as it's value. 

Then the Laravel `ConvertEmptyStringsToNull` middleware makes it null. Then the validator fails because it expects the `text_input` to be a string. Originally I was adding the `nullable` validation rule to the input, but that doesn't matter because the `form_field_submission` doesn't allow null values. So instead I just filter out the inputs that are null. 

It also turns out that the only validation error message displayed on the frontend that comes from FormKit is the `required` validation. All others are actually the Laravel native validation messages, which explains why the uuids were being returned instead of the labels. I've added the input labels as attributes in the validator which allows Laravel to swap the uuids in the message with the string.

### DIFF
--- a/app-modules/form/src/Actions/GenerateSubmissibleValidator.php
+++ b/app-modules/form/src/Actions/GenerateSubmissibleValidator.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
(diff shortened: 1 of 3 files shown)


Orrison (approved):
Nice work @mxm1070!

PR #794, opened 2024-06-07
[ADVAPP-307]: Restrict the display of available public profile features depending on whether the user has access to a CRM product

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-307

### Technical Description

> Based on CRM settings set for a user, he/she will have access to some of the profile  settings features.

### DIFF
--- a/app/Filament/Pages/EditProfile.php
+++ b/app/Filament/Pages/EditProfile.php
@@ -94,7 +94,10 @@ class EditProfile extends Page
     {
         /** @var User $user */
         $user = auth()->user();
-
+        $retentionLicense = $user->hasLicense(LicenseType::RetentionCrm);
+        $recruitmentLicense = $user->hasLicense(LicenseType::RecruitmentCrm);
+        $crmSetting = ($retentionLicense || $recruitmentLicense) ? true : false;
+        
         $connectedAccounts = collect([
             Grid::make()
                 ->schema([
@@ -149,7 +152,8 @@ class EditProfile extends Page
         return $form
             ->schema([
                 Section::make('Public Profile')
-                    ->aside()
+                    ->aside()                    
+                    ->visible($crmSetting)
                     ->schema([
                         Toggle::make('has_enabled_public_profile')
                             ->label('Enable public profile')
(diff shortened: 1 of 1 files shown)


app/Filament/Pages/EditProfile.php
```diff
@@ -92,6 +92,9 @@ public function form(Form $form): Form
     {
         /** @var User $user */
         $user = auth()->user();
+        $retentionLicense = $user->hasLicense(LicenseType::RetentionCrm);
+        $recruitmentLicense = $user->hasLicense(LicenseType::RecruitmentCrm);
+        $crmSetting = ($retentionLicense || $recruitmentLicense) ? true : false;
```
mxm1070, line 97: Please use `->hasAnyLicense([` instead and maybe call the variable something like `$hasCrmLicense` so it's clear what it's referencing.

mxm1070 (changes requested):

app/Filament/Pages/EditProfile.php
jayushi-canyon: @mxm1070 Made the changes.

mxm1070: @jayushi-canyon `hasAnyLicense` takes an array so you can do: 
```php
$hasCrmLicense = $user->hasAnyLicense([LicenseType::RetentionCrm, LicenseType::RecruitmentCrm]);
```

Also you can remove the calls to 
```php
->lockedWithoutAnyLicenses(user: auth()->user(), licenses: [LicenseType::RetentionCrm, LicenseType::RecruitmentCrm]),
```
because it looks like they're all covered by `->visible($hasCrmLicense)` now.

Orrison (changes requested):
Let's make @mxm1070's suggested changes.

`hasAnyLicense` can accept an array of License, so you should only need to pass that in and use it everywhere.

And yes because they will not be visible, we can remove the `lockedWithoutAnyLicenses` calls that are checking for the license usage.

app/Filament/Pages/EditProfile.php
jayushi-canyon: @mxm1070 Done with the changes

mxm1070 (approved):

Orrison (approved):

PR #802, opened 2024-06-10
[ADVAPP-522]: Introduce a global setting for a limit for the number of Custom Assistants

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-522

### Technical Description

Assistants can be archived/restored from the edit page. They may be filtered in the library to include/exclude archived assistants. You may create/restore assistants if you have not hit your limit, defined in the settings. Interacting with threads from an archived assistant shows a message at the bottom of the screen explaining the fact that the assistant is archived, instead of a textarea and associated buttons for interacting with the thread. Threads belonging to an archived assistant can still be edited, moved and deleted, but new ones may not be created.

The migrations must be complete otherwise the settings data class will not have the information it needs to construct.


app/DataTransferObjects/LicenseManagement/LicenseLimitsData.php
```diff
@@ -46,6 +46,7 @@ class LicenseLimitsData extends Data
 {
     public function __construct(
         public int $conversationalAiSeats,
+        public int $conversationalAiAssistants,
```
Orrison, line 49: Is it possible to set this equal to `0` here?

I believe that will set it as `0` as a default, which would prevent the error from happening if it was loaded before the migration ran.

Orrison (changes requested):

app/DataTransferObjects/LicenseManagement/LicenseLimitsData.php
danharrin: Sure, but params with defaults must come after params without, so I have now added defaults to all params to preserve the logical order of params

Orrison (changes requested):
Looks good, there are currently some merge conflicts though.

Orrison (approved):

PR #803, opened 2024-06-10
[ADVAPP-602]: Introduce the file upload feature on conversations

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-602

### Technical Description

> Adds support for file uploads on Assistants.

> No


Orrison (changes requested):
Sorry it looks like there are some merge conflicts here now

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -188,6 +190,12 @@ public function loadFirstThread(): void
         $this->selectThread($this->threadsWithoutAFolder->whereNull('assistant.archived_at')->first());
 
         if ($this->thread) {
+            if (! is_null($expiredVectorStore = $this->getExpiredVectorStoresForThread())) {
+                foreach ($expiredVectorStore as $expiredVectorStore) {
+                    $this->recreateVectorStoreForThread($this->thread, $expiredVectorStore);
+                }
+            }
```
Orrison, line 197:
```suggestion
            if (! is_null($expiredVectorStores = $this->getExpiredVectorStoresForThread())) {
                foreach ($expiredVectorStores as $expiredVectorStore) {
                    $this->recreateVectorStoreForThread($this->thread, $expiredVectorStore);
                }
            }
```

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -213,6 +221,108 @@ public function selectThread(?AiThread $thread): void
         }
 
         $this->thread = $thread;
+
+        if (! is_null($expiredVectorStore = $this->getExpiredVectorStoresForThread())) {
+            foreach ($expiredVectorStore as $expiredVectorStore) {
+                $this->recreateVectorStoreForThread($this->thread, $expiredVectorStore);
+            }
+        }
```
Orrison, line 229:
```suggestion
        if (! is_null($expiredVectorStores = $this->getExpiredVectorStoresForThread())) {
            foreach ($expiredVectorStores as $expiredVectorStore) {
                $this->recreateVectorStoreForThread($this->thread, $expiredVectorStore);
            }
        }
```

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -213,6 +221,108 @@ public function selectThread(?AiThread $thread): void
...
+
+    public function awaitVectorStoreProcessing(AiThread $thread, VectorStoresDataTransferObject $vectorStore): void
+    {
+        $timeout = 60;
+
+        $vectorStoreResponseStatus = $vectorStore->status;
+
+        while ($vectorStoreResponseStatus !== 'completed') {
+            if ($timeout <= 0) {
+                throw new AiThreadLockedException();
```
Orrison, line 317: I think we will want to throw an explicit message here that can describe to the User that it is processing files in the thread and to please wait or something like that.

It will also allow us to better see what errors are being thrown in error reporting.

app-modules/ai/src/Services/Contracts/AiService.php
```diff
@@ -101,5 +117,19 @@ public function retryMessage(AiMessage $message, Closure $saveResponse): Closure
...
 
+    public function retrieveFile(AiMessageFile $file): FilesDataTransferObject;
+
+    public function createVectorStore(array $parameters): VectorStoresDataTransferObject;
+
+    public function retrieveVectorStore(string $vectorStoreId): VectorStoresDataTransferObject;
+
+    public function modifyVectorStore(string $vectorStoreId, array $parameters): void;
+
+    public function retrieveVectorStoreFiles(AiThread $thread, string $vectorStoreId, array $params): VectorStoreFilesDataTransferObject;
```
Orrison, line 128: We may want to consider an additional interface called something like AiServiceWithVectorStores or something like that.

Not all AI services may have vector stores available. But we could also address this later when we encounter one that does not.

Thoughts @danharrin and @dgoetzit?

Orrison (changes requested):
Looking good, just some notes. I would also like @danharrin to review as well.

app-modules/ai/src/Services/Contracts/AiService.php
dgoetzit: These probably could just go into the `SupportsFileUploads` contract since the only reason we'd really want to interact with the stores is for the purpose of uploading/interacting with files.

Orrison (changes requested):
Currently some linting failures

app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -459,6 +459,24 @@ class="ml-auto flex h-6 w-6 cursor-pointer items-center gap-2 rounded-md p-1 tex
...
+                                            <x-filament::badge>
+                                                {{ $file['name'] }}
+                                            </x-filament::badge>
+                                            <x-filament::icon-button
+                                                aria-label="Remove uploaded file {{ $file['name'] }}"
+                                                size="xs"
+                                                icon="heroicon-o-x-mark"
+                                                wire:click="removeUploadedFile({{ $key }})"
+                                            >
+                                                </x-filament::button>
```
danharrin, line 475: This could be refactored to use the `deleteButton` feature as a slot inside the badge, which is the same UI that we use for the TagsInput

```blade
<x-slot
    name="deleteButton"
    label="Remove uploaded file {{ $file['name'] }}"
    wire:click="removeUploadedFile({{ $key }})"
></x-slot>
```

app-modules/ai/src/Actions/SendMessage.php
```diff
@@ -45,7 +45,7 @@
 
 class SendMessage
 {
-    public function __invoke(AiThread $thread, string $content): Closure
+    public function __invoke(AiThread $thread, array $data): Closure
```
danharrin, line 48: What do you think about passing the content and files in two separate parameters, each with their own type?

app-modules/ai/src/Actions/SendMessage.php
```diff
@@ -71,11 +71,13 @@ public function __invoke(AiThread $thread, string $content): Closure
 
         $aiService->ensureAssistantAndThreadExists($thread);
 
-        return $aiService->sendMessage($message, function (AiMessage $response) use ($thread) {
-            $response->thread()->associate($thread);
-            $response->save();
+        return $aiService
+            ->withFiles($data['files'])
```
danharrin, line 75: This makes the AI service stateful, since it stores the files in a property. The AI service is a singleton and can be shared across calls, and this means that in the future if the usage of the service changes it could leak information between requests. I think it would be better to introduce a new parameter to pass the files into `sendMessage()`, probably as the second parameter after the content and before the callback. I think you should introduce a named parameter to the callback argument to provide some clarity.

app-modules/ai/src/Actions/SendMessage.php
danharrin, line 1: Currently, this PR handles sending messages. However, retrying messages after failures use a different controller, action and service method that also needs to be accounted for.

app-modules/ai/src/Services/Contracts/AiService.php
```diff
@@ -56,6 +57,11 @@ public function createAssistant(AiAssistant $assistant): void;
      */
     public function updateAssistant(AiAssistant $assistant): void;
 
+    /**
+     * Override and update the existing tools set on an Assistant
+     */
+    public function updateAssistantTools(AiAssistant $assistant, array $tools): void;
```
danharrin, line 63: I am unsure if the concept of "tools" should leak outside of the Open AI module. How about something more specific for this feature such as `enableAssistantFileUploads()` as part of the `SupportsFileUploads` contract? And then checking for that contract before enabling file uploads.

app-modules/ai/src/Services/Contracts/AiService.php
```diff
@@ -102,4 +114,8 @@ public function retryMessage(AiMessage $message, Closure $saveResponse): Closure
     public function getMaxAssistantInstructionsLength(): int;
 
     public function getDeployment(): ?string;
+
+    public function getApiKey(): string;
+
+    public function getApiVersion(): string;
```
danharrin, line 120: Unlike the deployment, which is used to check if services are interchangeable and share a filesystem, I consider the API key and API version as implementation details of the specific services rather than belonging in this AI module. maybe some services have multiple API keys, or a username and password, or an evergreen backwards compatible API with no version.

By extension, the UploadsFiles trait that consumes these methods probably belongs inside the Open AI service class rather than in the AI module.

app-modules/integration-open-ai/src/Services/OpenAiGptTestService.php
```diff
@@ -45,13 +45,23 @@ public function __construct()
...
     public function getModel(): string
     {
         return 'test';
     }
 
-    public function getDeployment(): ?string
+    public function getDeployment(): string
     {
-        return null;
+        return 'test';
```
danharrin, line 65: This is a fake deployment, so it should return null since none of the data is persisted. If it's null, we don't need to transfer the data to a new service when the assistant model gets updated.

danharrin (changes requested):

app-modules/ai/src/Actions/SendMessage.php
dgoetzit: This is what I had originally done, and it does make sense, but it also then implies files are supported or potentially supported by any service, which is not true. I probably should not have even put this method/property on this class, but on the `OpenAiGpt4oService` since it is the only service currently supporting uploads. I'll change it back for now, but think we should strive for a more granular approach in the future, even with files being fundamentally tied to sending messages.

app-modules/ai/src/Services/Contracts/AiService.php
dgoetzit: For the purposes of the file upload, we still need some way to access the API key and version, as we can't use the Client that ultimately is constructed using these. The `UploadsFile` trait does currently exist strictly in the `integration-open-ai` module. We can resolve the settings within this trait and run a match on the service being used in order to get these values, but that seems like a fairly long way around the same end result.

app-modules/integration-open-ai/src/Services/OpenAiGptTestService.php
dgoetzit: This doesn't actually interact with the API, no? It just seemed a bit strange to enforce a that this potentially returns null when all practical use cases should not.

app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -459,6 +459,22 @@ class="ml-auto flex h-6 w-6 cursor-pointer items-center gap-2 rounded-md p-1 tex
...
+                                                {{ $file['name'] }}
+                                                <x-slot
+                                                    name="deleteButton"
+                                                    label="Remove uploaded file {{ $file['name'] }}"
+                                                    wire:click="removeUploadedFile({{ $key }})"
+                                                ></x-slot>
+                                            </x-filament::badge>
+                                        </div>
+                                    @endforeach
+                                </div>
```
danharrin, line 476: I expect this produces the same results

```suggestion
                                <div class="flex items-center justify-start gap-3 px-4 py-2">
                                    @foreach ($this->files as $key => $file)
                                        <x-filament::badge>
                                            {{ $file['name'] }}
                                            <x-slot
                                                name="deleteButton"
                                                label="Remove uploaded file {{ $file['name'] }}"
                                                wire:click="removeUploadedFile({{ $key }})"
                                            ></x-slot>
                                        </x-
                                    @endforeach
                                </div>
```

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -188,6 +188,8 @@ public function loadFirstThread(): void
         $this->selectThread($this->threadsWithoutAFolder->whereNull('assistant.archived_at')->first());
 
         if ($this->thread) {
+            $this->thread->assistant->model->getService()->afterLoadFirstThread($this->thread);
```
danharrin, line 191: I think we need an interface check before calling this method

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -213,6 +215,8 @@ public function selectThread(?AiThread $thread): void
         }
 
         $this->thread = $thread;
+
+        $this->thread->assistant->model->getService()->afterThreadSelected($this->thread);
```
danharrin, line 219: I think we need an interface check before calling this method

app-modules/ai/src/Services/BaseAiService.php
danharrin, line 1: I don't think we need this base class any more, any services that need the hooks can implement the contract and we can check the service based on that

app-modules/ai/src/Services/TestAiService.php
danharrin, line 1: Since this test service doesn't need the hooks I don't think we need to implement the contract or have the methods

app-modules/ai/src/Services/TestAiService.php
```diff
@@ -96,4 +109,22 @@ public function getDeployment(): ?string
     {
         return null;
     }
+
+    public function supportsFileUploads(): bool
+    {
+        return true;
+    }
+
+    public function createFiles(AiMessage $message, array $files): Collection
```
danharrin, line 118:
```suggestion
    protected function createFiles(AiMessage $message, array $files): Collection
```

app-modules/ai/src/Services/TestAiService.php
```diff
@@ -96,4 +109,22 @@ public function getDeployment(): ?string
...
+    {
+        return collect($files)->map(function ($file) {
```
danharrin, line 120:
```suggestion
        return collect($files)->map(function ($file): AiMessageFile {
```

Can you add the correct param type too, is it SplFileObject or something else?

app-modules/ai/tests/Feature/Actions/RetryMessageTest.php
danharrin, line 1: Maybe the action should have a default files array of `[]` instead of needing to pass it in each time?

app-modules/integration-open-ai/src/Services/Concerns/UploadsFiles.php
danharrin, line 1: Can some of these methods be protected if they are only serving internal purposes?

database/migrations/2024_06_10_164958_data_update_existing_assistants_to_use_file_search_tool.php
```diff
@@ -0,0 +1,76 @@
...
+use AdvisingApp\IntegrationOpenAi\Services\OpenAiGpt4oService;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        $assistants = AiAssistant::query()
+            ->get();
+
+        foreach ($assistants as $assistant) {
+            if ($assistant->model === AiModel::OpenAiGpt4o) {
```
danharrin, line 49: Maybe scope the query instead of using the `if` inside the loop?

danharrin (changes requested):

app-modules/ai/src/Services/TestAiService.php
dgoetzit: Just a string since we aren't actually dealing with a file in the test service.

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanManageThreads.php
```diff
@@ -188,6 +189,10 @@ public function loadFirstThread(): void
         $this->selectThread($this->threadsWithoutAFolder->whereNull('assistant.archived_at')->first());
 
         if ($this->thread) {
+            if ($this->thread->assistant->model->getService() instanceof AiServiceLifecycleHooks) {
+                $this->thread->assistant->model->getService()->afterLoadFirstThread($this->thread);
```
danharrin, line 193: Maybe save the thread service into a var so we don't need to call `getService()` twice? Also below

danharrin (approved):

Orrison (approved):

PR #822, opened 2024-06-14
[ADVAPP - 665] - Enhance failure notifications for engagements

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-665

### Technical Description

Add type in notification when engagements fail.


app-modules/engagement/src/Notifications/EngagementFailedNotification.php
```diff
@@ -72,7 +72,7 @@ public function toDatabase(object $notifiable): array
         return FilamentNotification::make()
             ->danger()
             ->title('Engagement Delivery Failed')
-            ->body("Your engagement failed to be delivered to {$this->engagement->recipient->display_name}.")
+            ->body("Your engagement email failed to be delivered to {$this->engagement->recipient->display_name}.")
```
Orrison, line 75: I believe this failed message is currently shared between email and sms right? So we should check which type the engagement is here and say "sms" or "email" respectively.

Orrison (changes requested):
Had one note on the change. Also it seems like there is something strange going on wit this branch. It seems to contain a strange repeat of a lot of items that have already been merged in.

PR #827, opened 2024-06-17
[ADVAPP-660]: Retire Analytics Resources feature and related models

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-660

### Technical Description

> removed features and related models and permissions.

> yes if found any error.
Please run below command if found any error.
composer remove canyon-gbs/advising-app-analytics
it will remove that module if its there in composer json, otherwise it will update composer and generate auto load files . because we are removing whole module so.


database/migrations/2024_06_17_174458_data_seed_remove_analytic_resources_permissions.php
Orrison, line 1: The prefix for name of this file should be `seed_permissions` not `data_seed`

database/migrations/2024_06_17_181437_data_seed_remove_analytics_resource_sources_permissions.php
Orrison, line 1: The prefix for name of this file should be `seed_permissions` not `data_seed`

database/migrations/2024_06_17_181448_data_seed_remove_analytics_resource_categories_permissions.php
Orrison, line 1: The prefix for name of this file should be `seed_permissions` not `data_seed`

database/migrations/2024_06_17_184610_data_seed_remove_analytics_resource_role.php
Orrison, line 1: This is mostly correct but we will need to do extra steps here.

There are two roles being removed here `analytics.analytics_management` for the `web` guard as well as the one for the `api` guard.

In order to delete each Role I think we are going to have to do the following:
- find that record in the `roles` table. Then find any relations to that role in the `model_has_roles` table and delete them
- Find any relations in the `role_has_permissions` table and delete them
- Delete the record from the `roles` table

For the `down` of this migration you will need to make sure to restore both roles, one for the `api` guard and one for the `web` guard.

You should also change the deleted files `app-modules/analytics/config/roles/api/analytics_management.php` and `app-modules/analytics/config/roles/web/analytics_management.php` to add back any relationship to permissions they should have. I believe the `web` guard one has all of the permissions for this module.

You may also want to have this migration run BEFORE the removal of the permissions migrations. As one the permissions are removed in the subsequent data migrations you technically are in an invalid state as there exists relations in the `role_has_permissions` table that are broken.

Orrison (changes requested):

database/migrations/2024_06_17_184610_data_seed_remove_analytics_resource_role.php
ketan-canyon: Hi @Orrison 
All other changes done. but I think there is no need to add code for remove role for api guard because I checked and found that the web guard one has all of the permissions for this module. see attached images of app-modules/analytics/config/roles/api/analytics_management.php and app-modules/analytics/config/roles/web/analytics_management.php files. still if there is need to add that code to remove api guard role then let me know i will add that.
image
image

database/migrations/2024_06_18_174043_data_seed_remove_analytics_resource_role.php
Orrison, line 1: This needs to be modified to ensure that both the web and the api role with this name is removed from each related table. So that no data is left behind in a broken state.

Orrison (changes requested):

database/migrations/2024_06_17_184610_data_seed_remove_analytics_resource_role.php
Orrison: @ketan-canyon the migrations do need to be modified to specifically find both roles of the same name between both api guards and remove the related tables data.

Regardless of whether or not the api one has permissions or not there could at the very least be models with that role attached.

database/migrations/2024_06_18_174043_data_seed_remove_analytics_resource_role.php
ketan-canyon: Done

database/migrations/2024_06_17_184610_data_seed_remove_analytics_resource_role.php
ketan-canyon: Done

database/migrations/2024_06_18_174043_data_seed_remove_analytics_resource_role.php
```diff
@@ -34,37 +34,40 @@
...
-                    ])
-                    ->columns(),
-            ]);
+        $role_details = DB::table('roles')->where('name', 'analytics.analytics_management')->whereIn('guard_name', ['web', 'api'])->get();
+
+        if (! $role_details->isEmpty()) {
+            foreach ($role_details as $role) {
+                DB::table('model_has_roles')->where('role_id', $role->id)->delete();
+                DB::table('role_has_permissions')->where('role_id', $role->id)->delete();
+                DB::table('roles')->where('name', $role->name)->delete();
```
Orrison, line 50:
```suggestion
                DB::table('roles')->where('id', $role->id)->delete();
```
Please make sure to test this change.

Without doing so the first pass of this will delete both roles

Orrison (changes requested):

database/migrations/2024_06_18_174043_data_seed_remove_analytics_resource_role.php
ketan-canyon: Hi @Orrison 
Changes done and I also tested things its working fine for me. 
see this screen shot there is two roles after run rollback migration.
image
See after again run migration, two roles deleted.
image

PR #836, opened 2024-06-19
[OLYMPUS-41]: Set the filesystem root configs during Tenant create

dgoetzit:
### Ticket(s) or GitHub Issue

https://canyongbs.atlassian.net/browse/OLYMPUS-41

### Technical Description

> Utilizes the tenant name to create the AWS s3 filesystem root definition.

Corresponding Olympus PR: https://github.com/canyongbs/olympus/pull/47

> No


app/Http/Controllers/Tenants/CreateTenantController.php
Orrison, line 1: I think we should append something like a dash and an 8 character random string to the end of the root name. To prevent collisions for Tenants with the same name.

Orrison (changes requested):

Orrison (changes requested):
How would you feel about generating a SQID with the current timestamp like we do for Service Requests in AdvisingApp and AidingApp.

But I think we can use "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789" for the charset, where in AdvisingApp we only used numbers and capital letters.

This would apply the current timestamp as the string appended, but it would reduce the size of the string length and also obscure what the string is

app/Services/SqidGenerator.php
Orrison, line 1: This class, pulled from the other projects, generates a string that is most of the time 14 characters. We don't need to make sure this string is a certain length.

I would prefer that we remove this class and just use a SQID encoded with the current time. And also to use the full a-zA-Z0-9 character set. We used only CAPS and numbers in other project service requests to make the ticket ID consistent.

Orrison (changes requested):

Orrison (approved):

PR #841, opened 2024-06-20
[ADVAPP-629]: Global search produces different results based on capitalization

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-629

### Technical Description

Introduces scoring to order results based on how they match common attributes, to the student and prospect resources.

Exact matches score 100% of the maximum score, matches at the start of the search term score 66% and matches that contain the search term anywhere score 33%. Different attributes have different maximum scores.

### DIFF
--- a/app-modules/student-data-model/src/Filament/Resources/StudentResource.php
+++ b/app-modules/student-data-model/src/Filament/Resources/StudentResource.php
@@ -36,12 +36,12 @@
 
 namespace AdvisingApp\StudentDataModel\Filament\Resources;
 
-use Filament\GlobalSearch\GlobalSearchResult;
 use Filament\Resources\Resource;
 use Filament\Resources\Pages\Page;
-use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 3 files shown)


app-modules/student-data-model/src/Filament/Resources/StudentResource.php
```diff
@@ -96,6 +109,7 @@ public static function getGloballySearchableAttributes(): array
     public static function getGlobalSearchResultDetails(Model $record): array
     {
         return array_filter([
+            'Score' => $record->score,
```
Orrison, line 112: Why is Score added here but not in the one for Prospect?

Orrison (changes requested):

app-modules/student-data-model/src/Filament/Resources/StudentResource.php
danharrin: This is my bad, it was remaining from testing

Orrison (approved):

PR #849, opened 2024-06-25
[ADVAPP-626]: Rename Caseload Management Population Segments

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-626

### Technical Description

> Introduces a new model called `Segment` to replace `Caseload`

> No


database/migrations/2024_06_24_125455_data_migrate_caseloads_to_segments_table.php
```diff
@@ -0,0 +1,91 @@
...
+                        $campaign->save();
+                    }
+                }
+            });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('segments');
+        Schema::dropIfExists('segment_subjects');
```
Orrison, line 88: We shouldn't drop these tables here should we? I think we should probably truncate them.

Orrison (changes requested):
Looks great, just one note on the down of the data migration

database/migrations/2024_06_24_125455_data_migrate_caseloads_to_segments_table.php
dgoetzit: Good catch - will update.

Orrison (approved):

PR #868, opened 2024-07-01
[ADVAPP-573]: Fix unique constraint with deleted Users

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-573

### Technical Description

> Add a restore functionality for the users who are soft deleted and also show status column which shows 'Archived' for the deleted users and 'Active' for non deleted users, and this filter and status column will only be visible if user has permission to restore user. 

This screenshot shows Filter to view users who are deleted.
Screenshot 2024-07-01 185107

This screenshot shows status column and restore icon which will be visible for deleted users.
Screenshot 2024-07-01 185147

This screenshot shows validation message that will be shown if the user who is deleted , his/her email is used again to create a new user.
Screenshot 2024-07-01 185753

### DIFF
--- a/app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
+++ b/app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
@@ -1,0 +1,159 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Advising App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use App\Models\User;
+
+use function Tests\asSuperAdmin;
+use function Pest\Laravel\actingAs;
+use function Pest\Livewire\livewire;
+
+use App\Filament\Resources\UserResource\Pages\ListUsers;
+use App\Filament\Resources\UserResource\Pages\CreateUser;
+use App\Filament\Resources\UserResource\Pages\EditUser;
+use Filament\Tables\Filters\TrashedFilter;
+use Filament\Tables\Actions\RestoreAction;
+
+it('show trashed filter only if user has user restore permission', function () {
+  $user = User::factory()->create();
+
+  $user->givePermissionTo('user.view-any', 'user.*.view', 'user.*.restore');
+  actingAs($user);
+
+  livewire(ListUsers::class)
+    ->assertTableFilterExists(TrashedFilter::class);
+});
+
+it('do not show soft deleted users when filter is not selected', function () {
+  asSuperAdmin();
+  $users = User::factory()->count(3)->create();
+
+  // Soft-delete one of the users
+  $softDeletedUser = $users->first();
+  $softDeletedUser->delete();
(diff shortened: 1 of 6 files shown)


app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
```diff
@@ -0,0 +1,157 @@
...
+use App\Filament\Resources\UserResource\Pages\CreateUser;
+
+it('show trashed filter only if user has user restore permission', function () {
+    $user = User::factory()->create();
+
+    $user->givePermissionTo('user.view-any', 'user.*.view', 'user.*.restore');
+    actingAs($user);
+
+    livewire(ListUsers::class)
+        ->assertTableFilterExists(TrashedFilter::class);
```
Orrison, line 56: We should assert that this table filter does not exist before giving the User the `user.*.restore`.

Though I actually think you will want to check "visible" rather than exist. As it will exist. It just won't be visible.

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
```diff
@@ -0,0 +1,157 @@
...
+    asSuperAdmin();
+    $users = User::factory()->count(3)->create();
+
+    // Soft-delete one of the users
+    $softDeletedUser = $users->first();
+    $softDeletedUser->delete();
+
+    livewire(ListUsers::class)
+        ->filterTable(TrashedFilter::class)
+        ->assertTableColumnExists('deleted_at')
```
Orrison, line 85: You may want to assert `visible` rather than exists here. And you probably should test that the filter is not visible before you apply `filterTable`.

You may be able to do so fluently, but if not you can assign `livewire(ListUsers::class)` to a variable.

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
```diff
@@ -0,0 +1,157 @@
...
+
+    // Soft-delete one of the users
+    $softDeletedUser = $users->first();
+    $softDeletedUser->delete();
+    $softDeletedUserRecords = User::onlyTrashed()->get();
+
+    livewire(ListUsers::class)
+        ->filterTable(TrashedFilter::class, 0)
+        ->assertCanSeeTableRecords($softDeletedUserRecords)
+        ->assertTableActionExists(RestoreAction::class);
```
Orrison, line 105: You will want to check if the action exists but it not visible before giving the `user.*.restore`, then give the user the permission and then test that it is visible

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
```diff
@@ -0,0 +1,157 @@
...
+
+    livewire(ListUsers::class)
+        ->filterTable(TrashedFilter::class, 0)
+        ->assertCanSeeTableRecords($softDeletedUserRecord)
+        ->callTableAction(RestoreAction::class, $trashedUserRecord);
+
+    expect($user->refresh())->deleted_at->toBe(null);
+});
+
+it('check if email uinque validations works properly while creating new user', function () {
```
Orrison, line 125:
```suggestion
it('check if email unique validations works properly while creating new user', function () {
```

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
```diff
@@ -0,0 +1,157 @@
...
+    livewire(CreateUser::class)
+        ->fillForm([
+            'name' => 'Tester',
+            'email' => $user->email,
+        ])
+        ->call('create')
+        ->assertHasFormErrors(['email' => 'unique']);
+});
+
+it('check if email uinque validations works properly while editing user', function () {
```
Orrison, line 141:
```suggestion
it('check if email unique validations works properly while editing user', function () {
```

app/Filament/Resources/UserResource/Pages/EditUser.php
```diff
@@ -76,7 +76,11 @@ public function form(Form $form): Form
                             ->label('Email address')
                             ->email()
                             ->required()
-                            ->maxLength(255),
+                            ->maxLength(255)
+                            ->unique()
```
Orrison, line 80: I think this may cause the edit to fail if you don't change the email address.

You may need to apply `ignoreRecord: true` here

Orrison (changes requested):

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
jayushi-canyon: @Orrison As discussed create new Test for users not having permission to restore.

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
jayushi-canyon: @Orrison As discussed assertTableColumnVisible() not exists.

app-modules/authorization/tests/Feature/Filament/Resources/User/RestoreUserTest.php
jayushi-canyon: @Orrison Done

app/Filament/Resources/UserResource/Pages/EditUser.php
jayushi-canyon: @Orrison Done

jayushi-canyon:
Screenshot 2024-07-03 213450

Screenshot 2024-07-03 213842

Screenshot 2024-07-03 213939

PR #880, opened 2024-07-09
[ADVAPP-718]: Inaccurate validation error presents when creating a new user with existing email address

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-718

### Technical Description

> Change validation message for deleted user email address.

### Screenshots

> 
Screenshot 2024-07-09 170120

>No.

> No.


jayushi-canyon:
> This change looks good, but I have asked a clarifying question to Product in the ticket.

I have made the changes based on the comment you made in ticket.

Orrison (changes requested):
This looks pretty good, but it is causing a test to fail as the test is expecting the `unique` rule specifically to fail. So we need to fix the test in some way. Perhaps you can turn this logic into a custom rule and assert that is what is returned?

PR #882, opened 2024-07-09
Introduce Draft with AI Assistant feature for new email engagement to a prospect or student

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-703

### Technical Description

The chat completion API is being used, since the completion API appears to be getting deprecated. This is a minor inconvenience as we have to format the completion as chat messages, but it works fine.

We cannot use the Open AI PHP Client for this, it appears to be generating a URL for the chat completion endpoint that is not compatible with Azure.

No

No


Orrison (changes requested):
This looks and works great!

We are missing the avatar of the AI Assistant to the left of the text in the modal as described in the ticket though. Let's get that added in.
Also, what do you think the lift is, and what do you think about making the text from the assistant in the modal (Hi [Name], I am happy to help...) appear as if being generated?

Orrison (approved):

PR #891, opened 2024-07-10
[ADVAPP-707]: Create report for the report library Student Engagement

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-707

### Technical Description

> Create report for the report library Student Engagement.

> No

> No


app-modules/student-data-model/src/Filament/Pages/StudentEngagementReport.php
```diff
@@ -0,0 +1,88 @@
...
+
+    protected static ?string $cluster = ReportLibrary::class;
+
+    protected static string $routePath = 'student-engagement-report';
+
+    protected static ?string $navigationIcon = 'heroicon-o-document-text';
+
+    protected static ?string $navigationGroup = 'Students';
+
+    protected $cacheTag = 'prospect-enagement-cache';
```
Orrison, line 58: Let's rename this cache tag to something more indicative that it is for the report page. Like `report-student-engagement`

app-modules/student-data-model/src/Filament/Resources/StudentResource.php
Orrison, line 1: Why was the retention CRM change in sort order here? I don't believe this was in the scope of this ticket but correct me if I am wrong.

app-modules/student-data-model/src/Filament/Widgets/MostEngagedStudentsTable.php
```diff
@@ -0,0 +1,86 @@
...
+    public function refreshWidget()
+    {
+        $this->dispatch('$refresh');
+    }
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->query(
+                Student::select('sisid', 'full_name', 'email')
```
Orrison, line 72: This is not limited to the most engaged 10.

Let's add a `->limit(10)` at the end of the query here and then a `->paginated(false)` to the table itself to accomplish that.

You will also need to sort by engagement count in the query here or you will not be getting the results back from the query that are correct.

The `defaultSort` you applied to the table is just for the table element itself and is probably not needed once you apply it to the query

app-modules/student-data-model/src/Filament/Widgets/MostEngagedStudentsTable.php
```diff
@@ -0,0 +1,86 @@
...
+                    ->withCount('engagements')
+            )
+            ->defaultSort('engagements_count', 'desc')
+            ->columns([
```
Orrison, line 76: Let's remove all `searchable` calls for all columns here. We don't need this table to be searchable.

Orrison (changes requested):

Orrison (approved):

PR #897, opened 2024-07-12
[ADVAPP-714] Enhance the display of the AI utilization report in the Report Library

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-714

### Technical Description

> Fix the chart height issues and improved the grid system.

> No

> Specifically Feature Flags to gate and control logic/flow changes to protect against database schema changes.
> No


Orrison (changes requested):
For the DoughnutChart empty checks, rather than splitting off the data generation into different methods and making specific checks based on the data I think we may be able to make this generic.

What do you think about reverting the `getData` methods back to the way they were and then in `render` checking the result of `getData` directly.

You could do something like this:
```php
$dataSets = $this->getData()['datasets'];
if (
   empty($dataSets)
   || is_array($dataSets) && collect($dataSets)->filter()->isEmpty()
)...
```
That may not be the best or valid code lol, but that's the gist of what I am thinking.

ankit-canyon:
> For the DoughnutChart empty checks, rather than splitting off the data generation into different methods and making specific checks based on the data I think we may be able to make this generic.
> 
> What do you think about reverting the `getData` methods back to the way they were and then in `render` checking the result of `getData` directly.
> 
> You could do something like this:
> 
> ```
> $dataSets = $this->getData()['datasets'];
> if (
>    empty($dataSets)
>    || is_array($dataSets) && collect($dataSets)->filter()->isEmpty()
> )
> ```
> 
> That may not be the best or valid code lol, but that's the gist of what I am thinking.

HI,

I tried refactor the logic again without creating separate function. Please check once.

Thank you.

Orrison (changes requested):
This looks good from a code perspective.

However, we need to adjust the no-data view to still include the widget's heading, and possibly the labels. Please take a look at the screenshot from the ticket below.

image

ankit-canyon:
> This looks good from a code perspective.
> 
> However, we need to adjust the no-data view to still include the widget's heading, and possibly the labels. Please take a look at the screenshot from the ticket below.
> 
> image

It's actually displaying only in SpecialActionsDoughnutChart class because labels are statically given, so to prevent this I added below code :-
```
list($emailCount,$cloneCount) =  $this->getData()['datasets'][0]['data'];

        if ($emailCount == 0 && $cloneCount == 0) {
            return view('livewire.noWidgetData');
        }
```

Orrison (approved):

PR #922, opened 2024-07-24
[ADVAPP-689]: Delete Files and Vector stores in Azure when an AI Thread is deleted

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-689

### Technical Description

Adds a process for clearing out vector stores for Threads before deletion then deletes the Thread in Azure.

Cascades soft deletes from Thread -> Messages -> Files.

Deletes the file in Azure when it is deleted.

No

No.


app-modules/integration-open-ai/src/Services/BaseOpenAiService.php
```diff
@@ -370,6 +375,11 @@ public function supportsAssistantFileUploads(): bool
         return true;
     }
 
+    public function fake()
```
danharrin, line 378:
```suggestion
    public function fake(): void
```

danharrin (changes requested):

danharrin (approved):

PR #923, opened 2024-07-24
[ADVAPP-751]: Enhance the user experience for usage auditing by administrators

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-751

### Technical Description

> Change the label for audit categories from "Personal Assistant" to "Assistant Utilization" and from "Other Records" to "System Administration."
> In the View System Administration section, set "System" in the Change Agent (User) field if it is empty.
> Two new columns, Change Agent (User) and Change Agent (ID), have been added for file exports in System Administration.
> A filter to exclude system users has been added.
> The export button label has been updated from "Export audits" to "Export Records" and from "Export legacy AI message logs" to "Export Records."

> We developed a custom resolver called ChangeAgentNameResolver. Therefore, it may be necessary to restart the queue on the server using the command: php artisan queue:restart.

> Yes. The name of the newly added Feature Flag is "change-agent-name".

### DIFF
--- a/app-modules/audit/database/migrations/2024_07_22_090701_add_change_agent_name_column_to_audits_table.php
+++ b/app-modules/audit/database/migrations/2024_07_22_090701_add_change_agent_name_column_to_audits_table.php
@@ -1,0 +1,21 @@
+<?php
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('audits', function (Blueprint $table) {
+            $table->string('change_agent_name')->nullable();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('audits', function (Blueprint $table) {
+            $table->dropColumn('change_agent_name');
+        });
+    }
+};
--- a/app-modules/audit/database/migrations/2024_07_22_090701_add_change_agent_name_column_to_audits_table.php
+++ b/app-modules/audit/database/migrations/2024_07_22_090701_add_change_agent_name_column_to_audits_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Advising App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 use Illuminate\Support\Facades\Schema;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
--- a/app-modules/audit/database/migrations/2024_07_22_091520_data_activate_change_agent_name.php
+++ b/app-modules/audit/database/migrations/2024_07_22_091520_data_activate_change_agent_name.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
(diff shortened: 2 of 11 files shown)


app-modules/audit/database/migrations/2024_07_22_092622_data_migration_agent_name.php
```diff
@@ -0,0 +1,79 @@
...
+                    DB::table('audits')
+                        ->where('id', $audit->id)
+                        ->update(['change_agent_name' => $audit->name]);
+                }
+
+                DB::commit();
+            } catch (Exception $e) {
+                DB::rollBack();
+
+                throw new Exception($e);
```
Orrison, line 63: You can just throw the exception again. You don't need to wrap it in a new one.

In fact, you don't really want to be wrapping it in a new `Exception` as that will cover up what the actual exception class is making it hard to get debugging information.

app-modules/audit/database/migrations/2024_07_22_092622_data_migration_agent_name.php
```diff
@@ -0,0 +1,79 @@
...
+                $audits = DB::table('audits')
+                    ->join('users', 'audits.change_agent_id', '=', 'users.id')
+                    ->select('audits.id', 'users.name')
+                    ->get();
+
+                foreach ($audits as $audit) {
+                    DB::table('audits')
+                        ->where('id', $audit->id)
+                        ->update(['change_agent_name' => $audit->name]);
+                }
```
Orrison, line 57: Due to the size of the audits table we should `chunkById` here instead of `get`. https://laravel.com/docs/10.x/queries#chunking-results

We may crash memory with a `get`

A good chunk size may be something like 100.

app-modules/audit/database/migrations/2024_07_22_092622_data_migration_agent_name.php
```diff
@@ -0,0 +1,79 @@
...
+                DB::rollBack();
+
+                throw new Exception($e);
+            }
+        }
+    }
+
+    public function down(): void
+    {
+        if (Feature::active('change-agent-name')) {
```
Orrison, line 70: I don't think the feature check is needed here. We know it will be active since the previous migration will have already been run.

app-modules/audit/src/Filament/Resources/AuditResource/Pages/ListAudits.php
```diff
@@ -63,14 +68,30 @@ public function table(Table $table): Table
                 TextColumn::make('auditable_type')
                     ->label('Auditable')
                     ->sortable(),
-                TextColumn::make('user.name')
+                TextColumn::make('change_agent_name')
                     ->label('Change Agent (User)')
-                    ->sortable(),
+                    ->sortable()
+                    ->default('System')
+                    ->visible(Feature::active('change-agent-name')),
```
Orrison, line 75: If the `change-agent-name` feature is not yet active, this will result in this column being missing, which would be considered a regression.

We need to one of the following options here:
1. Modify this column to display the name of the `user` relation if `change-agent-name` is NOT active. Like it did before
2. Add `change_agent_name` as its own column that is only visible if `change-agent-name` is active. And leave the `user.name` column but only make it visible if `change-agent-name` is NOT active. We would then remove the `user.name` column when we did a cleanup of the `change-agent-name` flag.

app-modules/audit/src/Filament/Resources/AuditResource/Pages/ListAudits.php
```diff
@@ -63,14 +68,30 @@ public function table(Table $table): Table
...
                 TextColumn::make('event')
                     ->label('Event')
                     ->sortable(),
             ])
             ->filters([
+                Filter::make('exclude_system_user')
+                    ->label('Exclude System User')
+                    ->query(function (Builder $query, array $data): Builder {
+                        if (isset($data['isActive']) && $data['isActive']) {
```
Orrison, line 84: These two checks can be combined into `!empty($data['isActive'])`

app-modules/audit/src/Filament/Resources/AuditResource/Pages/ListAudits.php
```diff
@@ -63,14 +68,30 @@ public function table(Table $table): Table
...
+                            $query->whereHas('user');
+                        }
+
+                        return $query;
```
Orrison, line 88: There is no need to return the `$query` here

app-modules/audit/src/Filament/Resources/AuditResource/Pages/ViewAudit.php
```diff
@@ -55,9 +58,10 @@ public function infolist(Infolist $infolist): Infolist
                     ->schema([
                         TextEntry::make('auditable_type')
                             ->label('Auditable'),
-                        TextEntry::make('user.name')
+                        TextEntry::make('change_agent_name')
```
Orrison, line 61: Same as in the Edit page. `user.name` still needs to work and be present before the migrations finish and the feature is activated.

Orrison (changes requested):

app-modules/audit/database/migrations/2024_07_22_090701_add_change_agent_name_column_to_audits_table.php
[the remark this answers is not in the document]
payal-canyon: It's already in the audit module. 🤔

app-modules/audit/database/migrations/2024_07_22_091520_data_activate_change_agent_name.php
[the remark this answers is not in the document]
Orrison: Ah. I apologize. I was mistaken

app-modules/audit/database/migrations/2024_07_22_092622_data_migration_agent_name.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public function down(): void
+    {
+        try {
+            DB::table('audits')
+                ->update(['change_agent_name' => null]);
+        } catch (Exception $error) {
+            throw $error;
+        }
```
Orrison, line 73: Because all you are doing is throwing the error you are catching here, this `try/catch` is not needed and can be removed.

Orrison (changes requested):

Orrison (approved):

PR #927, opened 2024-07-25
[ADVAPP-759]: Fix Campaign Engagement images

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-759

### Technical Description

Fixes campaign email images. Fixes actually sending emails and SMSes from campaigns. Fixes service request actions in campaigns. Makes it less likely for image placeholders to stick around after an image is uploaded.

No

No


app-modules/campaign/src/Filament/Blocks/ServiceRequestBlock.php
```diff
@@ -84,17 +85,13 @@ public function generateFields(string $fieldPrefix = ''): array
...
+                ->model(ServiceRequest::class)
                 ->label('Priority')
                 ->required()
                 ->exists((new ServiceRequestPriority())->getTable(), 'id'),
-            Select::make($fieldPrefix . 'type_id')
-                ->relationship('type', 'name')
-                ->preload()
-                ->label('Type')
-                ->required()
-                ->exists((new ServiceRequestType())->getTable(), 'id'),
```
Orrison, line 95: Why was this removed? Is this the cause of Service Request Campaigns not working?

If so, we still need to be able to define the type and priority here. So we will need to revert this and create a separate ticket for fixing this.

Unless it prevents the fix for Campaign Engagement images

Orrison (changes requested):

Orrison (approved):

app-modules/campaign/src/Filament/Blocks/ServiceRequestBlock.php
Orrison: As discussed, we will create a seperate ticket to address fixing this more in-depth in a future iteration.

PR #941, opened 2024-08-02
[ADVAPP-746]: Add error handling to AI Chat Completions

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-746

### Technical Description

When a connection error occurs, or when the API returns an unsuccessful status code, or where the JSON does not contain the correct key, an exception is thrown from the service. This is caught in the action implementation to send a notification to the user informing them of the error. The modal remains open for easy retries in case the service or connection is being flaky, there is no data loss from the prompt.

No

No


Orrison (changes requested):
This looks good but currently contains merge conflicts. I will work on resolving them soon.

Orrison (approved):

PR #943, opened 2024-08-05
[ADVAPP-784]: Restrict the display of report categories and reports based on a user's license settings

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-784

### Technical Description

> Create traits to show report groups based on the license settings logged in user has.

> No.

> No.

### DIFF
--- a/app-modules/report/src/Abstract/AiReport.php
+++ b/app-modules/report/src/Abstract/AiReport.php
@@ -34,17 +34,18 @@
 </COPYRIGHT>
 */
 
-namespace AdvisingApp\Report\Traits;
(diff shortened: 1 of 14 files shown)


Orrison (changes requested):
Per our chat in the group, let's change this to be an abstract class that extends `Filament\Pages\Dashboard` instead of a Trait. And then make sure that all reports us this pattern. Creating a abstract class per grouping, similar to what you have done with the traits here.

Orrison (approved):

PR #954, opened 2024-08-12
[ADVAPP-802] Update default logo images used for light and dark mode in the product

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-802

### Technical Description

> Updated new logo for advising app.

> No

> No


Orrison (changes requested):
@ankit-canyon I see that the assets that were provided to us were in SVG format. But these images that we have swapped out here are PNG.

Did you convert the SVG to PNG? If so, why?

Is it possible to use the SVG versions provided directly?

These new PNG versions are much smaller in resolution and therefore seem to be slightly blurry towards some edges.

ankit-canyon:
> @ankit-canyon I see that the assets that were provided to us were in SVG format. But these images that we have swapped out here are PNG.
> 
> Did you convert the SVG to PNG? If so, why?
> 
> Is it possible to use the SVG versions provided directly?
> 
> These new PNG versions are much smaller in resolution and therefore seem to be slightly blurry towards some edges.

Yeah ! as previous default files are in png format hence I converted these files too. Also with svg I need update some other places as well to get .svg file logos instead of .png

Orrison (approved):

PR #960, opened 2024-08-14
[ADVAPP-731] Improve the underlying method used for rendering multi-factor authentication badges on the user page

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-731

### Technical Description

> MFA badge will display with render hook now.

> No

> No


app/Filament/Resources/UserResource/Pages/ViewUser.php
```diff
@@ -94,26 +97,36 @@ public function form(Form $form): Form
             ]);
     }
 
+    public function boot()
+    {
+        FilamentView::registerRenderHook(
+            PanelsRenderHook::PAGE_HEADER_ACTIONS_AFTER,
+            fn (): View => view('filament.resources.user-resource.mfa-badge')
+                ->with(['class' => 'flex items-center space-x-2']),
```
danharrin, line 105: Please remove the class data, you are not using it

app/Filament/Resources/UserResource/Pages/ViewUser.php
```diff
@@ -94,26 +97,36 @@ public function form(Form $form): Form
...
+            //     ->label(fn (User $record) => match (true) {
+            //         $record->hasConfirmedMultifactor() => 'MFA Enabled',
+            //         $record->hasEnabledMultifactor() => 'MFA Enabled | Not Confirmed',
+            //         default => 'MFA Disabled',
+            //     })
+            //     ->color(fn (User $record) => match (true) {
+            //         $record->hasConfirmedMultifactor() => 'success',
+            //         $record->hasEnabledMultifactor() => 'warning',
+            //         default => 'gray',
+            //     }),
```
danharrin, line 129: I think this should be removed completely instead of commented

resources/views/filament/resources/user-resource/mfa-badge.blade.php
```diff
@@ -0,0 +1,49 @@
...
+                ? 'MFA Enabled | Not Confirmed'
+                : 'MFA Disabled')"
+        :color="$this->getRecord()->hasConfirmedMultifactor()
+            ? 'success'
+            : ($this->getRecord()->hasEnabledMultifactor()
+                ? 'warning'
+                : 'gray')"
+    >
+        {{ $this->getRecord()->hasConfirmedMultifactor() ? 'MFA Enabled' : ($this->getRecord()->hasEnabledMultifactor() ? 'MFA Enabled | Not Confirmed' : 'MFA Disabled') }}
+    </x-filament::badge>
```
danharrin, line 48: I think these would be more readable as 3 completely separate badges wrapped in `@if @elseif @else @endif`, since there are nested conditions

danharrin (changes requested):

app-modules/authorization/tests/Feature/Filament/Resources/User/ViewUserTest.php
```diff
@@ -124,59 +124,6 @@
     expect(auth()->id())->toBe($second->id);
 });
 
-it('does not display the mfa_status Action for an external User', function () {
```
Orrison, line 127: @danharrin and @ankit-canyon do y'all think there is going to be any easy way to replicate these tests for this new way of doing it?

Orrison (changes requested):

app-modules/authorization/tests/Feature/Filament/Resources/User/ViewUserTest.php
danharrin: You can at least assert the text with `assertSee()`

app-modules/authorization/tests/Feature/Filament/Resources/User/ViewUserTest.php
```diff
@@ -124,30 +124,31 @@
...
 
     asSuperAdmin();
 
     livewire(ViewUser::class, [
         'record' => $user->getRouteKey(),
     ])
-        ->assertActionHidden('mfa_status');
+        ->assertDontSeeText('MFA Enabled')
+        ->assertDontSeeText('MFA Enabled | Not Confirmed')
+        ->assertDontSeeText('MFA Disabled');
```
Orrison, line 137: @ankit-canyon, after giving this some more thought. What do you think about applying something like a data attribute to the MFA badge in the HTML and asserting that it exists with something like `assertSeeHtml('data-identifier="mfa-enabled"')` and like assertDontSeeHtml('data-identifier="mfa-enabled-not-confirmed"')`.

This way we are being very explicit about testing that the elemetns we expect are on the page or not on the page rather than checking whether this string exists.

Orrison (changes requested):

Orrison (approved):

Orrison:
Looks great! @danharrin, please re-review to clear your change request.

danharrin (approved):

PR #973, opened 2024-08-19
[ADVAPP-800]: Create new integration settings for Thesis Elements SIS integration

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-800

### Technical Description

Adds integration settings for Student Information Systems, sets up Thesis Elements as the first option. Then displays sync data and a sync button on the Student view page.

No.

Feature Flags:
- `SisIntegrationSettings`

### DIFF
--- a/_ide_helper_models.php
+++ b/_ide_helper_models.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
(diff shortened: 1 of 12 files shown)


danharrin (changes requested):
Can we have a discussion about how much knowledge Advising App needs to have about the actual SIS system? I think it should be informed by Olympus whether or not an SIS system is connected, but it probably doesn't need to know which one. What do you think?

Orrison:
> Can we have a discussion about how much knowledge Advising App needs to have about the actual SIS system? I think it should be informed by Olympus whether or not an SIS system is connected, but it probably doesn't need to know which one. What do you think?

Yeah I get what you are saying. But I think the intention of Advising App knowing what it is integrated with is so that college admins with access only to Advising App would see details pertaining to their SIS. This information would still be dictated by and eventually synced from Olympus

danharrin (approved):

PR #975, opened 2024-08-20
[OLYMPUS-84]: Enhance the partner experience by allowing partners on the same shared environment to each have and apply unique UI/UX branding experiences

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/OLYMPUS-84

### Technical Description

Introduces per-tenant theming through the `ThemeSettings` table. Replaces the old brand settings that affected the entire application.

No

Yes, `synced_theme_settings`


app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -179,10 +177,26 @@ public function panel(Panel $panel): Panel
...
+                if (! Tenant::current()) {
+                    return;
+                }
+
+                if (! Feature::active('synced_theme_settings')) {
+                    return;
+                }
+
+                $panel->darkMode(app(ThemeSettings::class)->has_dark_mode);
+            });
```
Orrison, line 199: With us still referencing the settings class here does this mean we still can only run composer commands in the container?

docs/local-setup.md
```diff
@@ -94,7 +94,6 @@ php artisan queue:restart
 php artisan schedule:interrupt
 npm ci
 php artisan filament:upgrade
-php artisan app:build-assets
```
Orrison, line 97: Should this be replaced with `npm run build`? Or is the thought that `npm run build` will have been run in the container by s6?

Orrison (changes requested):

app/Providers/Filament/AdminPanelProvider.php
danharrin: I don't believe so. All references are within functions. The colors function only needs to be called on render to output the layout. The render hooks function only gets called on render. The boot function only runs in middleware on pages within that panel.

app-modules/theme/src/Settings/ThemeSettings.php
```diff
@@ -37,15 +37,34 @@
...
 {
     public bool $is_logo_active;
 
     public bool $is_favicon_active;
 
+    public array $color_overrides = [];
+
+    public bool $has_dark_mode = true;
+
+    public ?string $url = null;
```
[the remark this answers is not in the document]
danharrin: I imagine that we should remove the defaults methods in the future, as part of the feature cleanup process. The idea here is that we can remove them without needing to go back to the properties to tidy up and add the defaults also. Wdyt? Still remove them from here?

docs/local-setup.md
Orrison: Nope, I think that's right just wanted to confirm.

app-modules/theme/src/Settings/ThemeSettings.php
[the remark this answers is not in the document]
danharrin: done

Orrison (approved):

PR #994, opened 2024-09-03
[ADVAPP-838]: When uploading a partner logo of the PNG type the logo renders with distorted proportions.

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-838

### Technical Description

> Set static position for header logo.

> No.

> No.

### DIFF
--- a/resources/views/vendor/filament-panels/components/logo.blade.php
+++ b/resources/views/vendor/filament-panels/components/logo.blade.php
@@ -50,7 +50,7 @@
         ) }}"
         alt="{{ config('app.name') }}"
         @class([
-            'w-56 h-9',
+            'h-full max-h-9 max-w-full',
             'dark:hidden' => $darkLogo,
         ])
     />
@@ -62,18 +62,18 @@
                 conversionName: 'logo-height-250px',
             ) }}"
             alt="{{ config('app.name') }}"
-            class="w-56 h-9 hidden dark:block"
+            class="h-full max-h-9 max-w-full hidden dark:block"
         />
     @endif
 @else
     <img
         src="{{ Vite::asset('resources/images/default-logo-light.svg') }}"
-        class="w-56 h-9 dark:hidden block"
+        class="h-full max-h-9 max-w-full dark:hidden block"
 
     />
 
     <img
         src="{{ Vite::asset('resources/images/default-logo-dark.svg') }}"
-        class="w-56 h-9 hidden dark:block"
+        class="h-full max-h-9 max-w-full hidden dark:block"
     />
 @endif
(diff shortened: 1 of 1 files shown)


Orrison (changes requested):
@jayushi-canyon you may be right here, but can you explain your reasoning behind adding the `w-56`.

This does need to work for any kind of logo uploaded in the future and I am just concerned that this with was added to possibly be specific to the example image.

The goal is to ensure that no matter what is uploaded that the aspect ration of the image is preserved so the image does not look distorted.

Perhaps we need something like a max width or perhaps a max height? And just let the width size accordingly.

@danharrin if you can weight in as well.

jayushi-canyon:
> @jayushi-canyon you may be right here, but can you explain your reasoning behind adding the `w-56`.
> 
> This does need to work for any kind of logo uploaded in the future and I am just concerned that this with was added to possibly be specific to the example image.
> 
> The goal is to ensure that no matter what is uploaded that the aspect ration of the image is preserved so the image does not look distorted.
> 
> Perhaps we need something like a max width or perhaps a max height? And just let the width size accordingly.
> 
> @danharrin if you can weight in as well.

I kept w-52, by considering that when any image is uploaded for the logo, It will always be the same width for any image.

jayushi-canyon:
> > > @jayushi-canyon you may be right here, but can you explain your reasoning behind adding the `w-56`.
> > > This does need to work for any kind of logo uploaded in the future and I am just concerned that this with was added to possibly be specific to the example image.
> > > The goal is to ensure that no matter what is uploaded that the aspect ration of the image is preserved so the image does not look distorted.
> > > Perhaps we need something like a max width or perhaps a max height? And just let the width size accordingly.
> > > @danharrin if you can weight in as well.
> > 
> > 
> > @Orrison I kept w-52, by considering that when any image is uploaded for the logo, It will always be the same width for any image.
> 
> Got it. Hmm but I don't think we want to force both dimensions on the image as that will not enforce the aspect ration of all logos.
> 
> Have you tried keeping the height as defined and then applying something like `w-auto`?

@Orrison Well rather than keeping w-auto, I think max-w-52 will make more sense as I tested the images with smaller width will not get distorted and one will larger will be set at w-52.

danharrin:
Does `max-w-full` work instead to allow the width to be dynamic?

jayushi-canyon:
> Does `max-w-full` work instead to allow the width to be dynamic?

@danharrin I think 'max-w-full' will create same issue as 'w-auto' as the images with larger width looks bigger with less padding where else I think as i have tested 'max-w-52' shows the logo with proper spacing, this would make sure no image goes beyond this pixel. What do you think?

danharrin:
How about `h-full max-h-9 max-w-full`?

jayushi-canyon:
> h-full max-h-9 max-w-full

@danharrin  I tested this classes with 3 different images :
Screenshot 2024-09-04 212316
Screenshot 2024-09-04 212327
Screenshot 2024-09-04 212341

danharrin:
That looks good to me, what do you think?

Orrison (approved):

PR #995, opened 2024-09-03
[ADVAPP-832] Resolve the security issue tracked under CVE-2024-45048

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-832

### Technical Description

> Resolve the security issue tracked under CVE-2024-45048, CVE-2024-45046

> No.

>No.

### DIFF
--- a/app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
+++ b/app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
@@ -53,7 +53,7 @@ class GenerateTipTapBodyJson
                 foreach ($node->content as $item) {
                     preg_match_all('/{{2}[\s\S]*?}{2}|\s*\S+\s*/', $item->text, $tokens);
 
-                    if (!$tokens && $tokens == null) {
+                    if (! $tokens && $tokens == null) {
                         continue;
                     }
 
--- a/app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
+++ b/app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
@@ -53,7 +53,7 @@ class GenerateTipTapBodyJson
                 foreach ($node->content as $item) {
                     preg_match_all('/{{2}[\s\S]*?}{2}|\s*\S+\s*/', $item->text, $tokens);
 
-                    if (empty($tokens)) {
+                    if (!$tokens && $tokens == null) {
                         continue;
                     }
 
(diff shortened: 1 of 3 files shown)


app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
```diff
@@ -53,7 +53,7 @@ public function __invoke(string $body, array $mergeTags = []): array
                 foreach ($node->content as $item) {
                     preg_match_all('/{{2}[\s\S]*?}{2}|\s*\S+\s*/', $item->text, $tokens);
 
-                    if (empty($tokens)) {
+                    if (! $tokens && $tokens == null) {
```
Orrison, line 56: What is this change for?

composer.json
```diff
@@ -106,7 +106,8 @@
         "talkroute/message-segment-calculator": "^1.0",
         "tapp/filament-timezone-field": "^3.0",
         "tpetry/laravel-postgresql-enhanced": "^0.40",
-        "twilio/sdk": "^7.15"
+        "twilio/sdk": "^7.15",
+        "phpoffice/phpspreadsheet": "dev-security-patch as 1.29.1"
```
Orrison, line 110: This does not seem to be in the correct spot alphabetically is it?

If you had ran `composer require phpoffice/phpspreadsheet` first it would have put it in the correct spot with `*` as the version. Then you can change it to `dev-security-patch as 1.29.1` and run update.

Or just move this to the right spot, either way.

Orrison (changes requested):

app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
ankit-canyon: Hi @Orrison 

On previous if condition Lint error is throwing : 

` Variable $tokens in empty() always exists and is not falsy.
` 
 Hence I need to change it from 
 
` if (empty($tokens)) `
 
 to
 
` if (! $tokens && $tokens == null) `

composer.json
ankit-canyon: Apologies ! Corrected the order in recent changes !

Orrison (changes requested):
@ankit-canyon apologies but it looks like as of this morning they just released the official patch.

So let's actually remove the override here for `phpoffice/phpspreadsheet` and instead install the update to our package: https://github.com/SpartnerNL/Laravel-Excel/releases/tag/3.1.57

app-modules/engagement/src/Actions/GenerateTipTapBodyJson.php
danharrin: This line reads weird to me, is it not the same as `if (blank($tokens)) {` which will check for an empty array or `null`?

danharrin (changes requested):

Orrison (approved):

PR #1001, opened 2024-09-04
[ADVAPP-815]: Eliminate the display issues in all places where the application name is used

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-815

### Technical Description

> Added Application Name in Global Setting's Partner Branding Setting.

> No

> No.


app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
```diff
@@ -73,6 +75,17 @@ public function form(Form $form): Form
...
         return $form
             ->schema([
+                Section::make('Application Name')
+                    ->aside()
+                    ->schema([
+                        TextInput::make('application_name')
+                            ->label(false)
+                            ->maxLength('255')
+                            ->model(
+                                ThemeSettings::getSettingsPropertyModel('theme.application_name'),
```
danharrin, line 85: I don't think the application name needs a model as its just a settings property & not media

resources/views/vendor/mail/html/header.blade.php
danharrin, line 1: If the tenancy switch task executed before this file is rendered? If so, are the switches from `config('app.name')` required, since it should be set correctly already?

app-modules/theme/database/migrations/2024_09_04_115229_add_application_name_to_themes_settings.php
```diff
@@ -0,0 +1,49 @@
...
+
+</COPYRIGHT>
+*/
+
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        $this->migrator->add('theme.application_name', 'Advising App');
```
danharrin, line 42: I think we spoke in planning about the default app name being `config('app.name')`, is there a reason we decided against that?

danharrin (changes requested):

app-modules/theme/database/migrations/2024_09_04_115229_add_application_name_to_themes_settings.php
payal-canyon: Hi @danharrin 
I thought it would be Advising App as in the tech planning notes, the default app name is given.

Orrison (changes requested):
This is looking good. There are currently some merge conflicts to be resolved

Orrison (approved):

Orrison (changes requested):
We will need to change this to work with the TenantConfig

app/Multitenancy/Tasks/SwitchAppName.php
```diff
@@ -0,0 +1,70 @@
...
+    ) {
+        $this->originalAppName ??= config('app.name');
+    }
+
+    public function makeCurrent(Tenant $tenant): void
+    {
+        /** @var TenantConfig $config */
+        $config = $tenant->config;
+
+        $this->originalAppName = $config->applicationName ?? config('app.name');
```
Orrison, line 56: You should not set `$this->originalAppName` here.

It is set in the constructor so that it can be reset to it during the `forgetCurrent` call.

app/Multitenancy/Tasks/SwitchAppName.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $this->setAppName($this->originalAppName);
```
Orrison, line 58: This should be set to `$config->applicationName` if it is not empty

app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
```diff
@@ -73,6 +78,14 @@ public function form(Form $form): Form
     {
         return $form
             ->schema([
+                Section::make('Application Name')
+                    ->aside()
+                    ->schema([
+                        TextInput::make('application_name')
+                            ->label(false)
+                            ->dehydrated(fn (?string $state): bool => filled($state))
+                            ->maxLength('255'),
```
Orrison, line 87: This should be required and can be dehydrated even if it was just filled with the default. That way on save it actually gets set, which is not a problem.

Orrison (changes requested):

app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
```diff
@@ -73,6 +78,15 @@ public function form(Form $form): Form
     {
         return $form
             ->schema([
+                Section::make('Application Name')
+                    ->aside()
+                    ->schema([
+                        TextInput::make('application_name')
+                            ->label('Application Name')
+                            ->required()
+                            ->dehydrated(fn (?string $state): bool => filled($state))
```
Orrison, line 87: Is this `dehydrated` still needed? Because this field is now required it would never not be `filled` right?

Orrison (changes requested):

Orrison (approved):

PR #1006, opened 2024-09-09
[ADVAPP-837]: When uploading a partner logo of the SVG type, the logo fails to render in product.

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-837

### Technical Description

> Solve .svg photo not getting display in header issue,

> No.

> No.


Orrison (changes requested):
@jayushi-canyon though the changes to `max-w-full` to `w-full` may resolve the size issue for the SVG somewhat I worry that this will affect the aspect ratio issue both for the SVG and definitely for any other image type uploaded.

jayushi-canyon:
> @jayushi-canyon though the changes to `max-w-full` to `w-full` may resolve the size issue for the SVG somewhat I worry that this will affect the aspect ratio issue both for the SVG and definitely for any other image type uploaded.

@Orrison I did some testing using a .png and also .svg image and some other, they didn't quite gave any issue but if you think so I will test it again with some different images. As the issue with this .svg issue was it was not getting enough width.

danharrin:
I'm trying to review this, but can you please explain the logic behind the conditional width classes? What's the issue there? It looks like SVGs now have `w-full` instead of `max-w-full`, but the old code also used `w-full`? So is that to fix non-SVGs instead of SVGs?

Orrison:
> I'm trying to review this, but can you please explain the logic behind the conditional width classes? What's the issue there? ~It looks like SVGs now have `w-full` instead of `max-w-full`, but the old code also used `w-full`? So is that to fix non-SVGs instead of SVGs?~

@danharrin the reasoning for my suggestion was an issue we saw when SVG logos were used. Because an SVG does not typically have sizing applied the SVG is none existent because it ends up with a width of `0`

resources/views/vendor/filament-panels/components/logo.blade.php
```diff
@@ -43,26 +43,38 @@
...
+            conversionName: $logo->hasGeneratedConversion('logo-height-250px') ? 'logo-height-250px' : '',
+        );
+        $class = ($logo->mime_type == 'image/svg+xml') ? 'h-full max-h-9 w-full' : 'h-full max-h-9 max-w-full';
+    @endphp
+    <img
+        src="{{ $logoUrl }}"
         alt="{{ config('app.name') }}"
         @class([
-            'h-full max-h-9 max-w-full',
+            $class,
```
danharrin, line 57: You can drop the variable and do this

```suggestion
            'h-full max-h-9',
            'w-full' => $logo->mime_type == 'image/svg+xml',
            'max-w-full' => $logo->mime_type != 'image/svg+xml',
```

resources/views/vendor/filament-panels/components/logo.blade.php
```diff
@@ -43,26 +43,38 @@
...
+            );
+            $class = ($darkLogo->mime_type == 'image/svg+xml') ? 'h-full max-h-9 w-full' : 'h-full max-h-9 max-w-full';
+        @endphp
+       
+        <img
+            src="{{ $darkLogoUrl }}"
             alt="{{ config('app.name') }}"
-            class="h-full max-h-9 max-w-full hidden dark:block"
+            @class([
+            $class,
```
danharrin, line 75: You can drop the variable and do this

```suggestion
            'h-full max-h-9',
            'w-full' => $darkLogo->mime_type == 'image/svg+xml',
            'max-w-full' => $darkLogo->mime_type != 'image/svg+xml',
```

There are also indentation issues on this `@class` directive, see the correct example above

danharrin (changes requested):

danharrin (approved):

Orrison (approved):

PR #1009, opened 2024-09-11
[ADVAPP-844]: Track portal views of articles and display them as article metadata

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-844

### Technical Description

>  Create new feature where every time the article is clicked for view it count increases and end user can see total number of view on that article in the view article page.

> No.

> `portal_view_count`

### DIFF
--- a/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
+++ b/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
@@ -34,12 +34,11 @@
 </COPYRIGHT>
 */
 
-use Illuminate\Database\Migrations\Migration;
-use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
 
-return new class extends Migration
-{
+return new class () extends Migration {
     public function up(): void
     {
         Schema::table('knowledge_base_articles', function (Blueprint $table) {
--- a/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
+++ b/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
@@ -1,0 +1,22 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
+{
+    public function up(): void
+    {
+        Schema::table('knowledge_base_articles', function (Blueprint $table) {
+            $table->integer('portal_view_count')->after('category_id')->default(0);
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::table('knowledge_base_articles', function (Blueprint $table) {
+            $table->dropColumn('portal_view_count');
+        });
+    }
+};
--- a/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
+++ b/app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
@@ -38,18 +38,18 @@ use Illuminate\Support\Facades\Schema;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Database\Migrations\Migration;
 
-return new class() extends Migration {
-  public function up(): void
-  {
-    Schema::table('knowledge_base_articles', function (Blueprint $table) {
-      $table->integer('portal_view_count')->default(0);
-    });
-  }
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('knowledge_base_articles', function (Blueprint $table) {
+            $table->integer('portal_view_count')->default(0);
+        });
+    }
 
-  public function down(): void
-  {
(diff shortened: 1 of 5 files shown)


danharrin (dismissed):
I think this needs a feature flag before the view counter is bumped

app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('knowledge_base_articles', function (Blueprint $table) {
+            $table->integer('portal_view_count')->after('category_id')->default(0);
```
Orrison, line 45: The `after` is not needed here an can be removed. Postgres does not allow for the ability the add columns at a particular position.

Orrison (changes requested):
+1 on needing the FeatureFlag. The bool of the FeatureFlag will also need to be passed to the JS to dynamically determine the rendering of the elements in the JS.

jayushi-canyon:
> I think this needs a feature flag before the view counter is bumped

Done

app-modules/knowledge-base/database/migrations/2024_09_10_171824_add_portal_view_count_in_knowledge_base_articles_table.php
jayushi-canyon: Done

jayushi-canyon:
> +1 on needing the FeatureFlag. The bool of the FeatureFlag will also need to be passed to the JS to dynamically determine the rendering of the elements in the JS.

Done

app-modules/knowledge-base/src/Models/KnowledgeBaseArticle.php
```diff
@@ -67,6 +67,7 @@ class KnowledgeBaseArticle extends BaseModel implements Auditable, HasMedia
     protected $fillable = [
         'article_details',
         'category_id',
+        'portal_view_count',
```
Orrison, line 70: This doesn't need to be fillable as we will always just increment. And it is safer to not be

app-modules/portal/src/Http/Controllers/KnowledgeManagement/KnowledgeManagementPortalArticleController.php
```diff
@@ -48,8 +49,14 @@ class KnowledgeManagementPortalArticleController extends Controller
 {
     public function show(KnowledgeBaseCategory $category, KnowledgeBaseArticle $article): JsonResponse
     {
+        $portalViewCountFlag = Feature::active('portal_view_count');
+
         $articleUpdatedAt = $article->updated_at->setTimezone(app(DisplaySettings::class)->timezone);
 
+        if ($portalViewCountFlag) {
+            $article->update(['portal_view_count' => $article->portal_view_count + 1]);
```
Orrison, line 57: We should use `increment` here instead as it is more performant and is an atomic operation: https://laravel.com/docs/11.x/queries#increment-and-decrement

portals/knowledge-management/src/Pages/ViewArticle.vue
```diff
@@ -60,6 +61,8 @@ const props = defineProps({
 const loading = ref(true);
 const category = ref(null);
 const article = ref(null);
+const portalViewCount = ref(0);
+const portalViewCountFlag = ref(0);
```
Orrison, line 65: `ref(false)` probably makes more sense here

Orrison (changes requested):

app-modules/portal/src/Http/Controllers/KnowledgeManagement/KnowledgeManagementPortalArticleController.php
jayushi-canyon: Done

portals/knowledge-management/src/Pages/ViewArticle.vue
jayushi-canyon: Done

portals/knowledge-management/src/Pages/ViewArticle.vue
```diff
@@ -112,9 +116,15 @@ function getData() {
                         <div class="flex flex-col gap-3">
                             <div class="prose max-w-none">
                                 <h1>{{ article.name }}</h1>
-                                <div class="text-gray-500 flex items-center space-x-1 mb-4">
-                                    <ClockIcon class="h-4 w-4 flex-shrink-0" aria-hidden="true" />
-                                    <span class="text-xs">Last updated: {{ article.lastUpdated }}</span>
+                                <div class="flex mb-4">
+                                  <div v-if="portalViewCountFlag" class="text-gray-500 flex items-center space-x-1 mr-2">
+                                      <EyeIcon class="h-4 w-4 flex-shrink-0" aria-hidden="true" />
```
Orrison, line 121: This does not seem to be rendering for me on the frontend? Is it for you?

Orrison (changes requested):
Hmm, shoot. I am seeing the views increment way to much when I navigation to the page or reload the page. It is going up by like 2-3 views everytime I load the page.

portals/knowledge-management/src/Pages/ViewArticle.vue
jayushi-canyon: @Orrison You ran the migrations?

Orrison: @jayushi-canyon I did. To be clear I mean I just don't see the Eye icon rendered on the screen. I think you forgot to import it at the top of the file like you did in AidingApp

PR #1011, opened 2024-09-12
[ADVAPP-848]: Enhance the Thesis Elements integration by allowing an end user user to initiate an on demand sync, and by displaying the status of the last sync time

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-848

### Technical Description

Actually kicks off the sync process when student sync is button is clicked.

No

No


app/Multitenancy/Tasks/SwitchSessionDriver.php
danharrin, line 1: It looks ok, but just checking if this change is intended for this PR?

app/Overrides/Laravel/StartSession.php
danharrin, line 1: As above

app/Providers/AppServiceProvider.php
danharrin, line 1: As above

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/ViewStudent.php
```diff
@@ -165,9 +171,45 @@ public function infolist(Infolist $infolist): Infolist
...
     public function sisRefresh()
     {
+        [$olympusUrl, $olympusKey] = Landlord::execute(function (): array {
+            $settings = app(OlympusSettings::class);
+
+            return [
+                rtrim($settings->url, '/'),
+                $settings->key,
+            ];
+        });
```
danharrin, line 181: Maybe we could have an `Olympus` service similar to `ThesisElements` with a pre-configured `makeRequest()` method that creates a pending request using the `baseUrl()` and auth token. Then it's reusable for future sync requests

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/ViewStudent.php
```diff
@@ -165,9 +171,45 @@ public function infolist(Infolist $infolist): Infolist
...
+        $student = $this->getRecord();
+
+        try {
+            $response = Http::withToken($olympusKey)
+                ->asJson()
+                ->post("{$olympusUrl}/integrations/{$tenantId}/student-on-demand-sync", [
+                    'sisid' => $student->getKey(),
+                    'otherid' => $student->otherid,
+                ])
+                ->throw();
```
danharrin, line 195: The code looks good, but I would suggest putting the actual request in an action class so it can be swapped out for a mock when testing (without needing to fake the entire Http facade). The Filament notification bit is UI-related so should stay in this Livewire component

danharrin (changes requested):

app/Multitenancy/Tasks/SwitchSessionDriver.php
Orrison: It is yes. It is a change we will need to do in Advising App as well.

Without this we are unable to get settings from the Landlord like I am doing, and then dispatch a Notification with Filament. Because the Landlord calls re-call all the Tenant Switch tasks, which before this would wipe the session data.

app/Overrides/Laravel/StartSession.php
Orrison: Commented above

app/Providers/AppServiceProvider.php
Orrison: Commented above

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/ViewStudent.php
Orrison: Yeah I think that is a good idea. I think it can be a lot simpler than `ThesisElements` for now with just a static method. You can let me know what you think.

I think ultimately we will improve this to instead actually come out an "Integration" for this side. We may want to consider AdvisingApp to consider Olympus a SIS integration that could be swapped out

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/ViewStudent.php
Orrison: Like I said above, I think I want to consider in the near future actually creating a SisIntegration concept for Advising App that would be called here instead. But I would like to start at a small increment like this before going much further to get the Feature out.

Knowing that, given how small the portion here is for the HTTP request, and how easy it is to mock fake the response I would propose leaving this as it is for the time being.

What do you think?

danharrin (approved):

PR #1029, opened 2024-09-23
[ADVAPP-885]: Improve the relationship between care team members and students and prospects by auto subscribing users

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-885

### Technical Description

> Applied CanTriggerAutoSubscription interface to the CareTeam Model.

> No

> No


app-modules/care-team/src/Observers/CareTeamObserver.php
```diff
@@ -0,0 +1,53 @@
...
+
+class CareTeamObserver
+{
+    public function created(CareTeam $careTeam): void
+    {
+        $user = auth()->user();
+
+        if ($user instanceof User) {
+            TriggeredAutoSubscription::dispatch($user, $careTeam);
+        }
```
Orrison, line 51: The user to subscribe is the User from the `$careTeam`. Not the currently authed User.

Orrison (changes requested):
Made one comment as the User being subscribed is not right here.

Also, I apologize, I should have included this in the ticket. But if it's not too much more lets add a simple test ensuring that this is working and adding the correct Subscription for the right User when a CareTeam record is created for them. It can be a simple Unit test.

app-modules/care-team/tests/CareTeamAutoSubscriptionTest.php
```diff
@@ -0,0 +1,105 @@
...
+
+    $prospect = Prospect::factory()->create();
+
+    livewire(ManageProspectCareTeam::class, [
+        'record' => $prospect->getKey(),
+    ])
+        ->callTableAction(
+            AttachAction::class,
+            data: ['recordId' => [$careTeamMember->getKey()]]
+        )->assertSuccessful();
```
Orrison, line 68: Rather than testing the Filament component I would suggest we go lower level in this Unit test and just test it by creating the CareTeam Model manually.

app-modules/care-team/tests/CareTeamAutoSubscriptionTest.php
```diff
@@ -0,0 +1,105 @@
...
+
+    $prospect->refresh();
+
+    expect($prospect->careTeam->pluck('id'))->toContain($careTeamMember->getKey());
+
+    assertDatabaseHas('care_teams', [
+        'educatable_id' => $prospect->getKey(),
+        'user_id' => $careTeamMember->getKey(),
+    ]);
```
Orrison, line 77: The `assertDatabaseHas` as most likely sufficient, so there is no need to do the refresh or the expect above. I would just recommend also adding `educatable_type` to the check by getting the `$prospect` (or `$eductable` after my other suggestion) and calling `getMorphClass`

Orrison: Though, actually this test is not supposed to be testing whether or not the `CareTeam` exists. It is supposed to be checking whether or not the Subscription was auto-created for the `User`

app-modules/care-team/tests/CareTeamAutoSubscriptionTest.php
```diff
@@ -0,0 +1,105 @@
...
+use AdvisingApp\Prospect\Models\Prospect;
+use Filament\Tables\Actions\AttachAction;
+
+use function Pest\Laravel\assertDatabaseHas;
+
+use AdvisingApp\StudentDataModel\Models\Student;
+use AdvisingApp\Prospect\Filament\Resources\ProspectResource\Pages\ManageProspectCareTeam;
+use AdvisingApp\StudentDataModel\Filament\Resources\StudentResource\Pages\ManageStudentCareTeam;
+
+it('can auto subscribe user to prospect upon adding in care team', function () {
```
Orrison, line 53: You can reduce this down to a single test in this file with a Dataset. (https://pestphp.com/docs/datasets)

The Dataset would provide an educatable, which can be a one-line Closure returning a factory create for a student and then a prospect and coming in as `$educatable` into the test function.
That would reduce all this code to be able to be run for both the Student and Prospect without duplication

Orrison (changes requested):

app-modules/care-team/tests/CareTeamAutoSubscriptionTest.php
```diff
@@ -0,0 +1,84 @@
...
+                'user_id' => $user->getKey(),
+                'subscribable_id' => $educatable->getKey(),
+                'subscribable_type' => $educatable->getMorphClass(),
+            ]);
+        });
+    });
+})
+    ->with([
+        'for Prospect' => [
+            'careTeam' => fn () => User::factory()->licensed(LicenseType::cases())->count(1)->create()->pluck('id')->toArray(),
```
Orrison, line 77: Especially because this is the exact same for each data set I don't think it is needed to have this as a part of the dataset. It can just be in the test.
It is also a little confusing to call it careteam when it is a User.

Orrison (changes requested):

Orrison (approved):

PR #1039, opened 2024-09-26
[ADVAPP-887] Enhance the timeline experience by tracking interactions

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-887

### Technical Description

> Interactions can be tracked in and viewed in timeline.

> No.

> No.


app-modules/interaction/resources/views/components/interaction-timeline-item.blade.php
```diff
@@ -0,0 +1,82 @@
...
+            {{ $viewRecordIcon }}
+        </div>
+    </div>
+
+    <time class="mb-2 block text-sm font-normal leading-none text-gray-400 dark:text-gray-500">
+        {{ $record?->type?->name }}
+    </time>
+
+    <time class="mb-2 block text-sm font-normal leading-none text-gray-400 dark:text-gray-500">
+        Sent {{ $record->created_at->diffForHumans() }}
```
Orrison, line 64: I think this is actually supposed to be `start_datetime` and `end_datetime`. Also "Sent" doesn't make sense here as it's meant to display the date/time that this interaction happened.

app-modules/interaction/resources/views/components/interaction-timeline-item.blade.php
```diff
@@ -0,0 +1,82 @@
...
+                <p class="text-xs text-gray-400 dark:text-gray-500">Subject:</p>
+                <p>{{ $record->subject }}</p>
+            </div>
+        @endif
+        <div class="flex flex-col">
+            <p class="text-xs text-gray-400 dark:text-gray-500">Description:</p>
+            <div class="prose dark:prose-invert">
+                {{ $record->description }}
+            </div>
+        </div>
```
Orrison, line 80: Description is also `nullable` so we may want somechecks there as well. But we probably could get rid of the checks and just add a null coalesce to display "N/A" if either are empty.

app-modules/interaction/src/Models/Concerns/HasManyMorphedInteractions.php
```diff
@@ -50,4 +50,10 @@ public function interactions(): MorphMany
             id: 'interactable_id'
         );
     }
+
+    public function orderedInteractions(): MorphMany
+    {
+        return $this->interactions()
+            ->orderBy('created_at', 'desc');
```
Orrison, line 57: These should probably be ordered by `start_datetime` rather than `created_at`

app-modules/timeline/src/Timelines/InteractionTimeline.php
```diff
@@ -0,0 +1,74 @@
...
+    ) {}
+
+    public function icon(): string
+    {
+        return 'heroicon-o-arrow-small-right';
+    }
+
+    public function sortableBy(): string
+    {
+        return $this->interaction->created_at;
```
Orrison, line 57: I think we might want to do `start_datetime` here

app-modules/timeline/src/Timelines/InteractionTimeline.php
```diff
@@ -0,0 +1,74 @@
...
+
+class InteractionTimeline extends CustomTimeline
+{
+    public function __construct(
+        public Interaction $interaction
+    ) {}
+
+    public function icon(): string
+    {
+        return 'heroicon-o-arrow-small-right';
```
Orrison, line 52: Should this be `pencil-square`? Or is this icon used for something else?

Orrison (changes requested):

app-modules/interaction/resources/views/components/interaction-timeline-item.blade.php
ankit-canyon: > I think this is actually supposed to be `start_datetime` and `end_datetime`. Also "Sent" doesn't make sense here as it's meant to display the date/time that this interaction happened.

Hmm ! as per screenshot it's displaying this

so, on text as per screenshot `1 week ago` we need to display start_datetime? and where to display the `end_datetime` on pop_up? when click on view icon?

image

Orrison: Ah got it. I don't think we need to do the end_datetime. But rather than making this time based on `created_at` kets do `start_datetime` which is the actual recorded time of the interaction.

If you want you can confirm with Product what is wanted here

app-modules/timeline/src/Timelines/InteractionTimeline.php
Orrison: Hmm, please check with Product what they want here. I am actually not sure now.

Orrison (approved):

PR #1040, opened 2024-09-26
[ADVAPP-883]: Improve message center search by ensuring that it is case insensitive

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-883

### Technical Description

> Improve message center search by ensuring that it is case-insensitive.

> No

> No

### DIFF
--- a/app-modules/engagement/src/Filament/Pages/MessageCenter.php
+++ b/app-modules/engagement/src/Filament/Pages/MessageCenter.php
@@ -39,6 +39,7 @@ namespace AdvisingApp\Engagement\Filament\Pages;
 use Carbon\Carbon;
 use App\Models\User;
 use Filament\Pages\Page;
+use Illuminate\Support\Str;
 use Filament\Actions\Action;
 use Livewire\Attributes\Url;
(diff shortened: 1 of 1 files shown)


app-modules/engagement/src/Filament/Pages/MessageCenter.php
```diff
@@ -337,7 +339,7 @@ protected function getViewData(): array
 
             $studentPopulationQuery = Student::query()
                 ->when($this->search, function ($query, $search) {
-                    $query->where('full_name', 'like', "%{$search}%")
+                    $query->where(new Expression('lower(full_name)'), 'like', '%' . Str::lower($search) . '%')
```
Orrison, line 342: Let's actually play it a little safer here and use binding to reduce the risk of SQL injection. Something like:

```php
->whereRaw('lower(full_name) like ?', ['%' . Str::lower($search) . '%'])
```

Orrison (changes requested):

Orrison (approved):

PR #1072, opened 2024-10-17
[ADVAPP-934]: Introduce the ability to set a prospective student conversion value for ROI analysis

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-934

### Technical Description

> Enable setting a conversion value for prospective students to support ROI analysis. This feature allows tracking and evaluating the financial impact of converting prospective students.

> No

> Yes. `ProspectConversion`


app-modules/prospect/src/Filament/Pages/ManageProspectConversionSettings.php
```diff
@@ -0,0 +1,145 @@
...
+use App\Settings\ProspectConversionSettings;
+use App\Filament\Clusters\ConstituentManagement;
+
+class ManageProspectConversionSettings extends SettingsPage
+{
+    protected static ?string $navigationIcon = 'heroicon-o-currency-dollar';
+
+    protected static string $settings = ProspectConversionSettings::class;
+
+    public $currency = 'USD';
```
Orrison, line 62: Let's add typing here. `string` I assume makes the most sense.

app-modules/prospect/src/Filament/Pages/ManageProspectConversionSettings.php
```diff
@@ -0,0 +1,145 @@
...
+    protected static ?string $navigationGroup = 'Prospect Management';
+
+    protected static ?string $cluster = ConstituentManagement::class;
+
+    public static function canAccess(): bool
+    {
+        /** @var User $user */
+        $user = auth()->user();
+
+        return ProspectConversion::active() && $user->can('prospect_conversion.manage');
```
Orrison, line 77: Let's make sure to include `parent::canAccess()` as well, just in case. Probably best between the feature flag check and this pages permission check.

app-modules/prospect/src/Filament/Pages/ManageProspectConversionSettings.php
```diff
@@ -0,0 +1,145 @@
...
+    public function save(): void
+    {
+        try {
+            $this->callHook('beforeValidate');
+            $data = $this->form->getState();
+
+            $this->callHook('afterValidate');
+            $settings = app(static::getSettings());
+            $this->callHook('beforeSave');
+            $settings->estimated_average_revenue = Money::parseByDecimal($data['estimated_average_revenue'], $this->currency);
```
Orrison, line 114: I think it actually may be better here to get rid of this custom `save` method override and just move this casting to an override of `mutateFormDataBeforeSave`

config/money.php
```diff
@@ -0,0 +1,56 @@
...
+     |--------------------------------------------------------------------------
+     */
+    'locale' => config('app.locale', 'en_US'),
+    'defaultCurrency' => config('app.currency', 'USD'),
+    'defaultFormatter' => null,
+    'defaultSerializer' => null,
+    'isoCurrenciesPath' => __DIR__ . '/../vendor/moneyphp/money/resources/currency.php',
+    'currencies' => [
+        'iso' => 'all',
+        'bitcoin' => 'all',
```
Orrison, line 50: Let's make this an empty array. We are not currently supporting any crypto currencies.

config/money.php
```diff
@@ -0,0 +1,56 @@
...
+     | Laravel money
+     |--------------------------------------------------------------------------
+     */
+    'locale' => config('app.locale', 'en_US'),
+    'defaultCurrency' => config('app.currency', 'USD'),
+    'defaultFormatter' => null,
+    'defaultSerializer' => null,
+    'isoCurrenciesPath' => __DIR__ . '/../vendor/moneyphp/money/resources/currency.php',
+    'currencies' => [
+        'iso' => 'all',
```
Orrison, line 49: Let's change this to an array with just 'USD' in it. That is currently all we are supporting.

database/settings/2024_10_15_082847_prospect_conversion_settings.php
Orrison, line 1: Let's move this to the prospects module

resources/views/filament/forms/components/heading-one.blade.php
Orrison, line 1: Would you mind also bringing over the `heading-two` and `heading-three` blade files from Aiding App over here as well?

Orrison (changes requested):

config/money.php
payal-canyon: Hi @Orrison 

Since the default currency has already been set, we don't need to include it here.

Thank You!

Orrison (changes requested):

app/DataTransferObjects/Casts/MoneySettingCast.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function set($payload): ?array
+    {
+        if (blank($payload) || ! ($payload instanceof Money)) {
+            return null;
+        }
+
+        return [
+            'value' => intval($payload->getAmount()),
```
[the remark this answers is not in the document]
payal-canyon: Hi @Orrison,

As from payload value is coming as `string` and If I'm saving the string without typecasting or `intval()` it causing issue. hence either we need to type cast the value on `get()` function if we remove `intval()` from `set()` function.

Thank You!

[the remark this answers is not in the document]
Orrison: @payal-canyon got it. Hmm, if that is the type that comes out of `getAmount` then that is what we should save it as and use in the `get`

Orrison (changes requested):

Orrison (dismissed):

Orrison (approved):

PR #1077, opened 2024-10-19
[ADVAPP-950]: Login page fails to preserve the padding in the top header when responding to screen size changes

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-950

### Technical Description

> Login page fails to preserve the padding in the top header when responding to screen size changes

> No

> No


Orrison (approved):
Looks great, nice work!

PR #1084, opened 2024-10-24
[ADVAPP-963]: Introduce population segment rule for first generation status

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-963

### Technical Description

Introduce population segment rule for first generation status

No

No


app-modules/student-data-model/src/Filament/Tables/StudentsTable.php
```diff
@@ -102,6 +102,9 @@ public function __invoke(Table $table): Table
                             ->label('SAP')
                             ->icon('heroicon-m-academic-cap'),
                         BooleanConstraint::make('dual'),
+                        BooleanConstraint::make('firstgen')
+                            ->label('First Generation')
+                            ->icon('heroicon-m-lock-open'),
```
Orrison, line 107: Please check with Product to confirm whether or not this is the icon that should be used.

Orrison (changes requested):

app-modules/student-data-model/src/Filament/Tables/StudentsTable.php
ketan-canyon: Icon updated as Product suggest in ticket in jira.

Orrison (approved):

PR #1088, opened 2024-10-25
[ADVAPP-952]: When the disable branding bar option is enabled in profile settings text appears next to the switch that should not be displayed

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-952

### Technical Description

> When the disable branding bar option is enabled in profile settings text appears next to the switch that should not be displayed.

> No.

> No.


kandarp-canyon:
test advisingapp local_profile

PR #1093, opened 2024-10-30
[ADVAPP-965] : Add filters for the campaign list view for my campaigns, enabled, and has been executed

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-965

### Technical Description

> Add filters for the campaign list view for my campaigns, enabled, and has been executed.

> No.

> No.

### DIFF
--- a/app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
+++ b/app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
@@ -38,6 +38,7 @@ namespace AdvisingApp\Campaign\Filament\Resources\CampaignResource\Pages;
 
 use Filament\Tables\Table;
 use Filament\Actions\CreateAction;
+use Filament\Tables\Filters\Filter;
 use Filament\Tables\Actions\EditAction;
 use Filament\Tables\Actions\ViewAction;
 use Filament\Tables\Columns\IconColumn;
@@ -46,6 +47,7 @@ use AdvisingApp\Campaign\Models\Campaign;
 use App\Filament\Tables\Columns\IdColumn;
 use Filament\Resources\Pages\ListRecords;
 use Filament\Tables\Actions\DeleteAction;
+use Illuminate\Database\Eloquent\Builder;
 use AdvisingApp\Campaign\Filament\Resources\CampaignResource;
 
 class ListCampaigns extends ListRecords
@@ -76,8 +78,17 @@ class ListCampaigns extends ListRecords
                     ->hidden(fn (Campaign $record) => $record->hasBeenExecuted() === true),
                 DeleteAction::make()
                     ->hidden(fn (Campaign $record) => $record->hasBeenExecuted() === true),
-            ])
-            ->bulkActions([
+            ])->filters([
+                Filter::make('My Campaigns')
+                    ->query(fn (Builder $query) => $query->where('user_id', auth()->user()->getKey())),
+                Filter::make('Enabled')
+                    ->query(fn (Builder $query) => $query->where('enabled', true)),
+                Filter::make('Completed')
+                    ->query(function (Builder $query) {
+                        $query->whereHas('actions', function (Builder $query) {
+                            $query->whereNotNull('successfully_executed_at');
+                        });
+                    }),
             ]);
     }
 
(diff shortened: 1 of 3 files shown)


app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
```diff
@@ -76,8 +78,17 @@ public function table(Table $table): Table
...
+            ])->filters([
+                Filter::make('My Campaigns')
+                    ->query(fn (Builder $query) => $query->where('user_id', auth()->user()->getKey())),
+                Filter::make('Enabled')
+                    ->query(fn (Builder $query) => $query->where('enabled', true)),
+                Filter::make('Completed')
+                    ->query(function (Builder $query) {
+                        $query->whereHas('actions', function (Builder $query) {
+                            $query->whereNotNull('successfully_executed_at');
+                        });
```
Orrison, line 90: I think this needs to be adjusted. I think it needs to be where doesn't have actions where `successfully_executed_at` is null.

I could be wrong but I think they way it is now it would pass if ANY of the actions are completed, not all.

app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
```diff
@@ -76,8 +78,17 @@ public function table(Table $table): Table
...
                 DeleteAction::make()
                     ->hidden(fn (Campaign $record) => $record->hasBeenExecuted() === true),
-            ])
-            ->bulkActions([
+            ])->filters([
+                Filter::make('My Campaigns')
+                    ->query(fn (Builder $query) => $query->where('user_id', auth()->user()->getKey())),
+                Filter::make('Enabled')
+                    ->query(fn (Builder $query) => $query->where('enabled', true)),
+                Filter::make('Completed')
```
Orrison, line 86: Let's adjust or create a `ListCampaignsTest` to test these filters and whether or not they work and display the records we would expect.

Orrison (changes requested):

app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
kandarp-canyon: You mean like this, 
`Filter::make('Completed')
                    ->query(function (Builder $query) {
                        $query->whereDoesntHave('actions', function (Builder $query) {
                            $query->whereNull('successfully_executed_at');
                        });
                    }),`
it is not working as expected

Orrison: Hmm, if that does not work then you will need to explore other queries. Or if you believe your original query works then we can stick with it.

We should prove it out with a test for this filter though, so go ahead and create one. That will better help you craft the query as you can use the test case to confirm whether or not it is right.

The test should test that the filter causes records that you expect to be shown are shown and vice versa.

kandarp-canyon: Okay. let me create a test and will check that.

app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
kandarp-canyon: @Orrison I've tried to create test case for filters, please let me know if I missed anything it's not correct.

app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
Orrison: That recommendation did end up working.

app-modules/campaign/src/Filament/Resources/CampaignResource/Pages/ListCampaigns.php
Orrison: @kandarp-canyon unfortunately the tests and some of the logic was not correct. For sake of deployment though I have fixed them.

Orrison (dismissed):
@kandarp-canyon unfortunately the tests were not quite right and some of the logic needed to be fixed. For sake of making it into deployment I have finished this off though.

Please review at some point sooner to get a better idea of what we would be looking for here.

Orrison (approved):

PR #1096, opened 2024-11-04
[ADVAPP-941]: Update knowledge management to serve as the resource hub for a customer

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-941

### Technical Description

> Update knowledge management to serve as the resource hub for a customer

> No

> Yes. `ResourceHub`


payal-canyon:
Hi @Orrison 

The classes' renaming from knowledge base/management to resource hub has been done, and as per the discussion, I created the PR for review. Meanwhile, I started working on the next part, which is the renaming of tables and columns.

Thank You!

app-modules/resource-hub/database/migrations/2024_11_05_121450_data_migrate_knowledge_base_related_data_to_resource_hub.php
Orrison, line 1: Similiar to the other comments, you cannot drop the views until after the Feature Flag is toggled. As doing so will result in downtime

app-modules/resource-hub/database/migrations/2024_10_30_055640_data_migration_knowledge_base_to_resource_hub_permissions.php
Orrison, line 1: This will need to be wrapped in a transaction with a Feature Flag being toggled in the end. Or we will end up with a state in the application where checked permissions will fail.

Orrison (changes requested):

PR #1109, opened 2024-11-12
[ADVAPP-974] Eliminate and cleanup unused modules for assets and changes

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-974

### Technical Description

> Eliminate and cleanup unused modules for assets and changes

> No.

> No.


database/migrations/2024_11_12_130822_data_remove_assets_module.php
danharrin, line 1: Please keep the old migration instead of deleting it

danharrin (changes requested):

database/migrations/2024_11_12_130822_data_remove_assets_module.php
Orrison, line 1: I think the schema migrations, removal of tables should be in a separate migration.

Permission changes should be in a separate migration with the prefix `seed_permissions_`

I also think I am find with deleting the old migrations creating the tables in this scenario. What do you think @danharrin?

Finally there should be a data migration to remove any of the timeline records related to these models as they would error out now.

Orrison (changes requested):

database/migrations/2024_11_12_130822_data_remove_assets_module.php
danharrin: Personally I don't think we should delete the create tables, but we could move them into a `/legacy` directory in `/migrations` that still gets run but are not relevant to the app anymore

I just feel like deleting tables (and then having to potentially go through other tables and remove foreign keys) is not standard practice

database/migrations/2024_11_12_130822_drop_assets_and_related_tables.php
```diff
@@ -34,19 +34,21 @@
...
-            'name' => fake()->word(),
-        ];
+        Schema::dropIfExists('asset_check_outs');
+        Schema::dropIfExists('asset_check_ins');
+        Schema::dropIfExists('maintenance_activities');
+        Schema::dropIfExists('maintenance_providers');
+        Schema::dropIfExists('assets');
+        Schema::dropIfExists('asset_locations');
+        Schema::dropIfExists('asset_statuses');
+        Schema::dropIfExists('asset_types');
```
Orrison, line 50: This really isn't how this migration should work. This should be changed to the more standard way of dropping tables. A migration per table dropped with a `down` that would restore the table to it's last state.

Orrison (changes requested):

database/migrations/2024_11_12_130822_drop_assets_and_related_tables.php
ankit-canyon: Hi @Orrison 

Yep, I recreated migration files for drop table and remove permissions.

Thank you.

Orrison (approved):

Orrison:
@danharrin, please resolve your change request for this one when you can.

PR #1117, opened 2024-11-18
[ADVAPP-992]: Track who creates campaigns and show this information in the campaign view mode

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-992

### Technical Description

> Track who creates campaigns and show this information in the campaign view mode

> No

> AddCreatedByFeature


app-modules/campaign/database/migrations/2024_11_15_054949_data_fill_created_by.php
```diff
@@ -0,0 +1,49 @@
...
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Campaign::with('user')->each(function ($campaign) {
+            $campaign->created_by_id = $campaign->user?->getKey();
+            $campaign->created_by_type = 'user';
+            $campaign->save();
+        });
```
Orrison, line 47: Please review the documentation on data migrations: https://github.com/canyongbs/advisingapp/blob/main/docs/data-migrations.md

Data migrations cannot contain references to classes like `Campaign`. To make changes in the database, you must use the DB facade.

Let's also make sure we have a fall back here just in case the campaign row does not have a `user_id` set and get the first record non-deleted record from the `users` table.

app-modules/campaign/database/migrations/2024_11_15_055809_change_created_by_id_to_non_nullable_in_campaigns_table.php
Orrison, line 1: Because you are affecting more than just the id column a better name would probably be `2024_11_15_055809_change_created_by_to_non_nullable_in_campaigns_table.php`

app-modules/campaign/database/migrations/2024_11_15_065127_data_activate_feature_flag_created_by_campaign.php
Orrison, line 1: After discussing this with Product and realizing that this `user` relationship and `user_id` column exist, let's get rid of it, as this new `created_by` relationship takes its place.

So before this data migration toggling the Feature Flag we should have a migration dropping the `user_id` column.

We will then also need to get rid of the relationship in the model and replace where it is used in Filters and anywhere else with this new createdBy relationship.

app-modules/campaign/src/Observers/CampaignObserver.php
```diff
@@ -44,6 +44,8 @@ public function creating(Campaign $campaign): void
     {
         if (is_null($campaign->user_id) && ! is_null(auth()->user())) {
             $campaign->user_id = auth()->user()->id;
+            $campaign->created_by_type = 'user';
+            $campaign->created_by_id = auth()->user()->id;
```
Orrison, line 48:
```suggestion
            $campaign->created_by_id = auth()->user()->getKey();
```

Orrison (changes requested):

app-modules/campaign/database/migrations/2024_11_15_054949_data_fill_created_by.php
amit-canyon: @Orrison 
What do you mean by "get the first record non-deleted record from the users table."? Can you please explain little bit here.

Orrison: I meant to get the actual first User in the database that was not soft deleted to use in case the `user_id` column was `null`

Orrison (approved):

PR #1121, opened 2024-11-19
[ADVAPP-1006]: Disassociate results in a page not found error

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1006

### Technical Description

> Disassociate results in a page not found error

> No

> No

### DIFF
--- a/app-modules/prospect/src/Filament/Resources/ProspectResource/Actions/ConvertToStudent.php
+++ b/app-modules/prospect/src/Filament/Resources/ProspectResource/Actions/ConvertToStudent.php
@@ -36,6 +36,7 @@
 
 namespace AdvisingApp\Prospect\Filament\Resources\ProspectResource\Actions;
 
+use Filament\Pages\Page;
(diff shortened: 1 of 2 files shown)


app-modules/prospect/src/Filament/Resources/ProspectResource/Actions/ConvertToStudent.php
```diff
@@ -94,6 +95,8 @@ protected function setUp(): void
                     ->title('Prospect converted to Student')
                     ->success()
                     ->send();
+
+                $this->redirect(ProspectResource::getUrl('view', ['record' => $this->record]));
```
Orrison, line 99: You should check whether or not we are actually on the view page or not. Because if you are, then you don't need to redirect.

Orrison (changes requested):

Orrison (approved):

PR #1127, opened 2024-11-21
[ADVAPP-1001]: Modify the iconography to be more consistent in the collapsed view of the navigation

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1001

### Technical Description

> Modify the iconography to be more consistent in the collapsed view of the navigation.

> No.

> No.

### DIFF
--- a/app-modules/prospect/src/Filament/Pages/RecruitmentCrmDashboard.php
+++ b/app-modules/prospect/src/Filament/Pages/RecruitmentCrmDashboard.php
@@ -56,6 +56,8 @@ class RecruitmentCrmDashboard extends BaseDashboard
 
     protected static string $routePath = 'recruitment-crm-dashboard';
 
+    protected static ?string $navigationIcon = '';
+
     public static function shouldRegisterNavigation(): bool
     {
         /** @var User $user */
--- a/app-modules/student-data-model/src/Filament/Pages/RetentionCrmDashboard.php
+++ b/app-modules/student-data-model/src/Filament/Pages/RetentionCrmDashboard.php
@@ -55,6 +55,8 @@ class RetentionCrmDashboard extends BaseDashboard
 
     protected static string $routePath = 'retention-crm-dashboard';
 
+    protected static ?string $navigationIcon = '';
+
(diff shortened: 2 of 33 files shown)


Orrison (changes requested):
There are currently merge conflicts.

app-modules/prospect/src/Filament/Pages/RecruitmentCrmDashboard.php
```diff
@@ -56,6 +56,8 @@ class RecruitmentCrmDashboard extends BaseDashboard
 
     protected static string $routePath = 'recruitment-crm-dashboard';
 
+    protected static ?string $navigationIcon = '';
```
Orrison, line 59: Does `null` work here? Or does it have to be an empty string to not default to something?

app-modules/student-data-model/src/Filament/Pages/RetentionCrmDashboard.php
```diff
@@ -55,6 +55,8 @@ class RetentionCrmDashboard extends BaseDashboard
 
     protected static string $routePath = 'retention-crm-dashboard';
 
+    protected static ?string $navigationIcon = '';
+
```
Orrison, line 59: Same question as the other dashboard.

Orrison (changes requested):
Some merge conflicts again.

app-modules/prospect/src/Filament/Pages/RecruitmentCrmDashboard.php
kandarp-canyon: I've tried with null as I mentioned in the call but it doesn't work, so I went with the empty string.

app-modules/student-data-model/src/Filament/Pages/RetentionCrmDashboard.php
kandarp-canyon: @Orrison null doesn't work for this.

Orrison (approved):

PR #1149, opened 2024-11-28
[ADVAPP-1032]: Simulate an inbound text message response when demo mode is enabled.

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1032

### Technical Description

If the setting is enabled, a response is created from the SMS driver using the same action that is used by the Twilio web hook controller.

No

`TwilioDemoAutoReplyModeFeature`

### DIFF
--- a/app-modules/integration-twilio/database/migrations/2024_11_28_151611_add_is_demo_auto_reply_mode_enabled_setting_to_twilio_settings.php
+++ b/app-modules/integration-twilio/database/migrations/2024_11_28_151611_add_is_demo_auto_reply_mode_enabled_setting_to_twilio_settings.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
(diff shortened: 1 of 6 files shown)


app-modules/integration-twilio/database/migrations/2024_11_28_151611_add_is_demo_auto_reply_mode_enabled_setting_to_twilio_settings.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Spatie\LaravelSettings\Migrations\SettingsBlueprint;
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        $this->migrator->inGroup('twilio', function (SettingsBlueprint $blueprint): void {
+            $blueprint->add('is_demo_auto_reply_mode_enabled', false);
+        });
```
Orrison, line 45: You will want to catch `SettingAlreadyExists` and ignore it if so. The way the settings works is if it is accessed before this migration runs it will create itself from the default in the class.

Orrison (changes requested):

app-modules/integration-twilio/database/migrations/2024_11_28_151611_add_is_demo_auto_reply_mode_enabled_setting_to_twilio_settings.php
danharrin: Good point, done

Orrison (approved):

PR #1151, opened 2024-11-29
[ADVAPP-1038]: Purge experimental reporting from the system

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1038

### Technical Description

Removed feature, adjusted settings table to remove references, and registered legacy migrations directory so they are still run.

No

No, data migration needs to remain since the settings are created in previous migrations


app/Providers/AppServiceProvider.php
```diff
@@ -93,6 +93,8 @@ public function register(): void
                 return $app->make(CacheFactory::class);
             });
         });
+
+        $this->loadMigrationsFrom(database_path('migrations/Legacy'));
```
Orrison, line 97: Were Legacy migrations not running before?

database/migrations/Legacy/2024_11_29_133232_data_drop_report_assistant_settings.php
```diff
@@ -30,6 +32,29 @@
...
---}}
+*/
+
+use Database\Migrations\Concerns\CanModifySettings;
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    use CanModifySettings;
+
+    public function up(): void
```
Orrison, line 43: Should this have a down to reverse?

Orrison (changes requested):

app/Providers/AppServiceProvider.php
danharrin: As far as I could tell, no

database/migrations/Legacy/2024_11_29_133232_data_drop_report_assistant_settings.php
danharrin: Done

Orrison (approved):

PR #1157, opened 2024-12-05
[ADVAPP-1041]: Prevent the removal of a population segment that is associated with an active campaign

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1041

### Technical Description

> Prevent the removal of a population segment that is associated with an active campaign.

> No.

> No.

### DIFF
--- a/app-modules/segment/src/Filament/Resources/SegmentResource/Pages/ListSegments.php
+++ b/app-modules/segment/src/Filament/Resources/SegmentResource/Pages/ListSegments.php
@@ -39,17 +39,16 @@ namespace AdvisingApp\Segment\Filament\Resources\SegmentResource\Pages;
 use Filament\Tables\Table;
 use Filament\Actions\CreateAction;
 use Filament\Tables\Filters\Filter;
+use AdvisingApp\Segment\Models\Segment;
 use Filament\Tables\Actions\EditAction;
 use Filament\Tables\Columns\TextColumn;
-use Filament\Notifications\Notification;
 use AdvisingApp\Prospect\Models\Prospect;
 use App\Filament\Tables\Columns\IdColumn;
 use Filament\Resources\Pages\ListRecords;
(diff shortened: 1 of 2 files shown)


app-modules/segment/src/Filament/Resources/SegmentResource/Pages/ListSegments.php
```diff
@@ -76,7 +77,24 @@ public function table(Table $table): Table
             ])
             ->actions([
                 EditAction::make(),
-                DeleteAction::make(),
+                DeleteAction::make()
+                    ->action(function (DeleteAction $action) {
+                        if ($action->getRecord()->campaigns()->exists()) {
```
Orrison, line 82: Let's do this check in the Policy of the Segment, returning a Deny response with the message.

We can then authorize the User here and return the notification with the message and cancel the action if it is denied.

Orrison (changes requested):

app-modules/segment/src/Filament/Resources/SegmentResource/Pages/ListSegments.php
ankit-canyon: > Let's do this check in the Policy of the Segment, returning a Deny response with the message.
> 
> We can then authorize the User here and return the notification with the message and cancel the action if it is denied.

For notification message , DeleteAction will not visible once we provide the condition in Policy.

Orrison: You still need to do the `action` override code you had but capture the denial and usage the message from it in the notification.

Orrison (changes requested):
I believe you would still do what you are doing before with overriding the action but check the auth and then return the message if it is denied.

Orrison (approved):

PR #1169, opened 2024-12-09
[ADVAPP-1055]: Adjust the scope of table pagination options for all table views

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1055

### Technical Description

> Adjust the scope of table pagination options for all table views

> No

> No

### DIFF
--- a/app/Providers/AppServiceProvider.php
+++ b/app/Providers/AppServiceProvider.php
@@ -155,12 +155,10 @@ class AppServiceProvider extends ServiceProvider
 
         Feature::discover();
 
-        if (! app()->runningUnitTests()) {
-            Table::configureUsing(function (Table $table): void {
-                $table
-                    ->paginationPageOptions([5, 10, 20, 'all'])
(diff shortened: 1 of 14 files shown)


app/Providers/AppServiceProvider.php
```diff
@@ -153,5 +154,13 @@ public function boot(): void
...
 
         Feature::discover();
+
+        if (! app()->runningUnitTests()) {
+            Table::configureUsing(function (Table $table): void {
+                $table
+                    ->paginationPageOptions([5, 10, 20, 'all'])
+                    ->defaultPaginationPageOption(5);
+            });
+        }
```
Orrison, line 164: I don't think we should only run this during non-testing.

The test state should be as close as possible to the Production state to which this deviates.

Let's remove the `app()->runningUnitTests()` here and modify the tests that need modifying to select a different pagination option if needed.

Orrison (changes requested):

PR #1174, opened 2024-12-10
[ADVAPP-1050]: Enhance permissions with a matrix role editor

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1050

### Technical Description

Introduce new `PermissionMatrix` field to manage role permissions across Create, View and Edit pages.

No

No


Orrison (commented):
This all looks and works great! Even on mobile, which is awesome.

I looked through the list of remaining permissions you sent me and will create a separate ticket to address most of them. But some of them are for `import` which we do have setup in some of our policies and I think we should keep.

Would it be easy to add `import` after `restore`?

Asked Product in the ticket if they want this or not.

Orrison (changes requested):
Per discussion with Product in the ticket, let's add `import` into the matrix as well after `delete` and before `force-delete`

danharrin:
@Orrison added Import

Orrison (approved):

PR #1176, opened 2024-12-11
[ADVAPP-1053]: Display auditable model name in list view and detail view in title caps

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1053

### Technical Description

> Display auditable model name in list view and detail view in title caps

> No

> No


Orrison (changes requested):
I was wrong in the tech planning notes that `title` is all that is needed. There should be a string helper method that cap cases and changes underscores and/or dashes to spaces. Which we need to do here to avoid instances like in this screenshot.

Screenshot 2024-12-11 at 11 42 02 AM

Orrison (approved):

Orrison (changes requested):
Looks like there are currently some merge conflicts

Orrison (approved):

PR #1183, opened 2024-12-13
[ADVAPP-1069]: Introduce a new report in the report library for user activity

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1069

### Technical Description

> Introduce a new report in the report library for user activity

No

> Feature Flag: UserTrackedEventsFeature


jayushi-canyon:
@Orrison The test in this PR is failing because of new unique key defined for `TrackedEventCount` table.
`RecordTrackedEvent` is not able to update the record whose `related_to` columns are null and creating new record for that same type,

app/Exceptions/Handler.php
```diff
@@ -83,7 +81,7 @@ public function register(): void
     protected function unauthenticated($request, AuthenticationException $exception)
     {
         return $this->shouldReturnJson($request, $exception)
-            ? response()->json(['message' => $exception->getMessage()], 401)
-            : redirect()->guest($exception->redirectTo() ?? url('/'));
+          ? response()->json(['message' => $exception->getMessage()], 401)
+          : redirect()->guest($exception->redirectTo() ?? url('/'));
```
Orrison, line 85: Though I am fine with the changes to `$levels` above, this reduces tabbing down to 2 spaces for these lines. This is against our formatting rules and I am not sure why CI is not fixing this.

Regardless I have seen this in a few of your PRs. Please check your IDE and change it so tabbing is 4 spaces, to 2. It may also be good to setup your IDE to use our linting rules. There is a PHPCSFixer plugin recommended in the Loop document for VSCode.

app-modules/report/src/Filament/Widgets/UserUniqueLoginCountLineChart.php
```diff
@@ -0,0 +1,100 @@
...
+        ];
+    }
+
+    protected function getData(): array
+    {
+        $runningTotalPerMonth = Cache::tags([$this->cacheTag])->remember('user-unique-login-count-line-chart', now()->addHours(24), function (): array {
+            $totalCreatedPerMonth = User::query()
+                ->selectRaw('date_trunc(\'month\', created_at) as month')
+                ->selectRaw('count(*) as total')
+                ->whereRelation('logins', 'type', TrackedEventType::UserLogin)
```
Orrison, line 76: You will not need to do this scoping of the type here once the relationship on the `User` Model is fixed.

app-modules/report/src/Filament/Widgets/UsersLoginCountTable.php
```diff
@@ -0,0 +1,158 @@
...
+            ->query(
+                function () {
+                    return User::with('loginsCount');
+                }
+            )
+            ->columns([
+                TextColumn::make('name')
+                    ->searchable(),
+                TextColumn::make('email'),
+                TextColumn::make('sisid')
```
Orrison, line 89: The id of the column does not make sense. It should probably be something like `has_logged_in`

app-modules/report/src/Filament/Widgets/UsersLoginCountTable.php
```diff
@@ -0,0 +1,158 @@
...
+                    ->date('m-d-Y'),
+                TextColumn::make('last_logged_in_at')
+                    ->label('Last Login')
+                    ->date('m-d-Y'),
+                TextColumn::make('loginsCount.count')
+                    ->label('Total Logins')
+                    ->default(0),
+            ])
+            ->filters([
+                SelectFilter::make('User')
```
Orrison, line 110: The id of this select filter does not make sense. It should be something like `has_logged_in`

app-modules/report/src/Filament/Widgets/UserUniqueLoginCountLineChart.php
Orrison, line 1: Let's add some Unit tests to ensure this calculates the correct values.

app-modules/report/src/Filament/Widgets/UsersLoginCountTable.php
Orrison, line 1: Let's add some unit tests ensuring the table and columns display the correct data and that the filters work properly.

app-modules/report/src/Filament/Widgets/UsersLoginCountTable.php
```diff
@@ -0,0 +1,158 @@
...
+                                fn (Builder $query, $date): Builder => $query->whereDate('last_logged_in_at', '>=', $date),
+                            )
+                            ->when(
+                                $data['last_logged_in_until'],
+                                fn (Builder $query, $date): Builder => $query->whereDate('last_logged_in_at', '<=', $date),
+                            );
+                    }),
+            ])
+            ->filtersFormWidth(MaxWidth::Small)
+            ->paginated([10]);
```
Orrison, line 156: You can remove the pagniation here. We now default all to 5.

Orrison, line 156: Let's add a default sort on this table for `last_logged_in_at` `desc`

app-modules/report/src/Filament/Widgets/UsersStats.php
Orrison, line 1: Let's add a Unit test for this ensuring it gives the proper data back.

app-modules/report/src/Filament/Widgets/UsersStats.php
```diff
@@ -0,0 +1,77 @@
...
+            )),
+            Stat::make('New Users', Number::abbreviate(
+                Cache::tags([$this->cacheTag])->remember('new-users-count', now()->addHours(24), function (): int {
+                    return User::where('created_at', '>=', Carbon::now()->subDays(30))->count();
+                }),
+                maxPrecision: 2,
+            )),
+            Stat::make('Unique Logins', Number::abbreviate(
+                Cache::tags([$this->cacheTag])->remember('unique-logins-count', now()->addHours(24), function (): int {
+                    return User::whereRelation('logins', 'type', TrackedEventType::UserLogin)->count();
```
Orrison, line 71: You won't need to do the type scoping here once the relationship is fixed in the User Model.

app-modules/report/src/Jobs/RecordUserUniqueLoginTrackedEvent.php
```diff
@@ -0,0 +1,116 @@
...
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    /**
+     * Create a new job instance.
+     */
+    public function __construct(
+        public TrackedEventType $type,
```
Orrison, line 62: Because this job is specific to tracking the User Login event. I think you are safe to not pass this in and hardcode it where it is used in this class.

app-modules/report/src/Jobs/RecordUserUniqueLoginTrackedEvent.php
```diff
@@ -0,0 +1,116 @@
...
+                    [
+                        'count' => DB::raw('tracked_event_counts.count + 1'),
+                        'last_occurred_at' => DB::raw("GREATEST(tracked_event_counts.last_occurred_at, '{$this->occurredAt}')"),
+                        'updated_at' => now(),
+                    ]
+                );
+
+            $this->user->update([
+                'first_login_at' => $this->user->first_login_at ?? now(),
+                'last_logged_in_at' => now(),
```
Orrison, line 106: This should use `$occurredAt` not `now()`. The time of the Event and when this job gets processed will be different.

app-modules/report/src/Listeners/CheckUserUniqueLoginTrackedEvent.php
```diff
@@ -0,0 +1,61 @@
...
+*/
+
+namespace AdvisingApp\Report\Listeners;
+
+use AdvisingApp\Report\Enums\TrackedEventType;
+use AdvisingApp\Report\Jobs\RecordUserUniqueLoginTrackedEvent;
+use App\Models\User;
+use Illuminate\Auth\Events\Login;
+
+class CheckUserUniqueLoginTrackedEvent
```
Orrison, line 44: "Check" doesn't really make sense here. Perhaps we can name this `ProcessUserUniqueLoginTrackedEvent`

Orrison (changes requested):
You are making changes to the database and referencing those changes in your code when setting data to these new or changed columns. Please keep in mind that all changes like this need a Feature Flag to be created and applied in areas where the code would break if ran before your migrations completed.

app-modules/report/src/Filament/Pages/Users.php
```diff
@@ -0,0 +1,82 @@
...
+
+    protected static ?string $navigationGroup = 'Users';
+
+    protected static ?string $navigationLabel = 'Overview';
+
+    protected static ?string $title = 'Users (Overview)';
+
+    protected static string $routePath = 'Users';
+
+    protected static ?int $navigationSort = 50;
```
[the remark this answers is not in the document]
jayushi-canyon: Yes you are right, I got it confused with User cluster navigation sort.

app/Exceptions/Handler.php
jayushi-canyon: I have setup PHPCSFixer plugin in my IDE still even i am not sure why this is happening

app-modules/report/src/Filament/Widgets/StatsOverviewReportWidget.php
```diff
@@ -43,7 +43,7 @@
 abstract class StatsOverviewReportWidget extends StatsOverviewWidget
 {
     #[Locked]
-    public string $cacheTag;
+    public string $cacheTag = 'default';
```
Orrison, line 46: This should not be given a default value. It needs to be set by a class that extends this.

app-modules/report/src/Filament/Widgets/UserUniqueLoginCountLineChart.php
```diff
@@ -0,0 +1,99 @@
...
+    protected function getData(): array
+    {
+        $runningTotalPerMonth = Cache::tags([$this->cacheTag])->remember('user-unique-login-count-line-chart', now()->addHours(24), function (): array {
+            $totalCreatedPerMonth = User::query()
+                ->selectRaw('date_trunc(\'month\', created_at) as month')
+                ->selectRaw('count(*) as total')
+                ->with('logins')
+                ->groupBy('month')
+                ->orderBy('month')
+                ->pluck('total', 'month');
```
Orrison, line 78: Are you sure this is getting the correct data? This kind of looks like it is getting the amount of Users created per month.. Not login events like it is supposed to.

I think you should be doing this based on the tracked event of the user login type.

app-modules/report/src/Filament/Widgets/UsersStats.php
```diff
@@ -0,0 +1,76 @@
...
+            )),
+            Stat::make('New Users', Number::abbreviate(
+                Cache::tags([$this->cacheTag])->remember('new-users-count', now()->addHours(24), function (): int {
+                    return User::where('created_at', '>=', Carbon::now()->subDays(30))->count();
+                }),
+                maxPrecision: 2,
+            )),
+            Stat::make('Unique Logins', Number::abbreviate(
+                Cache::tags([$this->cacheTag])->remember('unique-logins-count', now()->addHours(24), function (): int {
+                    return User::withCount(['logins'])->count();
```
Orrison, line 70: Are you sure this is correct? Are you sure this doesn't just return the total count of Users?

What is expected here is the total logins ever. So actually even ditching what the ticket says, this should probably be the value from the TrackedEventsCount of the user login type.

app-modules/report/src/Jobs/RecordUserUniqueLoginTrackedEvent.php
```diff
@@ -0,0 +1,115 @@
...
+                            'last_occurred_at' => $this->occurredAt,
+                            'updated_at' => now(),
+                            'created_at' => now(),
+                            'related_to_id' => $this->user->id,
+                            'related_to_type' => $this->user->getMorphClass(),
+                        ],
+                    ],
+                    ['related_to_type', 'related_to_id', 'type'],
+                    [
+                        'count' => DB::raw('tracked_event_counts.count + 1'),
```
Orrison, line 97: Can you write a test for me to ensure that this isn't effected by tracked_event_counts of other types?

app-modules/report/src/Jobs/RecordUserUniqueLoginTrackedEvent.php
```diff
@@ -0,0 +1,115 @@
...
+                        'last_occurred_at' => DB::raw("GREATEST(tracked_event_counts.last_occurred_at, '{$this->occurredAt}')"),
+                        'updated_at' => now(),
+                    ]
+                );
+
+            $this->user->update([
+                'first_login_at' => $this->user->first_login_at ?? now(),
```
Orrison, line 104: `now()` here should also be `$this->occurredAt`

app-modules/report/tests/Unit/UniqueLogin/UserStatsTest.php
```diff
@@ -0,0 +1,80 @@
...
+it('Check new users who were created within one month', function () {
+    $userCount = rand(1, 10);
+    $usersStats = new UsersStats();
+
+    User::factory()->count(3)->create([
+        'created_at' => now()->subMonths(2),
+    ]);
+    User::factory()->count($userCount)->create([
+        'created_at' => now()->subDays(rand(1, 30)),
+    ]);
```
Orrison, line 59: Rather than doing a random number here I would make it very specific and say 3 as well.

app-modules/report/tests/Unit/UniqueLogin/UserStatsTest.php
```diff
@@ -0,0 +1,80 @@
...
+
+it('Check total users with unique login event type', function () {
+    $userCount = rand(1, 10);
+    $usersStats = new UsersStats();
+
+    User::factory()->count($userCount)->create()->each(function ($user) {
+        dispatch(new RecordUserUniqueLoginTrackedEvent(
+            occurredAt: now(),
+            user: $user,
+        ));
```
Orrison, line 74: You should dispatch a known random amount of login events per User. Then track what the login state gets back.

I think the way this test is written is currently hiding the fact that you are in fact returning a count of Users, not the amount of login events.

Orrison (changes requested):

app-modules/report/src/Jobs/RecordUserUniqueLoginTrackedEvent.php
jayushi-canyon: @Orrison As discussed on call we are not supposed to create test case for this.

database/migrations/2024_12_16_165342_add_first_and_last_logged_at_in_users_table.php
danharrin, line 1: How about `add_first_login_at_and_last_logged_in_at_columns_to_users_table`

app-modules/report/src/Filament/Widgets/UsersLoginCountTable.php
```diff
@@ -0,0 +1,158 @@
...
+use Filament\Tables\Filters\Filter;
+use Filament\Tables\Filters\SelectFilter;
+use Filament\Tables\Table;
+use Filament\Widgets\TableWidget as BaseWidget;
+use Illuminate\Database\Eloquent\Builder;
+use Livewire\Attributes\On;
+
+class UsersLoginCountTable extends BaseWidget
+{
+    public string $cacheTag;
```
danharrin, line 52: All public `$cacheTag` properties need to be `#[Locked]` to prevent security issues, please ensure this is reflected across the module

danharrin (changes requested):

danharrin (changes requested):
`2024_12_13_123120_add_related_to_polymorphic_columns_in_tracked_event_counts_table` cannot be rolled back since the index does not exist/has the wrong name, please check that locally to inspect the error

app-modules/report/src/Filament/Pages/UserLoginActivity.php
```diff
@@ -0,0 +1,77 @@
...
+
+    protected static ?string $navigationLabel = 'Overview';
+
+    protected static ?string $title = 'Login Activity';
+
+    protected static string $routePath = 'Users';
+
+    protected static ?int $navigationSort = 10;
+
+    protected $cacheTag = 'report-users';
```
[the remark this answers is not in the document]
danharrin: Protected properties do not need to be locked since they are not tracked by Livewire anyway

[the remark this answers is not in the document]
Orrison: Ah yes good point, I wasn't paying attention here @jayushi-canyon Locked should be removed on this protected property,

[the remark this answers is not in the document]
Orrison: @jayushi-canyon please do still make the tag name change

Orrison (changes requested):

danharrin (approved):

app-modules/report/src/Filament/Pages/UserLoginActivity.php
```diff
@@ -0,0 +1,77 @@
...
+use AdvisingApp\Report\Filament\Widgets\RefreshWidget;
+use AdvisingApp\Report\Filament\Widgets\UsersLoginCountTable;
+use AdvisingApp\Report\Filament\Widgets\UsersStats;
+use AdvisingApp\Report\Filament\Widgets\UserUniqueLoginCountLineChart;
+
+class UserLoginActivity extends UserReport
+{
+    protected static ?string $navigationIcon = 'heroicon-o-document-text';
+
+    protected static ?string $navigationLabel = 'Overview';
```
Orrison, line 49:
```suggestion
    protected static ?string $navigationLabel = 'Login Activity';
```

Orrison (changes requested):

Orrison (approved):

PR #1192, opened 2024-12-17
[ADVAPP-1066]: Introduce the concept of smart prompts

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1066

### Technical Description

Smart prompts are stored in the existing prompts table, with `is_smart` set to true. When inserted, the ID of the prompt gets sent to the frontend to initiate the response, and the prompt itself is not leaked.

No

SmartPromptsFeature


app-modules/ai/src/Jobs/CloneAiThread.php
```diff
@@ -78,7 +79,7 @@ public function handle(): void
             $threadReplica->user()->associate($this->recipient);
             $threadReplica->save();
 
-            foreach ($this->thread->messages as $message) {
+            foreach ($this->thread->messages()->when(SmartPromptsFeature::active(), fn ($query) => $query->whereNull('prompt_id'))->get() as $message) {
```
Orrison, line 82: I think we need to still include the smart prompt triggers when you clone a chat. Or else the history of the conversation will not be there for the AI to understand.

app-modules/ai/src/Notifications/SendAssistantTranscriptNotification.php
```diff
@@ -71,7 +72,7 @@ public function toEmail(object $notifiable): MailMessage
                 ->line("Here is a copy of {$this->sender->name}'s chat with {$this->thread->assistant->name}:");
         }
 
-        $this->thread->messages()->with('user')->get()
+        $this->thread->messages()->when(SmartPromptsFeature::active(), fn ($query) => $query->whereNull('prompt_id'))->with('user')->get()
```
Orrison, line 75: Rather than filtering out the messages with a `prompt_id` I think we should probably include them but make the displayed content the same as in the UI

Orrison (changes requested):

app-modules/ai/src/Jobs/CloneAiThread.php
danharrin: Done

app-modules/ai/src/Notifications/SendAssistantTranscriptNotification.php
danharrin: Done

Orrison (approved):

PR #1196, opened 2024-12-19
[ADVAPP-1064]: Refactor Task access to not use individual permissions to manage access

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1064

### Technical Description

> Refactor Task access to not use individual permissions to manage access

> No

Data Migrations:
-`app-modules/task/database/migrations/2024_12_18_104302_delete_task_individual_update_permissions.php`

### DIFF
--- a/app-modules/task/src/Policies/TaskPolicy.php
+++ b/app-modules/task/src/Policies/TaskPolicy.php
@@ -99,10 +99,11 @@ class TaskPolicy implements PerformsChecksBeforeAuthorization
             return Response::deny('You do not have permission to update this task.');
         }
 
-        return $authenticatable->canOrElse(
-            abilities: ["task.{$task->getKey()}.update"],
-            denyResponse: 'You do not have permission to update this task.'
-        );
+        if ($authenticatable->getKey() !== $task?->assigned_to && $authenticatable->getKey() !== $task?->created_by) {
+            return Response::deny('You do not have permission to update this task.');
+        }
+
+        return Response::allow();
     }
 
     public function delete(Authenticatable $authenticatable, Task $task): Response
--- a/app-modules/task/src/Policies/TaskPolicy.php
+++ b/app-modules/task/src/Policies/TaskPolicy.php
@@ -99,11 +99,14 @@ class TaskPolicy implements PerformsChecksBeforeAuthorization
             return Response::deny('You do not have permission to update this task.');
         }
 
-        if ($authenticatable->getKey() !== $task?->assigned_to && $authenticatable->getKey() !== $task?->created_by) {
+        if ($authenticatable->getKey() !== $task->assigned_to && $authenticatable->getKey() !== $task->created_by) {
             return Response::deny('You do not have permission to update this task.');
         }
 
-        return Response::allow();
+        return $authenticatable->canOrElse(
+            abilities: ['task.*.update'],
+            denyResponse: 'You do not have permission to update this task.'
+        );
     }
 
     public function delete(Authenticatable $authenticatable, Task $task): Response
--- a/app-modules/task/tests/EditTaskTest.php
+++ b/app-modules/task/tests/EditTaskTest.php
@@ -100,4 +100,137 @@ test('EditTask is gated with proper access control', function () {
         ->assertHasNoFormErrors();
 
     // TODO: Check for changes
-})->only();
+});
+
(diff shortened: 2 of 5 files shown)


app-modules/task/src/Policies/TaskPolicy.php
```diff
@@ -99,10 +99,11 @@ public function update(Authenticatable $authenticatable, Task $task): Response
             return Response::deny('You do not have permission to update this task.');
         }
 
-        return $authenticatable->canOrElse(
-            abilities: ["task.{$task->getKey()}.update"],
-            denyResponse: 'You do not have permission to update this task.'
-        );
```
Orrison, line 105: I think we still need what you deleted here after the `assigned_to` and `created_by` check instead of `Response::allow()` because the User should still have to have `test.*.update` which this will cover.

app-modules/task/src/Policies/TaskPolicy.php
```diff
@@ -99,10 +99,11 @@ public function update(Authenticatable $authenticatable, Task $task): Response
...
+        if ($authenticatable->getKey() !== $task?->assigned_to && $authenticatable->getKey() !== $task?->created_by) {
```
Orrison, line 102: I don't think you need the `nullsafe` on `$task` in both instances here. There is not a scenario in this method where `$task` can be null

app-modules/task/tests/EditTaskTest.php
```diff
@@ -53,7 +54,9 @@
 test('EditTask is gated with proper access control', function () {
```
Orrison, line 54: Because of the nature of the change here this test will need to be modified are changed to ensure that a User must have the `task.view-any` and `task.*.update` permission and either be the creator or currently assigned.

We need to test that all scenarios are covered and the proper response is returned.

app-modules/task/tests/TaskAssignmentTest.php
```diff
@@ -34,49 +34,42 @@
...
-
-it('gives the proper permission to the creator of a Task', function () {
-    /** @var Task $task */
-    $task = Task::factory()->create();
-
-    expect($task->createdBy->can("task.{$task->getKey()}.update"))->toBeTrue();
-});
-
-it('gives the proper permission to the assigned User of a Task on create and update', function () {
+it('gives the proper permission to the assigned User of a Task on update', function () {
```
Orrison, line 50: I think we can just remove this test as well. Task Assignment now does not create or do anything to the access of the User directly.

Access is intrinsic based on the assignment. Which can be tested in the Edit page test.

Orrison (changes requested):

app-modules/task/tests/EditTaskTest.php
```diff
@@ -80,12 +92,12 @@
...
+    livewire(EditTask::class, [
         'record' => $task->getRouteKey(),
     ])
         ->fillForm($request->toArray())
         ->call('save')
         ->assertHasNoFormErrors();
 
     // TODO: Check for changes
-});
+})->only();
```
Orrison, line 103: This `only` needs to be removed. Please be careful to ensure these are not present before requesting review.

I think all team members should do a code review of their own PRs while in Draft before opening them up for peer review. To check for things like this and to compare against everything listed in the ticket.

app-modules/task/tests/EditTaskTest.php
Orrison, line 1: The changes you did in this test look good. But let's split this test into multiple to cover all scenarios:

- A user without proper permissions and that is not associated to the Task cannot access
- A User with proper permissions and that is not associated to the Task cannot access
- A User without proper permissions that is the assigned user cannot access
- A User without proper permissions that is the created by user cannot access
- A User with proper permissions that is the assigned user can access
- a User with proper permissions that is the created by user can access.

Orrison (changes requested):

app-modules/task/tests/EditTaskTest.php
payal-canyon: Hi @Orrison,
I'm sorry about that. I'll take care of this next time.

Thank You!

Orrison: No problem!

Orrison (approved):

PR #1205, opened 2024-12-23
[ADVAPP-1080]: Add filter in the student list for First Generation using a ternary filter

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1080

### Technical Description

Add filter in the student list for First Generation using a ternary filter

No

No


app-modules/student-data-model/tests/Student/ListStudentsTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+it('can filter students by first generation', function () {
+    Student::truncate();
+
+    $user = User::factory()->licensed(Student::getLicenseType())->create();
+
+    $user->givePermissionTo('student.view-any');
+    $user->givePermissionTo('student.*.view');
+
+    actingAs($user);
```
Orrison, line 52: Rather than worry about permissions here, it could be best just to create a Super Admin user.

app-modules/student-data-model/tests/Student/ListStudentsTest.php
```diff
@@ -0,0 +1,64 @@
...
+    $students = Student::factory()->count(5)->create();
+
+    livewire(ListStudents::class)
+        ->assertCanSeeTableRecords($students)
+        ->filterTable('firstgen', true)
+        ->assertCanSeeTableRecords($students->where('firstgen', true))
+        ->assertCanNotSeeTableRecords($students->where('firstgen', false))
+        ->filterTable('firstgen', false)
+        ->assertCanSeeTableRecords($students->where('firstgen', false))
+        ->assertCanNotSeeTableRecords($students->where('firstgen', true));
```
Orrison, line 63: There is too much randomness in this test. You should create a set group of Students with first_gen and a set group without. Then test the filter using those groups.

You can end up in a scenario here with no users in either scenario so then the test assertion is worthless.

Orrison (changes requested):

app-modules/student-data-model/tests/Student/ListStudentsTest.php
ketan-canyon: Done

app-modules/student-data-model/tests/Student/ListStudentsTest.php
ketan-canyon: Done

Orrison (approved):

PR #1207, opened 2024-12-24
[ADVAPP-1076] Include "Cases" in top menu bar after interactions for students and prospects

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1076

### Technical Description

>  Include "Cases" in top menu bar after interactions for students and prospects.

> No.

> No.

### DIFF
--- a/app-modules/case-management/src/Filament/Resources/CaseResource/Pages/Concerns/HasCaseHeader.php
+++ b/app-modules/case-management/src/Filament/Resources/CaseResource/Pages/Concerns/HasCaseHeader.php
@@ -36,11 +36,9 @@
 
 namespace AdvisingApp\CaseManagement\Filament\Resources\CaseResource\Pages\Concerns;
 
-use AdvisingApp\CaseManagement\Filament\Resources\CaseResource\Pages\ViewCase;
 use AdvisingApp\Prospect\Filament\Resources\ProspectResource;
 use AdvisingApp\Prospect\Models\Prospect;
 use AdvisingApp\StudentDataModel\Filament\Resources\StudentResource;
-use AdvisingApp\StudentDataModel\Filament\Resources\StudentResource\Pages\ViewStudent;
 use AdvisingApp\StudentDataModel\Models\Student;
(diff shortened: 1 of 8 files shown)


Orrison (changes requested):
This needs the option to go back to the Student in the header of the Case View page. Similiar to other pages like Subscriptions > Manage.

Perhaps adding a URL param of the Student key that if present when navigating to the Case View page it displays the back button.

app-modules/case-management/src/Filament/Resources/CaseResource/Pages/Concerns/HasCaseHeader.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        $from = app('router')->getRoutes()->match(app('request')->create(url()->previous()))->getName();
+
+        if ($from === 'filament.admin.resources.students.view' && $this->record->respondent instanceof Student) {
+            $backButtonLabel = 'Back to student';
+            $backButtonUrl = StudentResource::getUrl('view', ['record' => $this->record->respondent->getKey()]);
+        } elseif ($from === 'filament.admin.resources.prospects.view' && $this->record->respondent instanceof Prospect) {
+            $backButtonLabel = 'Back to prospect';
+            $backButtonUrl = ProspectResource::getUrl('view', ['record' => $this->record->respondent->getKey()]);
+        }
```
Orrison, line 57: @danharrin what are your thoughts on this? Rather than looking at what the previous URL is I would probably recommend adding the ability to read from a URL param what to redirect back to and the label. So it can be controlled at the link level and could be re-used between Case pages.

danharrin: Yes I agree, I think there should be a URL query parameter that stores `referrer=studentProfile` or `referrer=prospectProfile`

Orrison (approved):

PR #1208, opened 2024-12-26
[ADVAPP-1078]: Introduce the ability to filter by Smart prompts in the Prompt Library

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1078

### Technical Description

> Introduce the ability to filter by Smart prompts in the Prompt Library

> No

> No

### DIFF
--- a/app-modules/ai/database/factories/PromptFactory.php
+++ b/app-modules/ai/database/factories/PromptFactory.php
@@ -55,7 +55,6 @@ class PromptFactory extends Factory
             'description' => fake()->optional()->sentences(asText: true),
             'prompt' => fake()->sentences(asText: true),
             'type_id' => PromptType::query()->inRandomOrder()->first() ?? PromptType::factory()->create(),
-            'is_smart' => fake()->boolean(),
         ];
     }
 }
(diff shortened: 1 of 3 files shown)


app-modules/ai/database/factories/PromptFactory.php
```diff
@@ -55,6 +55,7 @@ public function definition(): array
             'description' => fake()->optional()->sentences(asText: true),
             'prompt' => fake()->sentences(asText: true),
             'type_id' => PromptType::query()->inRandomOrder()->first() ?? PromptType::factory()->create(),
+            'is_smart' => fake()->boolean(),
```
Orrison, line 58: You are getting random failures in other tests because you have now added `is_smart` as a random value into this factory. So the tests that are failing are most likely assuming that it is the default of `false`.

So you need to go to those tests and where it make sense fix them to be false. Or change this to a state and apply it where it is needed.

Orrison (changes requested):

Orrison (approved):

PR #1215, opened 2024-12-30
[ADVAPP-1090]: Users are unable to delete roles

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1090

### Technical Description

Seeds role permissions, adds delete action to Edit Role page. Removed role group permissions.

No

No

### DIFF
--- a/app-modules/authorization/database/migrations/2024_12_30_142107_data_seed_role_permissions.php
+++ b/app-modules/authorization/database/migrations/2024_12_30_142107_data_seed_role_permissions.php
@@ -1,0 +1,40 @@
+<?php
+
+use Database\Migrations\Concerns\CanModifyPermissions;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    use CanModifyPermissions;
+
+    private array $permissions = [
+        'role.*.delete' => 'Role',
+        'role.*.force-delete' => 'Role',
+        'role.*.restore' => 'Role',
+        'role.*.update' => 'Role',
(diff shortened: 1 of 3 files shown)


app-modules/authorization/database/migrations/2024_12_30_142107_data_seed_role_permissions.php
```diff
@@ -0,0 +1,87 @@
...
+            });
+
+        collect($this->guards)
+            ->each(fn (string $guard) => $this->deletePermissions($this->permissionsToDelete, $guard));
+    }
+
+    public function down(): void
+    {
+        collect($this->guards)
+            ->each(fn (string $guard) => $this->deletePermissions(array_keys($this->permissions), $guard));
```
Orrison, line 85: I think this should reverse and create the permissions that were deleted, what do you think?

Orrison (changes requested):

app-modules/authorization/database/migrations/2024_12_30_142107_data_seed_role_permissions.php
danharrin: I did not include this since these permissions have not been used for months, so having them missing would not cause issues. Additionally, the `up()` can be run multiple times and will not fail if permissions do not exist when they are deleted.

Please let me know if you still want me to recreate those permissions and I will do

Orrison (approved):

PR #1248, opened 2025-01-16
[ADVAPP-1131]: When creating a campaign task due dates are being automatically added even if not populated when configuring the journey step

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1131

### Technical Description

> When creating a campaign task, due dates are being automatically added even if not populated when configuring the journey step

> No

> No


jayushi-canyon:
@Orrison 
Along with reported issue when testing I found 2 more minor issues which I have resolved in this branch itself
1. In Task, when assigned_to field is kept empty in step 2 then in review step (step 3) it gives error.
2. In few blocks executed_at field are not showing value because of wrong field name, which I have solved as well in this PR.

Let me know your thoughts on the same.

Orrison (approved):

PR #1258, opened 2025-01-22
[ADVAPP-1148]: Exclude key system messages from demo mode configuration

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1148

### Technical Description

> Exclude key system messages from demo mode configuration

> No

> No


Orrison (changes requested):
Just one small suggestion. We may also want to find a good way to add a test for this. But we want to test whether an underlying email was sent, not a notification I think.

But @payal-canyon you don't need to make these changes and can continue one with other work. Thank you so much for your work here. It looks great!

Per discussion with @danharrin earlier today, he is going to take the work you did here and wrap it into his PR for the mailer channel changes to prevent merge conflicts.

PR #1259, opened 2025-01-22
[ADVAPP-1099] & [ADVAPP-1148]: Refactor Notifications to remove Channel traits and allow via to be customized like in traditional Laravel applications

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1099
- https://canyongbs.atlassian.net/browse/ADVAPP-1148

### Technical Description

Refactors notifications to use the traditional Laravel mechanisms for sending. Uses custom channels to implement wrapper logic around the standard notification class and traits. Database-only notifications do not have the `Queueable` trait anymore so do not get queued.

No

No


app-modules/notification/src/Models/Contracts/NotifiableInterface.php
Orrison, line 1: Should this still be here? It seems to duplicate `app-modules/notification/src/Models/Contracts/CanBeNotified.php` but this one violates PSR-4

app-modules/notification/src/Notifications/Channels/MailChannel.php
```diff
@@ -0,0 +1,176 @@
...
+
+        $resetWindow = $licenseSettings->data->limits->getResetWindow();
+
+        $currentQuotaUsage = OutboundDeliverable::query()
+            ->where('channel', NotificationChannel::Email)
+            ->whereBetween('created_at', [$resetWindow['start'], $resetWindow['end']])
+            ->sum('quota_usage');
+
+        return ($currentQuotaUsage + $estimatedQuotaUsage) <= $licenseSettings->data->limits->emails;
+    }
```
Orrison, line 175: Where did we end up with the discussion of these being different? Did we want to split off fixing/unifying it into a different ticket?

Orrison (changes requested):
This looks wonderful, great work @danharrin!

I left a couple of comments. Also, as discussed, please roll in the changes in https://github.com/canyongbs/advisingapp/pull/1258.
For the changes done in that PR I small comments that can be incorporated. A major piece is it would be good if you can add some testing for the SystemNotification work.

app-modules/notification/src/Models/Contracts/NotifiableInterface.php
danharrin: Yeah that's weird this ended up left over, I have deleted the file

app-modules/notification/src/Notifications/Channels/MailChannel.php
danharrin: I don't think we decided, but technically they are doing things. First method is checking the admin status of all recipients (`to`, `cc`, `bcc`), second is checking the admin status of `to` and assuming `cc` and `bcc` are not admins. I think the second should probably always be used but this is your call.

danharrin:
@Orrison this is ready for review again.

I am afraid the system notification tests aren't great. Since Laravel notifications don't dispatch mail using the `Mail` facade, we can't use the `Mail::fake()` feature to test if a mail was sent in demo mode, we have to rely on the outbound deliverable properties. At least we are sure if the correct path in the code is followed in each case.

app-modules/notification/src/Notifications/Channels/MailChannel.php
Orrison: We talked about this in a meeting so we are all good here.

Orrison (approved):

PR #1266, opened 2025-01-27
[ADVAPP-1160]: Sometimes emails fail to send and also sometimes send while demo mode is turned on

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1160

### Technical Description

Fixes issues with mail sending by clearing out all currently found bindings where mail config could be stored. Also overrides the Tenant `makeCurrent` to make sure that it is always refreshed so that old config values are to persisted if changed.

No

No


danharrin (dismissed):

danharrin:
Should this be targeted at a hot fix branch?

danharrin (approved):

PR #1279, opened 2025-01-31
[ADVAPP-1163]: Fix tests and logic for Interaction confidentiality

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1163

### Technical Description

> Fix tests and logic for Interaction confidentiality

> No

> No

### DIFF
--- a/app-modules/student-data-model/tests/Student/InteractionsRelationManagerTest.php
+++ b/app-modules/student-data-model/tests/Student/InteractionsRelationManagerTest.php
@@ -35,8 +35,11 @@
 */
 
 use AdvisingApp\Authorization\Enums\LicenseType;
+use AdvisingApp\CaseManagement\Filament\Resources\CaseResource\RelationManagers\InteractionsRelationManager;
 use AdvisingApp\Interaction\Filament\Resources\InteractionResource\Pages\ListInteractions;
 use AdvisingApp\Interaction\Models\Interaction;
+use AdvisingApp\StudentDataModel\Filament\Resources\StudentResource\Pages\ViewStudent;
+use AdvisingApp\StudentDataModel\Models\Student;
 use AdvisingApp\Team\Models\Team;
 use App\Models\User;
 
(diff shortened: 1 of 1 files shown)


Orrison (changes requested):
@amit-canyon I apologize for the run-around. But after tech lead discussion we have decided to change the testing strategy for this.

Please review the updated tech planning notes in the ticket and modify to meet that scoping.
Let me know if you have any questions.

Orrison (changes requested):

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
 use AdvisingApp\Team\Models\Team;
 use App\Models\User;
 
 use function Pest\Laravel\actingAs;
-use function Pest\Livewire\livewire;
 use function Tests\asSuperAdmin;
 
-test('ListInteration with display data using the is_confidential field', function () {
+test('Interaction model has applied global scope', function () {
+    Interaction::factory()->create();
```
Orrison, line 47: You do not need to create this record with this factory here. ALl you needed was the line below.

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
+
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => true,
+    ]);
 
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => false,
+    ]);
+
+    expect(Interaction::query()->count())->toBe(20);
```
Orrison, line 70: Would shouldn't just check count. We should also be checking the actual records that are returned to see if they are the records we expect.

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => true,
+    ]);
+
+    Interaction::factory()->count(10)->create([
         'is_confidential' => false,
     ]);
 
-    $allInteractions = $confidentialInteraction->merge($nonConfidentialInteraction);
+    expect(Interaction::query()->count())->toBe(20);
```
Orrison, line 92: Would shouldn't just check count. We should also be checking the actual records that are returned to see if they are the records we expect.

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
 
-    actingAs($teamUser);
-    livewire(ListInteractions::class)
-        ->set('tableRecordsPerPage', 20)
-        ->assertCanSeeTableRecords($allInteractions);
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => false,
+    ]);
+
+    expect(Interaction::query()->count())->toBe(20);
```
Orrison, line 112: Would shouldn't just check count. We should also be checking the actual records that are returned to see if they are the records we expect.

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
-        ->assertCanSeeTableRecords($allInteractions);
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => true,
+    ]);
+
+    Interaction::factory()->count(10)->create([
+        'is_confidential' => false,
+    ]);
+
+    expect(Interaction::query()->count())->toBe(20);
```
Orrison, line 125: Would shouldn't just check count. We should also be checking the actual records that are returned to see if they are the records we expect.

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
```diff
@@ -35,55 +35,92 @@
...
+});
+
+test('Interactions model with fetch data count for team user', function () {
     $teamUser = User::factory()->licensed(LicenseType::cases())->create();
     $teamUser->givePermissionTo('interaction.view-any');
```
Orrison, line 75: I don't thin you need to give this permission is this test or any of the others, right? The global scope doesn't check permission. Just if super admin.

Orrison (changes requested):

app-modules/interaction/tests/Models/Scopes/InteractionConfidentialScopeTest.php
amit-canyon: I added this because I encountered an error without the factory.

image

Orrison: Ah, hmm. Looking into it, perhaps try replacing the factory call here with `Interaction::bootHasGlobalScopes()`.

The global scopes I guess are not loaded till that is called, but the reason the factory fixes this is because the factory setup will call that method.

Orrison (changes requested):

Orrison (approved):

PR #1291, opened 2025-02-07
[ADVAPP-1180]: Optimize the display of interaction data for the default display of interactions

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1180

### Technical Description

> Optimize the display of interaction data for the default display of interactions

> No

> No


app-modules/interaction/src/Filament/Concerns/HasManyMorphedInteractionsTrait.php
```diff
@@ -142,8 +121,9 @@ function (Interaction $record) {
                 TextColumn::make('created_at')
                     ->state(fn ($record) => $record->end_datetime ? $record->end_datetime->diffForHumans($record->start_datetime, CarbonInterface::DIFF_ABSOLUTE, true, 6) : '-')
                     ->label('Duration'),
-                TextColumn::make('subject'),
-                TextColumn::make('description'),
+                TextColumn::make('subject')
+                    ->icon(fn ($record) => $record->is_confidential ? 'heroicon-m-lock-closed' : null)
+                    ->tooltip(fn ($record) => $record->is_confidential ? 'Confidential' : null),
```
Orrison, line 126: This isn't well described in the ticket, but let's move this to be just before `Start Time` so that by default, it is the first item in the list.

In fact let's then move all the toggled hidden by default columns to be just after `Duration` as well.

Orrison (changes requested):

PR #1307, opened 2025-02-19
[ADVAPP-1190]: Presentation of birthdates after CSV upload for students presents as a day earlier than expected

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1190

### Technical Description

> Presentation of birthdates after CSV upload for students presents as a day earlier than expected

> No

> No


app/Providers/FilamentServiceProvider.php
Orrison, line 1: @danharrin what do you think of this?

I remember we were discussing that we maybe don't want to apply timezone to `time` entries/columns right? So should this be changed?

And perhaps it would make more sense if the check was just whether or not it IS a `datetime` column?

Orrison (changes requested):

app/Providers/FilamentServiceProvider.php
amit-canyon: @Orrison  Anything is pending from my side?

amit-canyon: If, yes please explain little bit more here.
Thanks.

danharrin: I think this looks correct as it is

danharrin: I also think we need to ignore the non-datetime ones (also for the tooltip/icons) since the timezone offset depends on the day its being applied

Orrison: @amit-canyon let's change each of these `timezone`, `tooltip`, and `hintIcon` entries to instead return the value if `isDateTime()` is `true`. And `null` if it is not.

Orrison (changes requested):

PR #1309, opened 2025-02-20
[ADVAPP-1208]: Deploy survey email at the end of the case management experience when enabled.

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1208

### Technical Description

> Deploy survey email at the end of the case management experience when enabled.

> No

> Yes.

### DIFF
--- a/app-modules/case-management/src/Notifications/SendClosedCaseFeedbackNotification.php
+++ b/app-modules/case-management/src/Notifications/SendClosedCaseFeedbackNotification.php
@@ -1,0 +1,33 @@
+<?php
+
+namespace AdvisingApp\CaseManagement\Notifications;
+
+use AdvisingApp\CaseManagement\Models\CaseModel;
+use AdvisingApp\Notification\Notifications\Messages\MailMessage;
+use Illuminate\Notifications\Notification;
+
+class SendClosedCaseFeedbackNotification extends Notification
+{
+    public function __construct(
+        protected CaseModel $case,
+    ) {}
+
+    /**
+     * @return array<int, string>
+     */
+    public function via(object $notifiable): array
+    {
+        return ['mail'];
+    }
+
+    public function toMail(object $notifiable): MailMessage
+    {
+        return MailMessage::make()
+            ->subject("Feedback survey for {$this->case->case_number}")
+            ->greeting("Hi {$notifiable->display_name},")
+            ->line('To help us serve you better in the future, we’d love to hear about your experience with our support team.')
+            ->action('Rate Service', route('feedback.case', $this->case->id))
+            ->line('We appreciate your time and we value your feedback!')
+            ->salutation('Thank you.');
+    }
+}
--- a/portals/resource-hub/src/Services/Consumer.js
+++ b/portals/resource-hub/src/Services/Consumer.js
@@ -31,8 +31,8 @@
 
 </COPYRIGHT>
 */
-import axios from '@/Globals/Axios.js';
-import { useTokenStore } from '@/Stores/token.js';
+import axios from '../Globals/Axios.js';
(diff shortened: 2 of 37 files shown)


app-modules/case-management/src/Notifications/SendClosedCaseFeedbackNotification.php
```diff
@@ -0,0 +1,67 @@
...
+        return ['mail'];
+    }
+
+    public function toMail(object $notifiable): MailMessage
+    {
+        return MailMessage::make()
+            ->subject("Feedback survey for {$this->case->case_number}")
+            ->greeting("Hi {$notifiable->display_name},")
+            ->line('To help us serve you better in the future, we’d love to hear about your experience with our support team.')
+            ->action('Rate Service', route('feedback.case', $this->case->id))
```
Orrison, line 63: Let's use `getKey()` here instead of `id`

portals/resource-hub/src/Services/Consumer.js
```diff
@@ -31,8 +31,8 @@
 
 </COPYRIGHT>
 */
-import axios from '@/Globals/Axios.js';
-import { useTokenStore } from '@/Stores/token.js';
+import axios from '../Globals/Axios.js';
+import { useTokenStore } from '../Stores/token.js';
```
Orrison, line 35: Was this change needed for something? Did the `@` alias not work here?

portals/resource-hub/src/Services/DetermineIfUserIsAuthenticated.js
```diff
@@ -31,7 +31,7 @@
 
 </COPYRIGHT>
 */
-import { consumer } from '@/Services/Consumer.js';
+import { consumer } from './Consumer.js';
```
Orrison, line 34: Was this change needed for something? Did the `@` alias not work here?

widgets/case-feedback-form/src/Components/Footer.vue
```diff
@@ -0,0 +1,63 @@
...
+<template>
+    <div class="flex w-full flex-col">
+        <div class="mt-4 flex w-full justify-center">
+            <img class="block h-5" :src="logo" alt="Advising App Logo" />
+        </div>
+
+        <div class="flex w-full justify-center pb-4 pt-2">
+            <span class="w-11/12 text-center text-xs lg:w-3/4 xl:w-7/12">
+                © 2016-{{ new Date().getFullYear() }} Canyon GBS LLC. All Rights Reserved. Canyon GBS™ and Advising
+                App™ are trademarks of Canyon GBS LLC. For more information or inquiries, please visit our website at
```
Orrison, line 54: Instead of ™ for Canyon GBS and Advising App, these need to be changed to the registered trademark ®

Orrison (changes requested):

portals/resource-hub/src/Services/Consumer.js
payal-canyon: Hi @Orrison 

Yes, The `@` alias does not work here. because with this `@` aliases when build the changes then it's assuming these files from the case-feedback directory where it's actually in resource-hub directory.

portals/resource-hub/src/Services/DetermineIfUserIsAuthenticated.js
payal-canyon: Hi @Orrison 

Yes, The `@` alias does not work here. because with this `@` aliases when build the changes then it's assuming these files from the case-feedback directory where it's actually in resource-hub directory.

Orrison (approved):

PR #1316, opened 2025-02-25
[ADVAPP-1197]: Convert dashboard card headings into hyperlinks for quick access to list views

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1197

### Technical Description

> Add URLs to Prospect and Student stats for better navigation.

> No.

> No.

### DIFF
--- a/app-modules/prospect/src/Filament/Widgets/ProspectStats.php
+++ b/app-modules/prospect/src/Filament/Widgets/ProspectStats.php
@@ -37,8 +37,10 @@
 namespace AdvisingApp\Prospect\Filament\Widgets;
 
 use AdvisingApp\Alert\Enums\SystemAlertStatusClassification;
+use AdvisingApp\Prospect\Filament\Resources\ProspectResource;
 use AdvisingApp\Prospect\Models\Prospect;
 use AdvisingApp\Segment\Enums\SegmentModel;
+use AdvisingApp\Segment\Filament\Resources\SegmentResource;
 use App\Models\User;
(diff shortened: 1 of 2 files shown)


app-modules/prospect/src/Filament/Widgets/ProspectStats.php
```diff
@@ -52,28 +55,45 @@ protected function getStats(): array
...
+        $alertQueryParams = [
+            'tableFilters' => [
+                'alerts' => [
+                    'values' => array_values($alertStatusIds),
+                ],
+                'subscribed' => [
+                    'isActive' => 'true',
+                ],
+            ],
+        ];
```
Orrison, line 69: Let's move this to be inline where it is used rather than define it way up here.

app-modules/student-data-model/src/Filament/Widgets/StudentStats.php
```diff
@@ -52,28 +55,45 @@ protected function getStats(): array
...
+        $alertQueryParams = [
+            'tableFilters' => [
+                'alerts' => [
+                    'values' => array_values($alertStatusIds),
+                ],
+                'subscribed' => [
+                    'isActive' => 'true',
+                ],
+            ],
+        ];
```
Orrison, line 69: Let's move this to be inline where it is used rather than define it way up here.

Orrison (changes requested):

Orrison (approved):

PR #1349, opened 2025-03-06
[ADVAPP-1124]: Exceptions are thrown when login Azure profile image cannot be found

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1124

### Technical Description

> Exceptions are thrown when login Azure profile image cannot be found.

> No.

> No.

### DIFF
--- a/app-modules/authorization/src/Http/Controllers/SocialiteController.php
+++ b/app-modules/authorization/src/Http/Controllers/SocialiteController.php
@@ -43,6 +43,7 @@ use App\Models\User;
 use Filament\Facades\Filament;
 use Filament\Notifications\Notification;
 use Illuminate\Database\Query\Expression;
+use Illuminate\Http\Client\RequestException;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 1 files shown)


app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -114,6 +115,10 @@ public function callback(SocialiteProvider $provider)
                 }
             } catch (Throwable $e) {
                 report($e);
+            } catch (RequestException $e) {
```
Orrison, line 118: The order of `catch` in a `try/catch` matters. The more general `catch` must come later. This will result in no error ever getting to this block. So let's move this above the one for `Throwable`

Orrison (changes requested):

app-modules/authorization/src/Http/Controllers/SocialiteController.php
kandarp-canyon: Sure. Changed.

Orrison (approved):

PR #1352, opened 2025-03-07
[ADVAPP-1188]: Some tenants seem to be receiving unexpected notifications related to erroneous care team assignments

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1188

### Technical Description

> Some tenants seem to be receiving unexpected notifications related to erroneous care team assignments

> No

> No


app-modules/campaign/src/Models/CampaignAction.php
```diff
@@ -113,4 +113,11 @@ public function hasBeenExecuted(): bool
     {
         return ! is_null($this->successfully_executed_at);
     }
+
+    public function scopeLastExecutedAttemptNull(Builder $query): void
+    {
+        $query
+            ->whereNull('last_execution_attempt_at')
+            ->whereNull('last_execution_attempt_error');
+    }
```
Orrison, line 122: We don't want to add anymore local scopes into the system, if we want to use scopes we should use Tappable scopes. But seeing as this is only being used in one place right right I am thinking that we don't need a scope at all and can just apply this directly to the query.

Finally we just need to check if `last_execution_attempt_at` is `null` and don't need to worry about `last_execution_attempt_error`.

Orrison (changes requested):

app-modules/campaign/src/Models/CampaignAction.php
amit-canyon: @Orrison After testing various scenarios and removing the user as discussed, everything appears to be working as expected.

PR #1359, opened 2025-03-10
[ADVAPP-1224]: In the "Students (Overview)" report, if the number of students exceeds 99 million a display issue presents to end users

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1224

### Technical Description

> Students/Prospects overview numbers in the stats widget overflowed the cards. Now any number greater then 9,999,999 will have abbreviation based on figure

> No

> No


app/Concerns/Number.php
Orrison, line 1: I am not sure of the need for this custom class/method to do this formatting. The `Illuminate\Support\Number::format` method is probably better for our needs and directly maintained. I was assuming we would just only pass it to format if it exceeded `9999999`.
Can you explain the reasoning for this class/method? @danharrin what do you think?

Orrison (changes requested):

app/Concerns/Number.php
jayushi-canyon: @Orrison 
I created custom function because if we didn't had this I had to write this long code in 4 file:
 Stat::make('Students',
            (Student::count() > 9999999)
            ? Number::abbreviate(
                Cache::tags(['students'])
                    ->remember('students-count', now()->addHour(), function (): int {
                        return Student::count();
                    }),
                maxPrecision: 2,
            )
            : Number::format(
                Cache::tags(['students'])
                    ->remember('students-count', now()->addHour(), function (): int {
                        return Student::count();
                    }),
                maxPrecision: 2,
            ))
  And i tried by manipulating parameters of in-built function but none gave proper output. Therefor I rather thought to create our own custom method based on our requirements. Let me know your thoughts on same.

PR #1384, opened 2025-03-17
[ADVAPP-1254]: Fix the storage of job batches so they are properly managed within tenants

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1254

### Technical Description

This makes sure that the batch connection is changed to the Tenant when we switch to the Tenant context. This way job batches are stored in the Tenants database.

No

No


danharrin (approved):

danharrin:
Is there a ticket to do this on Aiding App too?

PR #1400, opened 2025-03-26
[ADVAPP-1302]: Resolve regression issue where users are unable to delete roles

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1302

### Technical Description

> Resolve regression issue where users are unable to delete roles

> No

> No


rebecca-canyon (dismissed):

danharrin (changes requested):
The issue is with the existing Delete button on the Edit page, so this isn't quite right. I investigated myself, but I can't reproduce the problem. After speaking to Joe, I am going to add a new column to the table for "Users count" to help debugging

danharrin (approved):

PR #1421, opened 2025-04-04
[ADVAPP-1238]: Integrate password policy validation into password set and password reset processes

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1238

### Technical Description

Configures the native Laravel password validation rule with custom checks, to that each time the app or a package uses it the checks will be applied correctly.

No

No


app/Multitenancy/Tasks/ConfigurePasswordValidation.php
```diff
@@ -0,0 +1,138 @@
...
+
+                        while (($commonPassword = fgets($file)) !== false) {
+                            if (Str::lower(trim($commonPassword)) === Str::lower(trim($value))) {
+                                fclose($file);
+
+                                $fail('The :attribute is too common. Please choose a different password.');
+
+                                return;
+                            }
+                        }
```
Orrison, line 128: Just to make this more readable, can we either:

A: Turn each of these closures to private functions in this class and reference them
B: Add a comment before each Closure naming/describing which rule it is for

Orrison (changes requested):
This looks great, just one comment. Also you will need to rebase this on `main` to get the changes to CI on here.

app/Multitenancy/Tasks/ConfigurePasswordValidation.php
danharrin: Done

Orrison (approved):

PR #1429, opened 2025-04-07
[ADVAPP-999]: Implement pruning for pipeline-educatable pivot records

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-999

### Technical Description

Introduces command to dispatch jobs to delete irrelevant records outside of the current segment.

No

No


Orrison (changes requested):
Looks good but there are some linting errors

PR #1444, opened 2025-04-11
[ADVAPP-1151]: Introduce support for use of merge values in email subjects in message center, campaign journey steps, email templates, and individual emails to students.

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1151

### Technical Description

> feat: add migration to refactor engagement subject to jsonb format

> No

> Yes. Feature Flag: `RefactorEngagementCampaignSubjectToJsonb`
migration: `app-modules/engagement/database/migrations/2025_04_16_142824_data_migration_refactor_engagement_subject_to_jsonb.php`


app-modules/campaign/src/Filament/Blocks/EngagementBatchEmailBlock.php
```diff
@@ -73,7 +74,25 @@ public function generateFields(string $fieldPrefix = ''): array
...
+                ->visible(! RefactorEngagementCampaignSubjectToJsonb::active()),
+            TiptapEditor::make($fieldPrefix . 'subject')
+                ->recordAttribute('data.subject')
+                ->label('Subject')
+                ->mergeTags($mergeTags = [
+                    'student first name',
+                    'student last name',
+                    'student full name',
+                    'student email',
+                    'student preferred name',
```
Orrison, line 87: Due to a recent change in one of Dan's ticket, this should now be `recipient` instead of `student`.

app-modules/engagement/database/migrations/2025_04_17_142824_data_migration_refactor_engagement_subject_to_jsonb.php
```diff
@@ -0,0 +1,224 @@
...
+                            ]],
+                        ]],
+                    ];
+                    DB::table('form_email_auto_replies')
+                        ->where('id', $row->id)
+                        ->update(['subject' => json_encode($json)]);
+                }
+            });
+
+            DB::table('campaign_actions')
```
Orrison, line 102: Can we use a json where query here to specify this should only be for `campaign_actions` that are for engagements?

app-modules/engagement/database/migrations/2025_04_17_142824_data_migration_refactor_engagement_subject_to_jsonb.php
```diff
@@ -0,0 +1,224 @@
...
+                        ->pluck('text')
+                        ->implode('');
+
+                    DB::table('form_email_auto_replies')
+                        ->where('id', $row->id)
+                        ->update(['subject' => $text]);
+                }
+            });
+
+            DB::table('campaign_actions')
```
Orrison, line 201: Same as above, we we can ensure this is only happening on ones that are engagements.

app-modules/engagement/src/Filament/Actions/BulkEngagementAction.php
```diff
@@ -89,7 +90,25 @@ public static function make(string $context)
...
+                            ->columnSpanFull()
+                            ->visible(! RefactorEngagementCampaignSubjectToJsonb::active()),
+                        TiptapEditor::make('subject')
+                            ->label('Subject')
+                            ->mergeTags([
+                                'student first name',
+                                'student last name',
+                                'student full name',
+                                'student email',
+                                'student preferred name',
```
Orrison, line 102: Due to a recent change in one of Dan's ticket, this should now be `recipient` instead of `student`.

app-modules/engagement/src/Filament/Actions/RelationManagerSendEngagementAction.php
```diff
@@ -145,7 +146,30 @@ protected function setUp(): void
...
+                            ->columnSpanFull()
+                            ->visible(! RefactorEngagementCampaignSubjectToJsonb::active()),
+                        TiptapEditor::make('subject')
+                            ->label('Subject')
+                            ->mergeTags([
+                                'student first name',
+                                'student last name',
+                                'student full name',
+                                'student email',
+                                'student preferred name',
```
Orrison, line 158: Due to a recent change in one of Dan's ticket, this should now be `recipient` instead of `student`.

app-modules/engagement/src/Models/Engagement.php
```diff
@@ -271,6 +288,17 @@ public function getDeliveryMethod(): NotificationChannel
         return $this->channel;
     }
 
+    /**
+     * @todo Remove this dynamic cast once `RefactorEngagementCampaignSubjectToJsonb` is removed.
+     *       Move 'subject' casting to the static `$casts` array: protected $casts = ['subject' => 'array'];
+     */
+    protected function casts(): array
```
Orrison, line 295: Does this overwrite the `$casts` property? If so, we will need to make sure that the other casts are in here as well.

app-modules/engagement/src/Models/EngagementBatch.php
```diff
@@ -130,4 +131,15 @@ public static function renderWithMergeTags(string $html): string
 
         return $dom->saveHTML();
     }
+
+    /**
+     * @todo Remove this dynamic cast once `RefactorEngagementCampaignSubjectToJsonb` is removed.
+     *       Move 'subject' casting to the static `$casts` array: protected $casts = ['subject' => 'array'];
+     */
+    protected function casts(): array
```
Orrison, line 139: Does this overwrite the `$casts` property? If so, we will need to make sure that the other casts are in here as well.

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormEmailAutoReply.php
```diff
@@ -92,7 +93,26 @@ public function form(FilamentForm $form): FilamentForm
...
+                        /**
+                         * @todo Remove 'temp_subject' name once `RefactorEngagementCampaignSubjectToJsonb` is removed.
+                         */
+                        TiptapEditor::make(RefactorEngagementCampaignSubjectToJsonb::active() ? 'subject' : 'temp_subject')
+                            ->mergeTags([
+                                'student first name',
+                                'student last name',
+                                'student full name',
+                                'student email',
+                                'student preferred name',
```
Orrison, line 107: Due to a recent change in one of Dan's ticket, this should now be `recipient` instead of `student`.

app-modules/form/src/Models/FormEmailAutoReply.php
```diff
@@ -92,4 +104,15 @@ public function getMergeData(Student|Prospect|null $author): array
             'student preferred name' => $author->getAttribute($author->displayPreferredNameKey()),
         ];
     }
+
+    /**
+     * @todo Remove this dynamic cast once `RefactorEngagementCampaignSubjectToJsonb` is removed.
+     *       Move 'subject' casting to the static `$casts` array: protected $casts = ['subject' => 'array'];
+     */
+    protected function casts(): array
```
Orrison, line 112: Does this overwrite the `$casts` property? If so, we will need to make sure that the other casts are in here as well.

Orrison (dismissed):
Had a few notes.

I would also like @danharrin to review this PR as well.

Finally, do we want to consider saving the subject to the EmailMessage when it is created? We do that for `content` so that we have a record of what was actually sent. I think we want to do the same here. What do y'all think @payal-canyon and @danharrin?

app-modules/engagement/src/Models/Engagement.php
payal-canyon: Hi @Orrison 

I’ve verified this, and it does not override the $casts property. Please refer to the screenshot below for confirmation.

image

Thank You!

payal-canyon:
Hi @Orrison 

We do not need a data migration for existing records in the email_messages.content column to accommodate the subject field.

We've converted the subject to a Tiptap-compatible array, and in order to maintain both forward and backward compatibility, we cast the subject to a string within MailMessage.php.

Since the subject is already stored as plain text, the only missing piece was the string cast. After adding the cast, everything is now working correctly for the engagement flow.

danharrin:
> Hi @Orrison
> 
> We do not need a data migration for existing records in the email_messages.content column to accommodate the subject field.
> 
> We've converted the subject to a Tiptap-compatible array, and in order to maintain both forward and backward compatibility, we cast the subject to a string within MailMessage.php.
> 
> Since the subject is already stored as plain text, the only missing piece was the string cast. After adding the cast, everything is now working correctly for the engagement flow.

@payal-canyon I think what Kevin says is correct, we should update the existing subjects to be JSON even though the change you made is backwards compatible. It is more easy for us to do data migrations etc if we know the data format is consistent. Just checking, did you make this change or not?

danharrin (approved):
Truly excellent work and attention to detail here Payal, well done.

PR #1446, opened 2025-04-14
advapp-1377 change the prompts to not include a signature for emails

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1377

### Technical Description
- All this should do is prompt the AI not to include a signature for emails. 
>

No
>

No
>


danharrin:
Hey @krlmrr! Just for future reference, please format the title with `[TICKET-123]: Exact ticket name`, so that the changelog is generated correctly :)

PR #1465, opened 2025-04-21
[ADVAPP-1395]: Add a filter for Message Center for inbound messages

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1395

### Technical Description

Add a filter for Message Center for inbound messages

No

No


app-modules/engagement/src/Filament/Pages/MessageCenter.php
```diff
@@ -309,6 +313,13 @@ public function applyFilters(Builder $query, string $dateColumn, string $idColum
                     $idColumn,
                     $this->user->careTeams()->pluck('educatable_id')
                 );
+            })
+            ->when($this->filterInboundMessages === true, function (Builder $query) use ($idColumn) {
+                $query->whereIn(
+                    $idColumn,
+                    EngagementResponse::query()
+                        ->pluck('sender_id')
+                );
```
Orrison, line 322: This is not a scalable solution for this filter. Eventually there will be millions of `EngagementResponse` records. It would not be efficient to query this way, let alone would it even be possible to load all IDs into memory or an SQL query like this.

You will need to find a different way to ensure it is filtering specifically for the correct records.

Orrison (changes requested):

app-modules/engagement/src/Filament/Pages/MessageCenter.php
ketan-canyon: Done

Orrison (approved):

PR #1467, opened 2025-04-21
[ADVAPP-1397]: Update the iconography in message center to use the standard icons for email or text messages regardless of direction

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1397

### Technical Description

> Update the iconography in message center to use the standard icons for email or text messages regardless of direction.

> No.

> No.


app-modules/timeline/src/Timelines/EngagementResponseTimeline.php
```diff
@@ -50,7 +51,11 @@ public function __construct(
 
     public function icon(): string
     {
-        return 'heroicon-o-arrow-small-left';
+        return match ($this->engagementResponse->getDeliveryMethod()) {
+            NotificationChannel::Email => 'heroicon-o-envelope',
+            NotificationChannel::Sms => 'heroicon-o-chat-bubble-left',
+            default => 'heroicon-o-arrow-small-right',
```
Orrison, line 57: You can just throw an `Exception` here as it should be one of the other two values and it will be a problem if it is not.

Orrison (changes requested):
This looks like it just covers `EngagementResponse`s. What about for outbound `Engagement`s?

app-modules/timeline/src/Timelines/EngagementResponseTimeline.php
kandarp-canyon: Hi @Orrison , I have integrated this from the EngagemetTimeline.php, which is used for the outbound Engagements. Do you still want to add the functionality for the Exception?

kandarp-canyon: I've added the exception, Please check and let me know.

app-modules/timeline/src/Timelines/EngagementResponseTimeline.php
```diff
@@ -54,7 +55,7 @@ public function icon(): string
         return match ($this->engagementResponse->getDeliveryMethod()) {
             NotificationChannel::Email => 'heroicon-o-envelope',
             NotificationChannel::Sms => 'heroicon-o-chat-bubble-left',
-            default => 'heroicon-o-arrow-small-right',
+            default => throw new InvalidArgumentException('Unsupported delivery method: ' . $this->engagementResponse->getDeliveryMethod()->value),
```
Orrison, line 58: Let's just use an `Exception`, not the language used `InvalidArgumentException`

Orrison (changes requested):
Added a comment for the exception type. Also please make sure to review these final review comments as well, not just the individual comments.

As per my last final review comment: This looks like it just covers `EngagementResponse`s. What about for outbound `Engagement`s?

kandarp-canyon:
> Added a comment for the exception type. Also please make sure to review these final review comments as well, not just the individual comments.
> 
> As per my last final review comment: This looks like it just covers `EngagementResponse`s. What about for outbound `Engagement`s?

Hey @Orrison , I have added the exception for Engagement too. Please check.

Orrison (approved):

PR #1491, opened 2025-04-25
[ADVAPP-1422]: Introduce the ability to duplicate roles

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1422

### Technical Description

Introduce the ability to duplicate roles

No

No


app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
```diff
@@ -63,6 +69,42 @@ public function getTabs(): array
 
     public function table(Table $table): Table
     {
+        $duplicateRoleAction = fn (): Action => Action::make('duplicateRole')
```
Orrison, line 72: What is the purpose of this beign defined in a variable and then used on line `138` like this? It can just be defined inline.

app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
```diff
@@ -63,6 +69,42 @@ public function getTabs(): array
...
+            ->modalHeading('Duplicate Role')
+            ->modalSubmitActionLabel('Duplicate')
+            ->form([
+                TextInput::make('name')
+                    ->label('Name')
+                    ->required(),
+            ])
+            ->action(function (Role $record, array $data) {
+                try {
+                    return DB::transaction(function () use ($record, $data) {
```
Orrison, line 85: Rather than wrap this in `DB::transaction` with a Closure and a use statement, it could be better to just use `DB::beginTransaction()`, `DB::commit`, and `DB::rollback()` in the `catch`.

Orrison (changes requested):

app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
ketan-canyon: Done

Orrison (approved):

PR #1495, opened 2025-04-28
[ADVAPP-1337]: Improve deliverability of email messaging by introducing rate limiting in alignment with our infrastructure providers requirements

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1337

### Technical Description

Introduces rate limiting for email notifications to stop us from sending our current limit of 14/s.

I also changed all email jobs to not limit by attempts, but instead `retryUntil` 2 hours.

No

No

### DIFF
--- a/app-modules/engagement/src/Jobs/CreateBatchedEngagement.php
+++ b/app-modules/engagement/src/Jobs/CreateBatchedEngagement.php
@@ -40,6 +40,7 @@ use AdvisingApp\Engagement\Models\Engagement;
 use AdvisingApp\Engagement\Models\EngagementBatch;
 use AdvisingApp\Engagement\Notifications\EngagementNotification;
 use AdvisingApp\Notification\Models\Contracts\CanBeNotified;
+use DateTime;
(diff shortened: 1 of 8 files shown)


app-modules/engagement/src/Jobs/CreateBatchedEngagement.php
```diff
@@ -56,11 +58,26 @@ class CreateBatchedEngagement implements ShouldQueue
...
 
+    /**
+     * @return array<object>
+     */
+    public function middleware(): array
+    {
+        return [new RateLimitedWithRedis('notification')];
+    }
+
+    public function retryUntil(): DateTime
```
danharrin, line 76: Please swap this out for `DateTimeInterface` so that if the `now()` implementation gets switched to being immutable, this code doesn't break

danharrin (changes requested):

danharrin (approved):

PR #1496, opened 2025-04-29
[ADVAPP-1412]: Some jobs that fail to suceed are not being reported to centralized error reporting system

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1412

### Technical Description

Ensure jobs with a `failed` method send the exception to Sentry, per https://docs.sentry.io/platforms/php/guides/laravel/usage/#queue-jobs

No

No


danharrin (approved):
I would have never known it was a problem. They should introduce some event that allows it to be reported even if its caught

PR #1498, opened 2025-04-30
[ADVAPP-1443]: Extend global administration to define the API version to be used with a specific model.

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1443

### Technical Description

> Extend global administration to define the API version to be used with a specific model..

> No.

> No.


Orrison:
@kandarp-canyon sorry we are going to cancel this one as the API version is too tied to the way the code calls the API, so it can no longer be User defined.

Thank you for your work on this one.

PR #1503, opened 2025-05-01
[ADVAPP-1431]: Allow cancelation of campaign journey steps that have not executed yet

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1431

### Technical Description

> Allow cancelation of campaign journey steps that have not executed yet

> No

> CancelCampaignAction

### DIFF
--- a/app-modules/campaign/src/Actions/ExecuteCampaignAction.php
+++ b/app-modules/campaign/src/Actions/ExecuteCampaignAction.php
@@ -45,29 +45,28 @@ use Illuminate\Foundation\Bus\Dispatchable;
 
 class ExecuteCampaignAction implements ShouldQueue, ShouldBeUnique
 {
-  use Dispatchable;
+    use Dispatchable;
 
-  public int $tries = 3;
+    public int $tries = 3;
 
-  public int $timeout = 600;
+    public int $timeout = 600;
 
-  public int $uniqueFor = 600 * 3;
+    public int $uniqueFor = 600 * 3;
 
-  public function __construct(
-    public CampaignAction $action
-  ) {}
+    public function __construct(
+        public CampaignAction $action
+    ) {}
 
-  public function uniqueId(): string
-  {
-    return Tenant::current()->getKey() . ':' . $this->action->getKey();
-  }
-
-  public function handle(): void
-  {
-    if (CancelCampaignAction::active() && $this->action->cancelled_at !== null) {
-
-      return;
+    public function uniqueId(): string
+    {
+        return Tenant::current()->getKey() . ':' . $this->action->getKey();
     }
-    $this->action->execute();
-  }
+
+    public function handle(): void
+    {
+        if (CancelCampaignAction::active() && $this->action->cancelled_at !== null) {
+            return;
+        }
+        $this->action->execute();
+    }
 }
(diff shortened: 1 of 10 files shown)


app-modules/campaign/src/Filament/Resources/CampaignResource/RelationManagers/CampaignActionsRelationManager.php
```diff
@@ -77,11 +79,20 @@ public function form(Form $form): Form
...
         return $table
             ->recordTitleAttribute('id')
             ->modifyQueryUsing(fn (QueryBuilder $query) => $query->orderBy('execute_at', 'ASC'))
             ->columns([
                 TextColumn::make('type')->label('Step Type'),
+                TextColumn::make('cancelled_at')
+                    ->label('')
+                    ->formatStateUsing(fn () => 'Cancelled')
+                    ->color('danger')
+                    ->hidden(fn (?CampaignAction $record) => ! CancelCampaignAction::active() || $record?->cancelled_at !== null)
```
Orrison, line 94: I could be wrong here but, can `$record` even be `nullable`? It wouldn't make sense for a row to exist and call this function if it wasn't defined.

```suggestion
                    ->hidden(fn (CampaignAction $record) => ! CancelCampaignAction::active() || $record->cancelled_at !== null)
```

app-modules/campaign/src/Actions/ExecuteCampaignAction.php
```diff
@@ -63,6 +64,9 @@ public function uniqueId(): string
 
     public function handle(): void
     {
+        if (CancelCampaignAction::active() && $this->action->cancelled_at !== null) {
+            return;
+        }
```
Orrison, line 69: We should give code some room to breath in most scenarios. It makes it a lot easier to read.

```suggestion
        if (CancelCampaignAction::active() && $this->action->cancelled_at !== null) {
            return;
        }
        
```

app/Models/Scopes/CampaignActionNotCancelled.php
Orrison, line 1: Please put this Scope in the same folder / namespace and module as the `CampaignAction` Model it is to be used with.

app-modules/campaign/src/Actions/ExecuteCampaignActions.php
```diff
@@ -37,6 +37,7 @@
 namespace AdvisingApp\Campaign\Actions;
 
 use AdvisingApp\Campaign\Models\CampaignAction;
+use App\Models\Scopes\CampaignActionNotCancelled;
```
Orrison, line 40: The namespace path on here will need to be fixed when it is moved.

Orrison (changes requested):
Looks good. Just a couple of small notes.

app-modules/campaign/src/Actions/ExecuteCampaignAction.php
jayushi-canyon: But when I run the format command it is removing the space

app-modules/campaign/src/Filament/Resources/CampaignResource/RelationManagers/CampaignActionsRelationManager.php
Orrison: Hmm, got it. I guess Filament may be loading in a null row. Alright.

Orrison (approved):

PR #1511, opened 2025-05-05
[ADVAPP-1461]: Improve the language used on the duplicate role modal overlay

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1461

### Technical Description

> Improve the language used on the duplicate role modal overlay

> No

> No

### DIFF
--- a/app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
+++ b/app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
@@ -100,14 +100,14 @@ class ListRoles extends ListRecords
             ])
             ->actions([
                 Action::make('duplicateRole')
-                    ->label('Duplicate')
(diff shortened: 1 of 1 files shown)


app-modules/authorization/src/Filament/Resources/RoleResource/Pages/ListRoles.php
```diff
@@ -100,14 +100,14 @@ public function table(Table $table): Table
             ])
             ->actions([
                 Action::make('duplicateRole')
-                    ->label('Duplicate')
```
Orrison, line 103: Are we sure we want to remove this `label` here? That changes the label of the table action button from `Duplicate` to `Duplicate role`. Which I don't believe was asked for in this ticket.

Orrison (changes requested):

Orrison (approved):

PR #1554, opened 2025-05-20
[ADVAPP-1482]: Setup Model applicability for the Jina AI Model settings

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1482

### Technical Description

Setup Model Name and Applicability for Jina AI Model adds new settings and renames old ones to match the true Model name.

No

No


app-modules/ai/src/Enums/AiModel.php
```diff
@@ -133,6 +137,8 @@ public function getServiceClass(): string
             self::OpenAiGpt41Mini => OpenAiGpt41MiniService::class,
             self::OpenAiGpt41Nano => OpenAiGpt41NanoService::class,
             self::OpenAiGptO4Mini => OpenAiGptO4MiniService::class,
+            // We don't have a service for this yet, but we can use the test service for now
```
danharrin, line 140: What is the reason we need to add this here?

IMO, the Jina dropdown for applicable features should only include the research advisor, so we should never need to call this method. Selecting anything other than the research advisor would not work. Non-Jina models should not be able to select the research advisor as a feature, as they don't work.

As per your comment in Jira, I think a `getResearchServiceClass()` method and interface should be introduced when we add a second model to the research advisor implementation, but at the moment we should only have one option.

danharrin (dismissed):

app-modules/ai/src/Enums/AiModel.php
Orrison: @danharrin I made you first requested change but am going to create a separate ticket for the other.

app-modules/ai/src/Enums/AiModel.php
```diff
@@ -135,6 +140,7 @@ public function getServiceClass(): string
             self::OpenAiGptO4Mini => OpenAiGptO4MiniService::class,
             self::OpenAiGptTest => OpenAiGptTestService::class,
             self::Test => TestAiService::class,
+            default => throw new Exception('No Service class found for this model.'),
         };
     }
```
rebecca-canyon, line 145: Does Jina not need to be added to this function?

rebecca-canyon (commented):
Looks good to me, but there is a failing test

rebecca-canyon (approved):

PR #1558, opened 2025-05-21
[ADVAPP-1487]: An error is thrown to Sentry if an Azure SSO login does not have a profile photo.

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1487

### Technical Description

> An error is thrown to Sentry if an Azure SSO login does not have a profile photo.

> No.

> No.


Orrison (changes requested):
@kandarp-canyon you can remove the test command and restore the `retry` that was removed in the Controller.

I am seeing possibly what happened. You actually fixed this a while ago in: https://github.com/canyongbs/advisingapp/pull/1349/files

But some how the changes are not here in `main` despite it merging in. So this might have been lost from a rebase or something.

Please restore that same change you did in the linked PR

PR #1568, opened 2025-05-27
[ADVAPP-1454]: Saved research reports can be emailed directly to colleagues, including formatted answers and citations

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1454

### Technical Description

> Saved research reports can be emailed directly to colleagues, including formatted answers and citations

> No

> No

### DIFF
--- a/app-modules/research/src/Notifications/ResearchTranscriptNotification.php
+++ b/app-modules/research/src/Notifications/ResearchTranscriptNotification.php
@@ -49,6 +49,7 @@ use Illuminate\Bus\Queueable;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Notifications\AnonymousNotifiable;
 use Illuminate\Notifications\Notification;
+use Illuminate\Support\Str;
 
 class ResearchTranscriptNotification extends Notification implements ShouldQueue, HasAfterSendHook
 {
@@ -85,7 +86,7 @@ class ResearchTranscriptNotification extends Notification implements ShouldQueue
         }
         $link = ManageResearchRequests::getUrl(['researchRequest' => $this->researchRequest]);
 
-        $message->line(str('Reasoning: <br>' . $this->researchRequest->results)
+        $message->line(str(Str::after($this->researchRequest->results, '</think>'))
             ->toHtmlString());
 
         if ($this->note) {


app-modules/research/src/Filament/Pages/ManageResearchRequests/Concerns/CanManageFolders.php
```diff
@@ -111,6 +122,68 @@ public function newFolderAction(): Action
             ->modalSubmitAction(fn (StaticAction $action) => $action->color('primary'));
     }
 
+    public function emailResearchRequestAction(): Action
```
Orrison: I feel it may make more sense for this to be in `app-modules/research/src/Filament/Pages/ManageResearchRequests/Concerns/CanManageRequests.php`

Since that file has to do with singular requests while this has to do with the folders.

What do you think?

app-modules/research/src/Jobs/PrepareResearchRequestEmailing.php
Orrison, line 1: I think this is how the original is written but looking at it all together now it is a little problematic.

The `EmailResearchRequest` job is not needed is all it does is end up dispatching another job via `notify`.

Because of those two degrees of separation it actually isn't a good idea to notify the User that anything has been sent yet because we don't actually know if it is going to succeed or not.

It could be best to instead change `EmailResearchRequest` to accept a boolean of whether or not it should notify the sender or not.

Then in the `handle` call `notifyNow` instead of `notify` and then trigger the notification if the passed value is true.

You would pass `true` for User one and false for a team one as the batch will take care of the notification in the end.

The batch should allow failures and then add `finally` instead of a `then` to inform the user of how many out of the users of the team succeeded to sent `

app-modules/research/src/Notifications/ResearchTranscriptNotification.php
```diff
@@ -0,0 +1,108 @@
...
+            ->settings($this->resolveNotificationSetting($notifiable))
+            ->greeting("Hello {$notifiable->name},");
+
+        $senderIsNotifiable = $this->sender->is($notifiable);
+
+        if ($senderIsNotifiable) {
+            $message->subject("Research Transcript: {$this->researchRequest->topic}")
+                ->line("Here is a copy of research on: {$this->researchRequest->topic}");
+        } else {
+            $message->subject("A Research Transcript has been shared with you: {$this->researchRequest->topic}")
```
Orrison:
```suggestion
            $message->subject("A Research Transcript has been shared with you")
```

jayushi-canyon: @Orrison I have added topic as in ticket product ha mentioned that email should contain 'The email includes the report title, answer, citations, and a link back to the report inside Advising App.'

Orrison: RIght, that is fine. But you have added the topic already below this (and in the item above) through a `line` so it is actually output in the email. Which is what you want.

The topic may be too large to fit in a subject so I am asking the topic to be removed from the subject and kept in the email itself.

Orrison (changes requested):

app-modules/research/src/Notifications/ResearchTranscriptNotification.php
Orrison, line 1: This all looks great! Though I think we need a little extra formatting in the output of the email due to the way the AI generates things.

If you look at `app-modules/research/resources/views/results.blade.php` and `app-modules/research/resources/js/results.js` right now, you can see that the AI returns it's response with it's "Reasoning" contained between HTML looking `<think>...</think>` tags.

And then the rest of the content, and then some footnotes.

We then split up that content by grabbing everything before `</think>`, stripping out `<think>` and placing that in a reasoning section with smaller text size and smaller character spacing so that it is distinct form the actual report.

We also do some similar styling with the footnotes section.

We may want to customize this email, maybe with a blade file (https://laravel.com/docs/12.x/notifications#other-mail-notification-formatting-options) or some other way to apply this formatting so that the reasoning and footnotes are visually different from the report itself.

Orrison (changes requested):
Great work! Just one improvement on the email output needed.

app-modules/research/src/Notifications/ResearchTranscriptNotification.php
Orrison: Let's actually strip (remove) all reasoning and not include it in the email

Orrison: Anything after a `</think>` tag

app-modules/research/src/Notifications/ResearchTranscriptNotification.php
```diff
@@ -0,0 +1,108 @@
...
+            $message->subject('Research Transcript')
+                ->line("Here is a copy of research on: {$this->researchRequest->topic}");
+        } else {
+            $message->subject('A Research Transcript has been shared with you')
+                ->line("Here is a copy of {$this->sender->name}'s research on: {$this->researchRequest->topic}");
+        }
+        $link = ManageResearchRequests::getUrl(['researchRequest' => $this->researchRequest]);
+
+        $message->line(str('Reasoning: <br>' . $this->researchRequest->results)
+            ->toHtmlString());
```
Orrison: This can be modified to remov the `'Reasoning: <br>' . `.

And then run `$this->researchRequest->results` through `Str::after` getting everything after `</think>`

Orrison (approved):

PR #1598, opened 2025-06-05
[ADVAPP-1481]: Rename Campaign Action permission group and permissions to Journey Step

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1481

### Technical Description

> Rename Campaign Action permission group and permissions to Journey Step.

> No

> Yes. Feature Flag: `JourneyStepPermissionRename`

### DIFF
--- a/app/Features/JourneyStepPermissionRename.php
+++ b/app/Features/JourneyStepPermissionRename.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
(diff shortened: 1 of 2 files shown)


app-modules/campaign/src/Policies/CampaignActionPolicy.php
Orrison, line 1: Let's use the wildcard `*` permission here instead of ->getKey() on the models

database/migrations/2025_06_05_115720_data_activate_journey_step_permission_rename.php
Orrison, line 1: It would be better to delete this migration and move the activation/deactivation to the end of your other migration.

Wrapping it all in a single DB transaction so it happens at once.

That way no User would lose access.

PR #1610, opened 2025-06-12
[ADVAPP-1583]: Improve the organization and navigation of profile settings for product users & [ADVAPP-1573]: Update labeling for branding bar to match other sections in profile settings

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1583
- https://canyongbs.atlassian.net/browse/ADVAPP-1573

### Technical Description

Split profile settings into a cluster of multiple pages.
Modify labels for Branding Bar setting.

No

No


app/Filament/Pages/ArtificialIntelligence.php
```diff
@@ -0,0 +1,323 @@
...
+
+    protected static ?string $cluster = ProfileSettings::class;
+
+    protected static ?string $slug = 'artificial-intelligence';
+
+    protected static ?string $title = 'Artifical Intelligence';
+
+    protected static ?int $navigationSort = 40;
+
+    protected static bool $shouldRegisterNavigation = true;
```
Orrison: I don't think this is needed as it's default is `true`

resources/views/filament/pages/profile-save.blade.php
```diff
@@ -0,0 +1,47 @@
...
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+--}}
+@php
+    use AdvisingApp\MultifactorAuthentication\Livewire\MultifactorAuthenticationManagement;
+@endphp
```
Orrison: Is this import still needed in this file?

app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -185,7 +185,7 @@ public function panel(Panel $panel): Panel
             ->userMenuItems([
                 MenuItem::make()
                     ->label('Profile Settings')
-                    ->url(fn () => EditProfile::getUrl())
+                    ->url(fn () => ProfileInformation::getUrl())
```
Orrison: @danharrin had mentioned that it was possible to just pass the URL from the Cluster class itself. Which would then go to the first page in the cluster.

Were you able to see if that was possible here?

app/Filament/Pages/ArtificialIntelligence.php
```diff
@@ -0,0 +1,323 @@
...
+        $this->data['passwordConfirmation'] = null;
+
+        $this->getSavedNotification()?->send();
+
+        $this->dispatch('refresh-branding-bar');
+
+        if ($redirectUrl = $this->getRedirectUrl()) {
+            $this->redirect($redirectUrl);
+        }
+    }
```
Orrison: I am noticing that these, and many other methods are exactly the same between all of the pages.

It would be best if we created something like an abstract base `ProfilePage` class or something like that that contained all of the shared method that are exactly the same and have all of these new pages extend that.

What do you think?

app/Filament/Pages/BrandingBar.php
```diff
@@ -0,0 +1,321 @@
...
+                        $this->getEmailFormComponent(),
+                        $this->getPasswordFormComponent(),
+                        $this->getPasswordConfirmationFormComponent(),
+                    ])
+                    ->operation('edit')
+                    ->model($this->getUser())
+                    ->statePath('data'),
+            ),
+        ];
+    }
```
Orrison: I don't think here (or in any of the other profile page files OTHER than the one with these fields, the ProfileInformation one, these are needed) Since these forms fields are not on those pages.

app/Filament/Pages/ProfileInformation.php
```diff
@@ -583,7 +457,7 @@ protected function getCancelFormAction(): Action
     protected function getSaveFormAction(): Action
     {
         return Action::make('save')
-            ->label(__('filament-panels::pages/auth/edit-profile.form.actions.save.label'))
+            ->label('Save')
```
Orrison: I think it may be fine for this to remain what it was. Especially since it looks like that is what you have for the other pages.

Orrison (changes requested):

app/Filament/Pages/ProfileInformation.php
rebecca-canyon: The ticket calls for changing the label to just Save, but thank you for pointing out that I hadn't changed it on the other pages!

app/Filament/Pages/BrandingBar.php
```diff
@@ -0,0 +1,72 @@
...
+{
+    protected static ?string $slug = 'branding-bar';
+
+    protected static ?string $title = 'Branding Bar';
+
+    protected static ?int $navigationSort = 50;
+
+    public static function canAccess(): bool
+    {
+        return app(CollegeBrandingSettings::class)->is_enabled && app(CollegeBrandingSettings::class)->dismissible;
```
Orrison: I think it would be good to add `&& parent::canAccess()` to the end here as well. Just so any Filament or system level default checks are also applied.

app/Filament/Pages/ConnectedAccounts.php
```diff
@@ -0,0 +1,136 @@
...
+{
+    protected static ?string $slug = 'connected-accounts';
+
+    protected static ?string $title = 'Connected Accounts';
+
+    protected static ?int $navigationSort = 60;
+
+    public static function canAccess(): bool
+    {
+        return self::getConnectedAccounts()->count() > 0;
```
Orrison: I think it would be good to add `&& parent::canAccess()` to the end here as well. Just so any Filament or system level default checks are also applied.

PR #1612, opened 2025-06-13
[ADVAPP-1579]: Some users have reported that the total student count in the Student Overview report does not match the current cumulative count

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1579

### Technical Description

Uses the existing `created_at_source` and `updated_at_source` columns to save data correctly in the student resouce. The ticket doesnt mention imports, do we have another ticket for those?

Ensures deleted students are not taken into account by custom student queries

Removes `$refresh` dispatches from widgets since its not required.

No

No

### DIFF
--- a/app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/CreateStudent.php
+++ b/app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/CreateStudent.php
@@ -315,6 +315,7 @@ class CreateStudent extends CreateRecord
         $student = $this->getRecord();
 
         $student->created_at_source = now();
+        $student->updated_at_source = now();
         $student->primaryEmailAddress()->associate($student->emailAddresses()->first());
         $student->primaryPhoneNumber()->associate($student->phoneNumbers()->first());
         $student->primaryAddress()->associate($student->addresses()->first());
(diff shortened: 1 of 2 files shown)


app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/CreateStudent.php
```diff
@@ -311,9 +311,10 @@ public function form(Form $form): Form
 
     protected function afterCreate(): void
     {
-        /** Student $student */
+        /** @var Student $student */
         $student = $this->getRecord();
 
+        $student->created_at_source = now();
```
Orrison, line 317: I beleive Laravel sets `updated_at` during create as well right? So do we want to replicate that here too?

Orrison (changes requested):
From what I remember of the discussion, the import was expected to be done here as well. So if we can do that as well to make sure the created and updated field are able to added that would be great.

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/CreateStudent.php
danharrin: yeah sure

Orrison (approved):

PR #1617, opened 2025-06-13
[ADVAPP-1581]: Improve reporting by introducing optional date filters on all the reports in the report library

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1581

### Technical Description

> Implement filters and refresh functionality in student reports and widgets.

> No

> No

### DIFF
--- a/app-modules/report/resources/views/filament/pages/report.blade.php
+++ b/app-modules/report/resources/views/filament/pages/report.blade.php
@@ -1,3 +1,36 @@
+{{--
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Advising App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+--}}
 @php
     use AdvisingApp\Report\Filament\Widgets\RefreshWidget;
 
--- a/app-modules/report/src/Abstract/Concerns/HasFiltersForm.php
+++ b/app-modules/report/src/Abstract/Concerns/HasFiltersForm.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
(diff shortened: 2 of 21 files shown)


app-modules/report/src/Filament/Widgets/MostEngagedProspectsTable.php
```diff
@@ -64,12 +68,29 @@ public function refreshWidget() {}
...
+
         return $table
             ->query(
                 Prospect::select('id', 'full_name', 'primary_email_id', 'status_id', 'created_by_id', 'created_at')
                     ->with('primaryEmailAddress:id,address')
                     ->with(['status', 'createdBy:id,name'])
-                    ->withCount('engagements')
+                    ->withCount([
+                        'engagements as engagements_count' => fn ($query) => $query
+                            ->when($startDate && $endDate, fn ($q) => $q->whereBetween('created_at', [$startDate, $endDate])),
```
Orrison: Let's not use single character variable names. You can still just call this `$query` because you are not using the other `$query` here. And please also type hint these. This goes for all similar entries in this PR.

app-modules/report/src/Filament/Widgets/MostRecentStudentsTable.php
```diff
@@ -69,24 +73,35 @@ public function refreshWidget() {}
...
+                $endDate = filled($this->filters['endDate'] ?? null)
+                    ? Carbon::parse($this->filters['endDate'])->endOfDay()
+                    : null;
+
+                return Student::whereIn($key, function ($query) use ($key, $startDate, $endDate) {
+                    $query->select($key)
+                        ->from((new Student())->getTable())
+                        ->whereNotNull('created_at_source')
+                        ->whereNull('deleted_at')
+                        ->when($startDate, function ($query) use ($startDate, $endDate) {
```
Orrison: You will want to confirm that both `$startDate` and `$endDate` are not null in this when here. And make sure to type hint all queries

app-modules/report/src/Filament/Widgets/MostRecentTasksTable.php
```diff
@@ -72,10 +76,19 @@ public function refreshWidget() {}
...
+
+        $endDate = filled($this->filters['endDate'] ?? null)
+            ? Carbon::parse($this->filters['endDate'])->endOfDay()
+            : null;
+
         return $table
             ->query(
                 Task::query()
                     ->with(['concern'])
+                    ->when($startDate, fn ($query) => $query->whereBetween('created_at', [$startDate, $endDate]))
```
Orrison: You will want to confirm that both `$startDate` and `$endDate` are not null in this when here. And make sure to type hint all queries

app-modules/report/resources/views/filament/pages/reports.blade.php
danharrin, line 1: I don't think we need to check if the method and properties exist, they are guaranteed to exist. We only do those checks in Filament core because using the filters trait is optional.

danharrin: I think you can return the `$this->filters` from within `getWidgetData()`, and then simplify the attribute to be `:data="$this->getWidgetData()"`

danharrin: Maybe this should just be `report.blade.php` instead of `reports.blade.php`?

app-modules/report/src/Abstract/EngagementReport.php
danharrin, line 1: Maybe think about introducing `AdvisingApp\Report\Abstract\Concerns\HasFiltersForm` as a trait, and inside there add `use HasFiltersForm` from Filament and add the `filtersForm()` method, so it doesn't need to be defined three times?

app-modules/report/src/Filament/Widgets/MostRecentStudentsTable.php
```diff
@@ -69,24 +73,35 @@ public function refreshWidget() {}
...
-                            ->orderBy('created_at_source', 'desc')
-                            ->take(100);
-                    })->orderBy('created_at_source', 'desc');
-                }
-            )
+            ->query(function () {
+                $key = (new Student())->getKeyName();
+
+                $startDate = filled($this->filters['startDate'] ?? null)
+                    ? Carbon::parse($this->filters['startDate'])->startOfDay()
```
danharrin: Maybe introduce a new `AdvisingApp\Report\Filament\Widgets\Concerns\InteractsWithPageFilters` trait, that adds `use InteractsWithPageFilters` from Filament, and adds functions for fetching the Carbon instances for start/end date: `$this->getStartDate()` and `$this->getEndDate()`. This should avoid duplication across many widget classes for the date parsing logic?

danharrin: Maybe add the trait to the base `StatsOverviewReportWidget` class too so you don't need to use that across multiple stats overview widget classes

Orrison:
> @Orrison what do you think about using these long raw queries instead of using the Laravel query builder? I think the vast majority of the custom queries written here could use the query builder instead

I agree, if they can be written with the query builderI think I would prefer that. Especially so we can use like `when()` to apply the filter without having to have large repeats of the same, slightly different, query.

Orrison (approved):

PR #1625, opened 2025-06-17
[ADVAPP-1571] - Underline Display Name

krlmrr:
### Ticket(s) or GitHub Issue
- https://canyongbs.atlassian.net/browse/ADVAPP-1571

### Technical Description
- Add `underline` Tailwind class to Display Name

- No

- No

### DIFF
--- a/app-modules/engagement/resources/views/components/message-center-content.blade.php
+++ b/app-modules/engagement/resources/views/components/message-center-content.blade.php
@@ -9,22 +9,22 @@
     Notice:
 
     - You may not provide the software to third parties as a hosted or managed
-    service, where the service provides users with access to any substantial set of
-    the features or functionality of the software.
+      service, where the service provides users with access to any substantial set of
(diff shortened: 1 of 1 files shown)


app-modules/engagement/resources/views/components/message-center-content.blade.php
```diff
@@ -46,7 +46,7 @@ class="max-h-full overflow-y-auto"
                 id="message-center-content"
             >
                 <div
-                    class="sticky top-0 z-[5] flex h-12 w-full items-center justify-between bg-gray-100 px-4 dark:bg-gray-700">
+                    class="sticky top-0 z-[5] flex h-12 w-full items-center justify-between bg-gray-100 px-4 underline dark:bg-gray-700">
                     <h1 class="ml-2 flex">
                         <a
                             class="hover:underline"
```
Orrison: We can probably remove this class now since the whole thing is always underlined.

PR #1643, opened 2025-06-23
[ADVAPP-1603]: Improve messaging around Multifactor Authentication when user account is configured to use SSO

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1603

### Technical Description

> Add disabled SSO toggle for external users in profile settings

> No

> No


Orrison (changes requested):
Rather than make a while new livewire component I would maybe just modify `MultifactorAuthenticationManagement` to handle this scenario.

PR #1647, opened 2025-06-25
[ADVAPP-1601]: Introduce enhanced file management capabilities for institutional advisor and custom advisor

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1601

### Technical Description

Uploads files to LlamaParse for parsing, and then attaches them to messages when they are sent.

No

No


app-modules/ai/src/Console/Commands/FetchAiAssistantFilesParsingResults.php
```diff
@@ -0,0 +1,64 @@
...
+    public function handle(): void
+    {
+        if (! LlamaParse::active()) {
+            return;
+        }
+
+        AiAssistantFile::query()
+            ->whereNull('parsing_results')
+            ->where('created_at', '>=', now()->subHour())
+            ->each(fn (AiAssistantFile $file) => dispatch(new FetchAiAssistantFileParsingResults($file)));
```
Orrison: Should we chunk this just to be safe?

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanUploadFiles.php
```diff
@@ -73,15 +139,55 @@ public function uploadFilesAction(): Action
...
+            ->action(function (Action $action, array $data) {
                 /** @var TemporaryUploadedFile $attachment */
                 $attachment = $data['attachment'];
 
-                $this->files[] = [
-                    'temporaryUrl' => $attachment->temporaryUrl(),
-                    'mimeType' => $attachment->getMimeType(),
-                    'name' => $attachment->getClientOriginalName(),
-                ];
+                $file = new AiMessageFile();
```
Orrison, line 146: Should we wrap this all in a transaction so that the `AiMessageFile` record is not kept if this fails?

app-modules/ai/src/Filament/Resources/AiAssistantResource/Concerns/HandlesFileUploads.php
```diff
@@ -38,85 +38,67 @@
...
-    }
+        foreach ($files as $attachment) {
+            if (! ($attachment instanceof TemporaryUploadedFile)) {
+                continue;
+            }
 
-    protected function uploadFilesToAssistant(AiService $aiService, AiAssistant $assistant, array $uploadedFiles): void
-    {
-        $aiAssistantFiles = $this->createAiAssistantFiles($assistant, $uploadedFiles);
+            $file = new AiAssistantFile();
```
Orrison, line 60: Should we wrap this all in a transaction so that the `AiAssistantFile` record is not persisted if it fails?

app-modules/integration-open-ai/src/Services/BaseOpenAiResponsesService.php
```diff
@@ -89,9 +88,11 @@ public function complete(string $prompt, string $content): string
                 ->using('azure_open_ai', $this->getModel())
                 ->withClientOptions([
                     'apiKey' => $this->getApiKey(),
-                    'apiVersion' => $this->getApiVersion(),
```
Orrison, line 92: I think we still want the `apiVersion` to be based on the abstract. So we can be explicit if we want to to be `preview` or, in the future, the GA value.

app-modules/integration-open-ai/src/Services/BaseOpenAiService.php
```diff
@@ -128,11 +129,6 @@ public function createAssistant(AiAssistant $assistant): void
             'metadata' => [
                 'last_updated_at' => now(),
             ],
-            'tools' => [
-                [
-                    'type' => 'file_search',
-                ],
-            ],
```
Orrison, line 135: Won't the old chats still need this tool to search their existing vector stores?

app-modules/ai/src/Console/Commands/FetchAiAssistantFilesParsingResults.php
danharrin: `each()` does chunk, but I have switched to `eachById()` which is a little safer

app-modules/ai/src/Filament/Pages/Assistant/Concerns/CanUploadFiles.php
danharrin: We don't save the message if the HTTP call fails, its after the early return

app-modules/ai/src/Filament/Resources/AiAssistantResource/Concerns/HandlesFileUploads.php
danharrin: Same as above

app-modules/integration-open-ai/src/Services/BaseOpenAiResponsesService.php
danharrin: Done

app-modules/integration-open-ai/src/Services/BaseOpenAiService.php
danharrin: This method creates a new assistant, doesn't affect old chats

app-modules/ai/src/Console/Commands/FetchAiAssistantFilesParsingResults.php
Orrison: Ah, that's right I forgot

Orrison (approved):

PR #1657, opened 2025-06-27
[ADVAPP-1606]: Create models for Workflows

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1606

### Technical Description

Create models for Workflow, Workflow Step, Workflow Trigger, and Workflow Actions. Create feature flag, though it is not active.

No

WorkflowFeature; it is not yet active


app-modules/workflow/database/migrations/2025_07_01_003621_create_workflow_steps_table.php
```diff
@@ -0,0 +1,65 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('workflow_steps', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('details_id');
+            $table->string('details_type');
```
Orrison: I would use the UUID morph helper here so that it also creates proper indexes for you

app-modules/workflow/database/migrations/2025_07_01_003536_create_workflow_triggers_table.php
```diff
@@ -0,0 +1,60 @@
...
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('workflow_triggers', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->string('type');
+            $table->string('related_id');
+            $table->string('related_type');
```
Orrison, line 49: You will want to make sure that an index is created for this. You can look at how morph or uuidMorph helpers do it. But you are right to make this a string here I think. Just in case we were ever needing to relate a workflow trigger to an action taken on a Student.

app-modules/workflow/database/migrations/2025_07_03_041202_data_activate_workflow_feature_flag.php
Orrison, line 1: My hope is to merge this PR into `main` when it is ready, regardless of whether the entire feature is ready. So based on that, you will just need to delete this migration and either recreate it temporarily locally when working on it.
Because an environment will run this migration regardless of it's contents, then never run it again.

We probably want to come up with a good mechanism for manually toggle feature flags via easy command line action. If you want we can pair on looking at how to do that. Just let me know if/when you want to find some time.

Orrison:
Great work @rebecca-canyon! Only a few minor comments

app-modules/workflow/src/Models/WorkflowStep.php
```diff
@@ -0,0 +1,63 @@
...
+    use AuditableTrait;
+    use HasUuids;
+
+    /**
+     * @return MorphTo<Model, $this>
+     */
+    public function workflow(): MorphTo
+    {
+        return $this->morphTo();
+    }
```
Orrison: Is this not just a `BelongsTo`? I think the details is polymorphic. But this I think is just a regular relationship

Orrison (changes requested):

Orrison (approved):

PR #1676, opened 2025-07-02
[ADVAPP-1612]: Introduce automated assignment capabilities for new cases by case type in case management

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1612

### Technical Description

> Introduce automated assignment capabilities for new cases by case type in case management

> No

> Yes. Feature Flag: `AssignmentsFeature`

### DIFF
--- a/app-modules/campaign/src/Filament/Blocks/CaseBlock.php
+++ b/app-modules/campaign/src/Filament/Blocks/CaseBlock.php
@@ -124,7 +124,7 @@ class CaseBlock extends CampaignActionBlock
                     if (! $caseTypeId) {
                         return [];
                     }
-                    
+
                     $caseType = CaseType::find($caseTypeId);
 
                     if (! $caseType) {


app-modules/case-management/src/Policies/CaseModelPolicy.php
```diff
@@ -87,6 +87,14 @@ public function view(Authenticatable $authenticatable, CaseModel $case): Respons
 
     public function create(Authenticatable $authenticatable): Response
     {
+        if (! auth()->user()->isSuperAdmin()) {
+            $team = auth()->user()->team;
+
+            if (! $team?->manageableCaseTypes()->exists()) {
+                return Response::deny("You don't have permission to create cases because you're not a manager of any case types.");
```
rebecca-canyon: This is extremely minor, but I think for the sake of consistency I would change the deny message to be the same as the one for create.

app-modules/case-management/src/Services/CaseType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,78 @@
...
+        $lastAssignee = $caseType->lastAssignedUser;
+        $user = null;
+
+        if ($lastAssignee) {
+            $user = User::query()->whereRelation('team.manageableCaseTypes', 'case_types.id', $caseType->getKey())
+                ->where('name', '>=', $lastAssignee->name)
+                ->where(fn (Builder $query) => $query
+                    ->where('name', '!=', $lastAssignee->name)
+                    ->orWhere('users.id', '>', $lastAssignee->getKey()))
+                ->orderBy('name')->orderBy('id')->first();
```
rebecca-canyon, line 59: Maybe you've already tested this, but could we make the check on line 55 a strictly greater than and remove line 57?

app-modules/case-management/src/Services/CaseType/WorkloadAssigner.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        $caseType = $case->priority->type;
+
+        $lastAssignee = $caseType->lastAssignedUser;
+        $user = null;
+
+        if ($lastAssignee) {
+            $lowestCase = User::query()->whereRelation('team.manageableCaseTypes', 'case_types.id', $caseType->getKey())
+                ->withCount([
+                    'cases as case_count' => function (Builder $query) {
```
rebecca-canyon, line 58: I think the key here should just be `case_count`.

app-modules/case-management/src/Services/CaseType/WorkloadAssigner.php
```diff
@@ -0,0 +1,111 @@
...
+                                ->where('classification', '!=', SystemCaseClassification::Closed)
+                                ->whereNull('case_statuses.deleted_at');
+                        })
+                        ->whereNull('cases.deleted_at')
+                        ->whereNull('case_assignments.deleted_at');
+                }, '<=', $lowestCase)
+                ->where('name', '>=', $lastAssignee->name)
+                ->where(fn (Builder $query) => $query
+                    ->where('name', '!=', $lastAssignee->name)
+                    ->orWhere('users.id', '>', $lastAssignee->getKey()))
```
rebecca-canyon, line 85: Maybe you've already tested this, but could we make the check on line 82 a strictly greater than and remove line 84?

app-modules/case-management/src/Services/CaseType/RoundRobinAssigner.php
payal-canyon: Hie @rebecca-canyon 
I did try implementing the change you suggested, but it ended up breaking the functionality, so I hope it's okay to leave it as is for now.

For example, in Team A with 3 users and the Assignment Type set to Round Robin, the logic always assigns to the first user instead of rotating through the team members. So unfortunately, it didn't work as expected in a round-robin fashion.

app-modules/case-management/src/Services/CaseType/WorkloadAssigner.php
payal-canyon: same as above!

app-modules/case-management/src/Services/CaseType/WorkloadAssigner.php
payal-canyon: Hi @rebecca-canyon,
Just using `case_count` doesn’t work in this case because:
As far as I know, when using a closure inside `withCount()`, we need to specify the relationship name followed by an alias using as, like I did with `cases as case_count`.

`withCount()` expects a valid relationship name — optionally followed by an alias.
Since `case_count` is not a defined relationship on the model, Laravel throws an error like:
`Call to undefined relationship [case_count] on model [User].`

app-modules/case-management/src/Services/CaseType/RoundRobinAssigner.php
rebecca-canyon: Gotcha- thank you for testing it out! Leaving it as is and working sounds great, then.

app-modules/case-management/src/Services/CaseType/WorkloadAssigner.php
rebecca-canyon: Gotcha, I didn't realize that. Good to know, thank you!

PR #1682, opened 2025-07-04
[ADVAPP-1611]: Introduce email templates configuration for notification messages in case management

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1611

### Technical Description

> Introduce email templates configuration for notification messages in case management

> No

> CaseTypeEmailTemplateFeature 


app-modules/case-management/database/migrations/2025_07_02_091124_create_case_type_email_templates_table.php
```diff
@@ -0,0 +1,60 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('case_type_email_templates', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->foreignUuid('case_type_id')->constrained('case_types');
+            $table->string('type');
+            $table->unique(['case_type_id', 'type', 'role']);
```
Orrison: Does this work properly given that the `role` column is defined below this?

Regardless I think we should move this `unique` to be defined after the timestamps and a newline. Which is a common way to define constraints in a migration, separating it from the column definitions.

Orrison (changes requested):
Nice work! Just one quick note.

Also just a reminder for myself to create a ticket to finish these tests once this is merged in.

Orrison (approved):

PR #1683, opened 2025-07-08
[ADVAPP-1653]: Enhance the Research Advisor and introduce the ability to add any relevant institutional data in step 4 and any relevant website links in step 5

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1653
- https://canyongbs.atlassian.net/browse/ADVAPP-1654

### Technical Description

**Merges into ADVAPP-1655**, since these do nothing without the work in that ticket. I have opened this PR to make the future review a little more manageable.

No

No

### DIFF
--- a/app-modules/research/src/Models/ResearchRequest.php
+++ b/app-modules/research/src/Models/ResearchRequest.php
@@ -42,14 +42,18 @@ use App\Models\User;
 use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\HasMany;
+use Spatie\MediaLibrary\HasMedia;
+use Spatie\MediaLibrary\InteractsWithMedia;
 
 /**
  * @mixin IdeHelperResearchRequest
  */
-class ResearchRequest extends BaseModel
+class ResearchRequest extends BaseModel implements HasMedia
 {
     /** @use HasFactory<ResearchRequestFactory> */
     use HasFactory;
+
+    use InteractsWithMedia;
 
     protected $fillable = [
         'title',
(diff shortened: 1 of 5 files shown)


krlmrr (approved):

app-modules/research/src/Models/ResearchRequest.php
```diff
@@ -99,4 +105,9 @@ public function hasStarted(): bool
 
         return filled($this->questions->get(3)?->response);
     }
+
+    public function registerMediaCollections(): void
+    {
+        $this->addMediaCollection('files');
```
Orrison: We may want to add some sort of file size limits and mime type limits here as well. But can happen later if that's the plan.

danharrin: I added them to the media upload field, do we also add them here usually? I don't see that

danharrin:
I think this should be merged into the branch, doesn't need to be draft

app-modules/research/src/Models/ResearchRequest.php
Orrison: We have been inconsistent in doing this. But my preference would be that we define it at the collection and on the input level.

Orrison: The best place to have it is at the collection level so we never forget to handle it where we add to the collection. The file input level provides frontend UI most of the time though.

danharrin: I had added the accepted file types here, but I don't see how to define the max size. Can you help please?

Orrison: Ah, my bad. I don't think this exists. Just the file types is fine

PR #1686, opened 2025-07-08
[ADVAPP-1634]: Introduce external REST API for prospects

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1634

### Technical Description

> Introduce external REST API for prospects

> No

> No


app-modules/prospect/tests/Tenant/Http/Controllers/Api/V1/Prospects/ProspectEmailAddresses/CreateProspectEmailAddressControllerTest.php
```diff
@@ -0,0 +1,133 @@
...
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.view-any');
+    Sanctum::actingAs($user, ['api']);
+    postJson(route('api.v1.prospects.email-addresses.create', ['prospect' => $prospect], false), $createProspectEmailAddressRequestData);
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.*.update');
+    Sanctum::actingAs($user, ['api']);
+    postJson(route('api.v1.prospects.email-addresses.create', ['prospect' => $prospect], false), $createProspectEmailAddressRequestData);
```
rebecca-canyon: I believe that all three of the `postJson` methods should have an `assertForbidden` chained on to them.

app-modules/prospect/tests/Tenant/Http/Controllers/Api/V1/Prospects/ProspectEmailAddresses/UpdateProspectEmailAddressControllerTest.php
```diff
@@ -0,0 +1,133 @@
...
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.view-any');
+    Sanctum::actingAs($user, ['api']);
+    patchJson(route('api.v1.prospects.email-addresses.update', ['prospect' => $prospect, 'prospectEmailAddress' => $prospectEmailAddress], false), $updateProspectEmailAddressRequestData);
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.*.update');
+    Sanctum::actingAs($user, ['api']);
+    patchJson(route('api.v1.prospects.email-addresses.update', ['prospect' => $prospect, 'prospectEmailAddress' => $prospectEmailAddress], false), $updateProspectEmailAddressRequestData);
```
rebecca-canyon: I believe that all three of the `postJson` methods should have an `assertForbidden` chained on to them.

app-modules/prospect/tests/Tenant/Http/Controllers/Api/V1/Prospects/UpdateProspectControllerTest.php
```diff
@@ -0,0 +1,230 @@
...
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.view-any');
+    Sanctum::actingAs($user, ['api']);
+    patchJson(route('api.v1.prospects.update', ['prospect' => $prospect], false), $updateProspectRequestData);
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('prospect.*.update');
+    Sanctum::actingAs($user, ['api']);
+    patchJson(route('api.v1.prospects.update', ['prospect' => $prospect], false), $updateProspectRequestData);
```
rebecca-canyon: I believe that all three of the `postJson` methods should have an `assertForbidden` chained on to them.

rebecca-canyon (changes requested):

app-modules/prospect/src/Actions/CreateProspect.php
```diff
@@ -0,0 +1,85 @@
...
+            throw_if(
+                ! $status,
+                ValidationException::withMessages(['status' => 'Status does not exist.'])
+            );
+
+            $source = ProspectSource::whereRaw('LOWER(name) = ?', [Str::lower($data->source)])->first();
+            throw_if(
+                ! $source,
+                ValidationException::withMessages(['source' => 'Source does not exist.'])
+            );
```
Orrison: IMO this validation belongs in the controller and the data object should contain an already resolved status Model.

app-modules/prospect/src/Actions/CreateProspect.php
```diff
@@ -0,0 +1,85 @@
...
+
+            $prospect->fill($data->except('emailAddresses', 'status', 'source')->toArray());
+
+            $prospect->status_id = $status->getKey();
+            $prospect->source_id = $source->getKey();
```
Orrison: I would set the status using the relationship not the columns directly.

app-modules/prospect/src/Actions/UpdateProspect.php
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        if ($data->source) {
+            $source = ProspectSource::whereRaw('LOWER(name) = ?', [Str::lower($data->source)])->first();
+            throw_if(
+                ! $source,
+                ValidationException::withMessages(['source' => 'Source does not exist.'])
+            );
+            $prospect->source_id = $source->getKey();
+        }
```
Orrison: Same as for the create, I would do this validation in the controller and have the data object have the models themselves stored in it.

And set the relationships using the relationship methods.

Orrison (changes requested):

app-modules/prospect/src/DataTransferObjects/CreateProspectData.php
```diff
@@ -0,0 +1,66 @@
...
+    public function __construct(
+        public string $firstName,
+        public string $lastName,
+        public string $fullName,
+        public string | Optional | null $preferred,
+        public string | Optional | null $description,
+        public bool | Optional | null $smsOptOut,
+        public bool | Optional | null $emailBounce,
+        public string $status,
+        public string $source,
```
Orrison: I was thinking these would be the actual Models themselves. Is there a particular reason having their IDs be what is stored here was chosen over the Models?

app-modules/prospect/src/Http/Controllers/Api/V1/Prospects/CreateProspectController.php
```diff
@@ -0,0 +1,111 @@
...
+        $status = ProspectStatus::whereRaw('LOWER(name) = ?', [Str::lower($data['status'])])->first();
+        throw_if(! $status, ValidationException::withMessages(['status' => 'Status does not exist.']));
+        $source = ProspectSource::whereRaw('LOWER(name) = ?', [Str::lower($data['source'])])->first();
+        throw_if(! $source, ValidationException::withMessages(['source' => 'Source does not exist.']));
+
+        $data = CreateProspectData::from([
+            ...$data,
+            'status' => $status->getKey(),
+            'source' => $source->getKey(),
+        ]);
```
Orrison: I would actually pass and store the Models themselves in the DTO. Is there any reason not to do so?

app-modules/prospect/src/DataTransferObjects/UpdateProspectData.php
```diff
@@ -0,0 +1,61 @@
...
+    public function __construct(
+        public string | Optional | null $firstName,
+        public string | Optional | null $lastName,
+        public string | Optional | null $fullName,
+        public string | Optional | null $preferred,
+        public string | Optional | null $description,
+        public bool | Optional | null $smsOptOut,
+        public bool | Optional | null $emailBounce,
+        public string | null $status,
+        public string | null $source,
```
Orrison: I am pretty sure we do not allow `status` or `source` on a Prospect to be `null` from a database perspective or in your controller. So I think `null` should be removed here.

I think you may have been thinking of `Optional` which his appropriate here in the update as they could just not be wanting to update this.

app-modules/prospect/src/Http/Controllers/Api/V1/Prospects/ListProspectsController.php
```diff
@@ -0,0 +1,152 @@
...
+    #[QueryParameter('filter[first_name]', description: 'Filter the results where the prospect\'s first name contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[last_name]', description: 'Filter the results where the prospect\'s last name contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[full_name]', description: 'Filter the results where the prospect\'s full name contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[preferred]', description: 'Filter the results where the prospect\'s preferred name contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[sms_opt_out]', description: 'Filter the results where the prospect\'s sms_opt_out matches the provided boolean.', type: 'boolean')]
+    #[QueryParameter('filter[email_bounce]', description: 'Filter the results where the prospect\'s email_bounce matches the provided boolean.', type: 'boolean')]
+    #[QueryParameter('filter[birthdate]', description: 'Filter the results where the prospect\'s birthdate matches the provided date.', type: 'date')]
```
Orrison: This description doesn't seem super accurate. Since it is a dynamic filter, can't you filter based on the date with different operators?

app-modules/prospect/src/Http/Controllers/Api/V1/Prospects/ListProspectsController.php
```diff
@@ -0,0 +1,152 @@
...
+    #[QueryParameter('filter[status]', description: 'Filter the results where the prospect\'s status matches the provided status name.', type: 'string')]
+    #[QueryParameter('filter[source]', description: 'Filter the results where the prospect\'s source matches the provided source name.', type: 'string')]
+    #[QueryParameter('filter[hsgrad]', description: 'Filter the results where the prospect\'s high school graduation year matches the provided integer.', type: 'integer')]
```
Orrison: This description doesn't seem super accurate. Since it is a dynamic filter, can't you filter based on the date with different operators?

app-modules/prospect/src/Http/Controllers/Api/V1/Prospects/UpdateProspectController.php
```diff
@@ -0,0 +1,129 @@
...
+
+        if (isset($data['source']) && ! $source) {
+            throw ValidationException::withMessages(['source' => 'Source does not exist.']);
+        }
+
+        $data = UpdateProspectData::from([
+            ...$data,
+            'status' => $status?->getKey(),
+            'source' => $source?->getKey(),
+        ]);
```
Orrison: I would store the actual model instead of key

app-modules/prospect/src/DataTransferObjects/UpdateProspectData.php
payal-canyon: Hey @Orrison 
We do need to allow null for status and source in the DTO, just like we do for firstName, etc. If we omit null, the type check in the action (e.g., `if ($data->status))` can throw an error when the field is not provided. Using `ProspectStatus | Optional | null` ensures that the field is truly optional and can be safely ignored during partial updates.
Thank You!

app-modules/prospect/src/Http/Controllers/Api/V1/Prospects/ListProspectsController.php
payal-canyon: Hey @Orrison ,
I tested this in Postman — the filter works both with and without the = operator. For example:

With =:
Params: `filter[birthdate] = =2006-01-13`
URL: `?filter[birthdate]==2006-01-13`

Without =:
Params: `filter[birthdate] = 2006-01-13`
URL: `?filter[birthdate]=2006-01-13`

Both cases return the correct match.

Currently, I’ve set the description for birthdate as:
`Filter the results where the prospect's birthdate matches the provided date using comparison operators. Supported: =, <, <=, >, >=. Format: YYYY-MM-DD.`

I’ve kept the = in the description so it's clear that it's supported — otherwise, users might assume it isn’t. Let me know your thoughts on whether we should keep it, remove it, or any other suggestions!

Orrison: Perfect, describing what operators are supported is what we were looking for.

Orrison (approved):

PR #1687, opened 2025-07-08
[ADVAPP-1595]: EngagementCampaignActionJob threw an Exception about the educatable not being the correct class

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1595

### Technical Description

Fixes issues with different global scoping and error catches that could have caused this issue.

No

No

### DIFF
--- a/app-modules/campaign/src/Models/Campaign.php
+++ b/app-modules/campaign/src/Models/Campaign.php
@@ -103,4 +103,15 @@ class Campaign extends BaseModel implements Auditable
     {
         return $this->morphTo();
     }
+
+    protected static function booted(): void
+    {
+        static::addGlobalScope('licensed', function (Builder $builder) {
+            if (! auth()->guard('web')->check()) {
+                return;
+            }
+
+            $builder->whereHas('segment');
+        });
+    }
 }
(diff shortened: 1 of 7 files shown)


app-modules/campaign/src/Models/Campaign.php
danharrin, line 1: Is it intentional that the scope should be removed here?

Orrison: It was yeah. After looking through everything I am not finding a reason for it to exist at all. It seems it was created to NOT render Campaigns where the segment was deleted. But a global scope seems like a strange place to handle this issue.

danharrin: Wasn't the intention that users without a license for the retention CRM could not see campaigns with student segments, and users without a license for the recruitment CRM shouldn't be able to see prospect segments?

danharrin: Because the `whereHas()` will not just check basic existence, it will check existence based on the global scope of the related model (segment), which does license checks

Orrison: Ah, yes. Good point and good catch. Reverted the change.

PR #1688, opened 2025-07-09
[ADAPP-1629]: Introduce external REST API for programs

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1629

### Technical Description

> Introduce external REST API for programs.

> No.

> No.


app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
```diff
@@ -0,0 +1,81 @@
...
+use Illuminate\Http\Resources\Json\JsonResource;
+use Illuminate\Support\Facades\Gate;
+
+class StudentProgramsController
+{
+    /**
+     * @response StudentProgramResource
+     */
+    #[Group('Students')]
+    public function __invoke(Request $request, UpdateStudentProgram $program, Student $student): JsonResource
```
rebecca-canyon: I believe there should be a `#[QueryParameter]` defined here, with a description and examples.

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        Gate::authorize('viewAny', Student::class);
+        Gate::authorize('viewAny', Program::class);
+        Gate::authorize('create', Program::class);
```
rebecca-canyon: I think that the delete authorization is needed here, too, since the transaction will delete what exists, then create from the provided information.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('program.view-any');
+    Sanctum::actingAs($user, ['api']);
+    putJson(route('api.v1.students.programs.put', ['student' => $student], false), $createStudenProgramRequestData);
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('program.create');
+    Sanctum::actingAs($user, ['api']);
+    putJson(route('api.v1.students.programs.put', ['student' => $student], false), $createStudenProgramRequestData);
```
rebecca-canyon: I believe that all of the `postJson` methods here should have an `assertForbidden` chained on to them.

Additionally, it looks like there's a typo: `$createStudenProgramRequestData` should probably be `$createStudentProgramRequestData`.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo(['student.view-any', 'program.view-any', 'program.create']);
+    Sanctum::actingAs($user, ['api']);
+
+    $createStudenProgramRequestData = [
+        'program' => [StudentProgramRequestFactory::new()->create()],
+    ];
+
+    $response = putJson(route('api.v1.students.programs.put', ['student' => $student], false), $createStudenProgramRequestData);
```
rebecca-canyon: It looks like there's a typo: `$createStudenProgramRequestData` should probably be `$createStudentProgramRequestData`.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+    $before?->call($this);
+
+    $createStudenProgramRequestData = [
+        'program' => [
+            StudentProgramRequestFactory::new()->create(),
+        ],
+        ...$requestAttributes,
+    ];
+
+    $response = putJson(route('api.v1.students.programs.put', ['student' => $student], false), $createStudenProgramRequestData);
```
rebecca-canyon: It looks like there's a typo: `$createStudenProgramRequestData` should probably be `$createStudentProgramRequestData`.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+        ],
+        '`program.*.division` is required' => [
+            ['program' => [['division' => str_repeat('a', 256)]]],
+            'program.0.division',
+            'The program.0.division may not be greater than 255 characters.',
+        ],
+        '`program.*.prog_status` is required' => [
+            ['program' => [['prog_status' => str_repeat('a', 256)]]],
+            'program.0.prog_status',
+            'The program.0.prog_status may not be greater than 255 characters.',
```
rebecca-canyon: All three of these cases are titled that the tested field is required, but then actually tests that the field is within the allowable size instead. Both are good to have, so please add cases for the fields being required.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+        ],
+        '`program.*.cum_gpa` must have 0-2 decimal places' => [
+            ['program' => [['cum_gpa' => 'test']]],
+            'program.0.cum_gpa',
+            'The program.0.cum_gpa field must have 0-2 decimal places.',
+        ],
```
rebecca-canyon: This is titled that the `cum_gpa` field should have the correct number of decimal places, but then actually tests it being the wrong type. Both are good to have, so please add the case for having a number with the correct amount of decimal places.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramControllerTest.php
```diff
@@ -0,0 +1,239 @@
...
+        '`program.*.descr` is required' => [
+            ['program' => [['descr' => str_repeat('a', 256)]]],
+            'program.0.descr',
+            'The program.0.descr may not be greater than 255 characters.',
+        ],
+        '`program.*.foi` is required' => [
+            ['program' => [['foi' => str_repeat('a', 256)]]],
+            'program.0.foi',
+            'The program.0.foi may not be greater than 255 characters.',
```
rebecca-canyon: All three of these cases are titled that the tested field is required, but then actually tests that the field is within the allowable size instead. Both are good to have, so please add cases for the fields being required.

rebecca-canyon (changes requested):

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
ankit-canyon: Hi @rebecca-canyon , 

Actually we're passing bulk program here and Gate only authorize single record here as I see in other place.
hence I just put the create program permission here. 

@Orrison let me know your thoughts on it! 

Thanks.

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
ankit-canyon: Hi @rebecca-canyon,

#[QueryParameter] actually on updating data there is no Query Param we're using. We're using only Path Param and Body data.

Orrison: I believe @ankit-canyon is correct. No Query parameter is needed here.

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
Orrison: Hmm... This is a really good question...

We have yet to have a `PUT` like this in the API so there is no precedence for this yet.

I think I would agree with @rebecca-canyon. Because with a `PUT` you are (or at least are able to) creating AND deleting records. That you should also need the delete permission for programs. And honestly probably the update one as well.

Orrison (commented):
@rebecca-canyon please re-request my review once you have reached approval on yours.

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsControllerTest.php
```diff
@@ -0,0 +1,252 @@
...
+it('is gated with proper access control', function () {
+    $student = Student::factory()->create();
+
+    $createStudentProgramRequestData = [
+        'programs' => [StudentProgramRequestFactory::new()->create()],
+    ];
+
+    $user = SystemUser::factory()->create();
+    Sanctum::actingAs($user, ['api']);
+    putJson(route('api.v1.students.programs.put', ['student' => $student], false), $createStudentProgramRequestData);
```
rebecca-canyon: This should also have an `->assertForbidden()` chained on to it.

rebecca-canyon (approved):

app-modules/student-data-model/routes/api.php
```diff
@@ -84,5 +86,7 @@
                     Route::patch('/{studentPhoneNumber}', UpdateStudentPhoneNumberController::class)->name('update');
                     Route::delete('/{studentPhoneNumber}', StudentPhoneNumbersDeleteStudentPhoneNumberController::class)->name('delete');
                 });
+            Route::get('{student}/programs', ListStudentProgramsController::class)->name('programs.index');
+            Route::put('{student}/programs', StudentProgramsController::class)->name('programs.put');
```
Orrison: This could be put into a group so the route and names can be shared like the email addresses and phone numbers routes above.

app-modules/student-data-model/src/DataTransferObjects/StudentProgramRequestData.php
```diff
@@ -0,0 +1,52 @@
...
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\StudentDataModel\DataTransferObjects;
+
+use Spatie\LaravelData\Attributes\DataCollectionOf;
+use Spatie\LaravelData\Data;
+use Spatie\LaravelData\DataCollection;
+
+class StudentProgramRequestData extends Data
```
Orrison: I would maybe call this the plural form of just having multiple of these items.

It may also be not worth while having a while data class just for this collection given there is no other data on it. You could just use a data collection unnamed: https://spatie.be/docs/laravel-data/v4/as-a-data-transfer-object/collections

```suggestion
class StudentProgramsData extends Data
```

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/ListStudentProgramsController.php
```diff
@@ -0,0 +1,128 @@
...
+{
+    /**
+     * @response AnonymousResourceCollection<LengthAwarePaginator<StudentProgramResource>>
+     */
+    #[Group('Students')]
+    #[QueryParameter('filter[acad_career]', description: 'Filter the results where the program\'s Academic Career contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[division]', description: 'Filter the results where the program\'s Division contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[acad_plan]', description: 'Filter the results where the program\'s Academic Plan contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[prog_status]', description: 'Filter the results where the program\'s Program Status contains the provided string.', type: 'string')]
+    #[QueryParameter('filter[cum_gpa]', description: 'Filter the results where the program\'s Cumulative GPA contains the provided number.', type: 'float')]
```
Orrison: We actually probably want dynamic comparison operators here.

People are going to want to search by Students with X CumGPA or higher, etc.

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/ListStudentProgramsController.php
```diff
@@ -0,0 +1,128 @@
...
+        return QueryBuilder::for($student->programs())
+            ->allowedFilters([
+                AllowedFilter::partial('sisid'),
+                AllowedFilter::partial('acad_career'),
+                AllowedFilter::partial('division'),
+                AllowedFilter::callback('acad_plan', function (Builder $query, string $value) {
+                    $query->whereRaw('acad_plan::text ILIKE ?', ["%{$value}%"]);
+                }),
+                AllowedFilter::partial('prog_status'),
+                AllowedFilter::operator('cum_gpa', FilterOperator::DYNAMIC),
```
Orrison, line 100: Oh it looks like this is already a dynamic operator, so it looks like you just need to update the documentation messaging that I commented on above.

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
```diff
@@ -0,0 +1,91 @@
...
+use AdvisingApp\StudentDataModel\DataTransferObjects\StudentProgramRequestData;
+use AdvisingApp\StudentDataModel\Http\Resources\Api\V1\StudentProgramResource;
+use AdvisingApp\StudentDataModel\Models\Program;
+use AdvisingApp\StudentDataModel\Models\Student;
+use Dedoc\Scramble\Attributes\Group;
+use Illuminate\Http\Request;
+use Illuminate\Http\Resources\Json\JsonResource;
+use Illuminate\Support\Facades\Gate;
+
+class StudentProgramsController
```
Orrison: We may want to name this `PutStudentProgramsController`

app-modules/student-data-model/src/Http/Controllers/Api/V1/Students/StudentPrograms/StudentProgramsController.php
```diff
@@ -0,0 +1,91 @@
...
+            'programs.*.foi' => ['sometimes', 'string', 'max:255'],
+            'programs.*.change_dt' => ['sometimes', 'date', 'date_format:Y-m-d H:i:s'],
+            'programs.*.declare_dt' => ['required', 'date', 'date_format:Y-m-d H:i:s'],
+        ]);
+
+        $studentProgramRequestData = StudentProgramRequestData::from([
+            'programs' => $data['programs'],
+        ]);
+
+        $programs = $program->execute($student, $studentProgramRequestData);
```
Orrison: Like mentioned above you may not need an entire data object just for a collection of programs. You could just make a data collection with the `collect` method on the StudentProgramData DTO.

app-modules/student-data-model/src/Actions/UpdateStudentProgram.php
```diff
@@ -0,0 +1,69 @@
...
+use Illuminate\Support\Facades\DB;
+
+class UpdateStudentProgram
+{
+    /**
+     * @return Collection<int, Program>
+     */
+    public function execute(Student $student, StudentProgramRequestData $requestData): Collection
+    {
+        $student->programs()->delete();
```
Orrison: This should be within the database transaction.

If it isn't, if something fails inside the database transaction part then we will error out and end up in a state where the Students programs have been deleted but then not replaced.

app-modules/student-data-model/src/Actions/UpdateStudentProgram.php
```diff
@@ -0,0 +1,69 @@
...
+                $program = new Program();
+                $program->student()->associate($student);
+                $programData = $data;
+                $programData['acad_plan'] = json_encode($data['acad_plan']);
+                $program->fill($programData);
+                $program->save();
+
+                return $program;
+            })->values();
+        });
```
Orrison: Is there a need to create a Collection this way? I am unsure, but does the DataCollection object not have a `map` or `foreach`?

app-modules/student-data-model/src/DataTransferObjects/StudentProgramRequestData.php
ankit-canyon: Hi @Orrison 

Due to new changes and discuss in office hour meet , we're going with the same file name that is StudentProgramData

app-modules/student-data-model/src/Actions/UpdateStudentProgram.php
```diff
@@ -0,0 +1,63 @@
...
+*/
+
+namespace AdvisingApp\StudentDataModel\Actions;
+
+use AdvisingApp\StudentDataModel\DataTransferObjects\StudentProgramData;
+use AdvisingApp\StudentDataModel\Models\Program;
+use AdvisingApp\StudentDataModel\Models\Student;
+use Illuminate\Support\Facades\DB;
+
+class UpdateStudentProgram
```
Orrison: I think it makes more sense to name this action by what it is doing.

```suggestion
class PutStudentPrograms
```

Orrison (changes requested):

PR #1706, opened 2025-07-14
[ADVAPP-1661]: Update Selects to use the non-native Select by default

krlmrr:
### Ticket(s) or GitHub Issue
- https://canyongbs.atlassian.net/browse/ADVAPP-1661

### Technical Description
- Update the FilamentServiceProvider
- Remove the unneeded Select::native(false) calls.

- No

- No


app-modules/basic-needs/src/Filament/Resources/BasicNeedsProgramResource/Pages/ListBasicNeedsPrograms.php
```diff
@@ -107,7 +107,7 @@ public function table(Table $table): Table
                 SelectFilter::make('basic_category_id')
                     ->label('Program Category')
                     ->relationship('basicNeedsCategories', 'name')
-                    ->native(false)
+                    ->native(condition: false)
```
danharrin: Is there a reason for retaining and modifying this line?

Orrison: We probably need a configureUsing for `SelectFilter` just like we did for `Select`. And then we also will want to do that for the other two projects in different PRs as I already merged the others in.

krlmrr:
@Orrison / @danharrin Do we want to go ahead and approve and merge this and I can piggy back off of this for SelectFilters? Or do we create another ticket for SelectFilters?

danharrin:
I think `SelectFilter` should be handled here, otherwise it will look like the ticket isn't completed to product. They don't understand that the components are configured different.

krlmrr:
> I think `SelectFilter` should be handled here, otherwise it will look like the ticket isn't completed to product. They don't understand that the components are configured different.

Fair enough I will get this knocked out today.

danharrin:
Just seen test failures

krlmrr:
> Just seen test failures

Fixed!

danharrin (approved):

PR #1711, opened 2025-07-16
[ADVAPP-1663]: Cleanup AssignmentsFeature

krlmrr:
### Ticket(s) or GitHub Issue
- https://canyongbs.atlassian.net/browse/ADVAPP-1663

### Technical Description
- Cleanup AssignmentsFeature

- No

- No


app-modules/case-management/src/Filament/Resources/CaseTypeResource/Pages/EditCaseTypeAssignments.php
```diff
@@ -69,7 +68,7 @@ public function getRelationManagers(): array
 
     public static function canAccess(array $parameters = []): bool
     {
-        return AssignmentsFeature::active() && parent::canAccess($parameters);
+        return parent::canAccess($parameters);
     }
```
Orrison: Because all you are doing here now is just returning the parent, the entire method override can just be removed.

Orrison (changes requested):

PR #1713, opened 2025-07-17
[ADVAPP-1666]: Refactor smart prompt technology to allow the introduction of more prompts

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1666

### Technical Description

> Refactor smart prompt technology to allow the introduction of more prompts

> No

> No

### DIFF
--- a/app-modules/ai/src/Actions/SendMessage.php
+++ b/app-modules/ai/src/Actions/SendMessage.php
@@ -63,7 +63,23 @@ class SendMessage
         }
 
         $message = new AiMessage();
-        $message->content = ($content instanceof Prompt) ? $content->prompt : $content;
+        $smartPrompt = $content;
+
+        if ($content instanceof Prompt) {
+            $promptText = $content->prompt;
+
+            if ($content->is_smart) {
+                $additionalContent = "Below I will provide you the input content for a prompt with the name {$content->title}, in the category {$content->type->title}, with the description {$content->description}.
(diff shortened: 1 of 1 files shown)


app-modules/ai/src/Actions/SendMessage.php
```diff
@@ -63,7 +63,30 @@ public function __invoke(AiThread $thread, string | Prompt $content, array $file
...
+                The prompt may have variables {{ VARIABLE }} that are needed in order to effectively serve your function. Begin by analyzing the prompt.
+                Begin by introducing yourself as an AI Advisor, and based on the prompt name, category, and description, explain what your purpose is. Then if the prompt has any variables in it, ask the user for that information, one variable at a time, explaining why you need that input from the user. Once all the variables are collected, return a response for the prompt supplied below.
+                Note: If there are no variables, then just return a response for the prompt supplied below.';
+
+                $smartPrompt = $additionalContent . "\n\n" . $promptText;
+            } else {
+                $smartPrompt = $promptText;
+            }
+        }
+        $message->content = $smartPrompt;
```
Orrison: Overall I think this looks pretty good but the variable naming and structure is a bit confusing.

Please take another pass at cleaning this up.

My suggestions would be to not name any variable here like `smartPrompt` especially in a context in which it could NOT be one.

I would just set `$message->content` based on some different conditionals. and for the construction of the message it could be easier to read if constructed with somelike like a HEREDOC.

Orrison (changes requested):

Orrison (approved):

PR #1717, opened 2025-07-18
[ADVAPP-1606]: Workflows UI

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1606

### Technical Description

Create user interface for workflows with forms.

No

WorkflowFeature; created in [ADVAPP-1606: Create models for Workflows](https://github.com/canyongbs/advisingapp/pull/1657).


app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+
+class ManageFormWorkflows extends ManageRelatedRecords
+{
+    protected static string $resource = FormResource::class;
+
+    protected static string $relationship = 'workflows';
+
+    public static function canAccess(array $arguments = []): bool
+    {
+        return WorkflowFeature::active();
```
Orrison: Also add `&& parent::canAccess()`

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+            ->recordUrl(fn (Workflow $record) => WorkflowResource::getUrl('edit', [$record]));
+    }
+
+    /**
+     * @return array<Action>
+     */
+    public function getHeaderActions(): array
+    {
+        return [
+            $this->createWorkflowAction(),
```
Orrison: Unneeded to have this is a seperate method in this class

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+        ];
+    }
+
+    public function createWorkflowAction(): Action
+    {
+        return Action::make('create')
+            ->label('Create New Workflow')
+            ->action(function () {
+                $workflowTrigger = WorkflowTrigger::create([
+                    'type' => 'Time Based',
```
Orrison: `type` here and throughout should be an ENUM. This type actually probably should be `event_based` (a campaign would be `time_based`

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+                    'related_type' => Form::class,
+                    'related_id' => $this->getOwnerRecord()->getKey(),
+                    'created_by_type' => User::class,
+                    'created_by_id' => auth()->user()->getKey(),
+                ]);
```
Orrison: Turn the `WorkflowTrigger::create` into `WorkflowTrigger::make` then set the related and created by relationships using the relationship methods then save the workflow

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+
+                $workflow = Workflow::create([
+                    'workflow_trigger_id' => $workflowTrigger->getKey(),
+                    'name' => 'Form Workflow',
+                    'is_enabled' => false,
+                ]);
```
Orrison: Same as above with the relationship between the two

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,118 @@
...
+
+                redirect(WorkflowResource::getUrl('edit', [$workflow]));
```
Orrison: I would surround all of this in a single DB transaction

Orrison (changes requested):

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,102 @@
...
+        $workflowTriggerId = WorkflowTrigger::whereRelatedType(Form::class)
+            ->whereRelatedId($event->submission->getKey())
+            ->first()
+            ->getKey();
+
+        $steps = Workflow::whereWorkflowTriggerId($workflowTriggerId)
+            ->whereNotNull('deleted_at')
+            ->whereIsEnabled()
+            ->first()
+            ->workflowSteps();
```
Orrison: Use the `workflowTriggers` relationship you have on `Form` to get the `WorkflowTrigger` and then from the `WorkflowTrigger` get he `Workflow` to do what you are need here and below.

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,102 @@
...
+            assert($step->details instanceof WorkflowDetails);
+
+            if (is_null($step->previous_step_id)) {
+                WorkflowRun::create([
+                    'started_at' => now(),
+                    'workflow_trigger_id' => $workflowTriggerId->getKey(),
+                    'related_type' => $event->submission->author_type,
+                    'related_id' => $event->submission->author_id,
+                ]);
+            }
```
Orrison: Now and after refactor you should be able to create the workflow run here outside of the step loop

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,102 @@
...
+            ->getKey();
+
+        $steps = Workflow::whereWorkflowTriggerId($workflowTriggerId)
+            ->whereNotNull('deleted_at')
+            ->whereIsEnabled()
+            ->first()
+            ->workflowSteps();
+
+        $steps->each(function (WorkflowStep $step) use ($event, $workflowTriggerId) {
+            assert($step->details instanceof WorkflowDetails);
```
Orrison: `details` is not an existing relationship on `WorkflowStep`

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,127 @@
...
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->recordTitleAttribute('id')
+            ->columns([
+                TextColumn::make('name'),
+                IconColumn::make('is_enabled')
+                    ->label('Enabled')
+                    ->state(fn (Workflow $record) => $record->is_enabled)
```
Orrison: Is this `state` call needed? Because the field is named `is_enabled` it should putt a bool attribute from the record already.

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,109 @@
...
+        $workflowTriggerId = WorkflowTrigger::whereRelatedType(Form::class)
+            ->where('related_id', $event->submission->getKey())
+            ->first()
+            ->getKey();
+
+        $steps = Workflow::whereWorkflowTriggerId($workflowTriggerId)
+            ->whereNotNull('deleted_at')
+            ->where('is_enabled', true)
+            ->first()
+            ->workflowSteps();
```
Orrison: Like we discussed, I would recommend never using Laravel magic `where` methods as they make it super hard to track down if the column or anything about it changes.

In fact, do do what you are doing here I think you can go from the `Form`'s relationship to the workflow trigger then to the workflow to make this all easier.

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,109 @@
...
+                    'started_at' => now(),
+                    'related_type' => $event->submission->author_type,
+                    'related_id' => $event->submission->author_id,
+                ]);
+
+                $workflowRun->workflowTrigger()->associate($workflowTriggerId);
+                $workflowRun->related()->associate($event->submission->author);
+
+                $workflowRun->save();
+            }
```
Orrison: I would handle the creation of this outside of the steps loop.

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,109 @@
...
+            ->workflowSteps();
+
+        $steps->each(function (WorkflowStep $step) use ($event, $workflowTriggerId) {
+            assert($step->details instanceof WorkflowDetails);
+
+            if (is_null($step->previous_step_id)) {
+                $workflowRun = new WorkflowRun([
+                    'started_at' => now(),
+                    'related_type' => $event->submission->author_type,
+                    'related_id' => $event->submission->author_id,
```
Orrison: This is no longer needed since you are defining the relationship below

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,109 @@
...
+                $workflowRun->related()->associate($event->submission->author);
+
+                $workflowRun->save();
+            }
+
+            $workflowRunStep = new WorkflowRunStep([
+                'execute_at' => $this->getStepScheduledAt($step, $event),
+            ]);
+
+            $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowTriggerId)->get();
```
Orrison: Magic where

app-modules/form/src/Models/Form.php
```diff
@@ -105,4 +111,21 @@ public function emailAutoReply(): HasOne
...
     }
+
+    /**
+     * @return HasMany<WorkflowTrigger, $this>
+     */
+    public function workflowTriggers(): HasMany
+    {
+        return $this->hasMany(WorkflowTrigger::class, 'related_id')
+            ->where('related_type', Form::class);
+    }
```
Orrison: Is this not a polymorphic relationship on the other end? And so should be here as well I think?

app-modules/workflow/database/migrations/.gitkeep
Orrison, line 1: This file can be deleted since you have migrations

app-modules/workflow/database/migrations/2025_07_09_223432_create_workflow_run_steps_table.php
```diff
@@ -0,0 +1,66 @@
...
+        Schema::create('workflow_run_steps', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->dateTime('execute_at');
+            $table->dateTime('dispatched_at')->nullable();
+            $table->dateTime('succeeded_at')->nullable();
+            $table->dateTime('last_failed_at')->nullable();
+            $table->foreignUuid('workflow_run_id')->constrained('workflow_runs');
+            $table->string('details_id');
+            $table->string('details_type');
```
Orrison: You can use the normal uuid morph helper here since we know the detail will have a UUID primary key. We do this string thing when it COULD be a Student, which has a string identifier.

app-modules/workflow/database/migrations/2025_07_23_224210_drop_workflow_step_id__and_make_appropriate_columns_nullable_from_workflow_case_details_table.php
```diff
@@ -0,0 +1,59 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('workflow_case_details', function (Blueprint $table) {
+            $table->dropConstrainedForeignId('workflow_step_id');
```
Orrison, line 45: I think we still want to keep this, right?

It is not very common but this pattern where you leave the relationship here on the detail but then also have the current details id one the step itself allows for two thing.

The Workflow Step has details of exactly what it's "current" details is for us in creating new runs.

The detail has the ability to "remember" what workflow step it is/was related to.

Now in your code right now, we really only care about what the current one is. But if we ever wanted to trace back that could be useful.
I could have also forgotten a conversation we had before lol

app-modules/workflow/src/Filament/Resources/WorkflowResource/Pages/ListWorkflows.php
```diff
@@ -0,0 +1,53 @@
...
+use Filament\Resources\Pages\ListRecords;
+
+class ListWorkflows extends ListRecords
+{
+    protected static string $resource = WorkflowResource::class;
+
+    protected function getHeaderActions(): array
+    {
+        return [
+            Actions\CreateAction::make(),
```
Orrison: Let's fully import this class here.

app-modules/workflow/src/Filament/Resources/WorkflowResource/RelationManagers/WorkflowActionsRelationManager.php
```diff
@@ -0,0 +1,208 @@
...
+use Filament\Tables\Actions\DeleteAction;
+use Filament\Tables\Actions\DeleteBulkAction;
+use Filament\Tables\Actions\EditAction;
+use Filament\Tables\Columns\TextColumn;
+use Filament\Tables\Table;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+use Throwable;
+
+class WorkflowActionsRelationManager extends RelationManager
```
Orrison: Would this make more sense to be called `WorkflowStepsRelationManager`?

app-modules/workflow/src/Filament/Resources/WorkflowResource/RelationManagers/WorkflowActionsRelationManager.php
```diff
@@ -0,0 +1,208 @@
...
+                                        $action = $this->createWorkflowDetails($block, $data);
+
+                                        $delayMinutes = ($data['days'] * 24 * 60) + ($data['hours'] * 60) + $data['minutes'];
+
+                                        $workflowStep = new WorkflowStep([
+                                            'delay_minutes' => $delayMinutes,
+                                            'workflow_id' => $record->getKey(),
+                                            'current_details_type' => $action->getType(),
+                                            'current_details_id' => $action->getKey(),
+                                            'previous_step_id' => $previousStepId,
```
Orrison: I would save these using the relationships

app-modules/workflow/src/Jobs/ExecuteWorkflowActionOnEducatableJob.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace AdvisingApp\Workflow\Jobs;
+
+use AdvisingApp\Campaign\Jobs\Middleware\FailIfBatchCancelled;
+use AdvisingApp\Workflow\Models\WorkflowRunStep;
+use Illuminate\Bus\Batchable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Queue\Queueable;
+
+abstract class ExecuteWorkflowActionOnEducatableJob implements ShouldQueue
```
Orrison: It may make sense to change the name here since we don't have a workflow action Educatable.

```suggestion
abstract class ExecuteWorkflowActionJob implements ShouldQueue
```

app-modules/workflow/src/Models/.gitkeep
Orrison, line 1: This can be removed

app-modules/workflow/src/Jobs/CaseWorkflowActionJob.php
```diff
@@ -0,0 +1,107 @@
...
+                throw new Exception('The Case Management addon is not enabled.');
+            }
+
+            DB::beginTransaction();
+
+            $educatable = $this->workflowRunStep->workflowRun->related;
+
+            assert($educatable instanceof Educatable);
+
+            $details = WorkflowCaseDetails::whereId($this->workflowRunStep->details_id)->first();
```
Orrison: Don't use magic wheres here. You should get the details through the `WorkflowRunStep`s relationship to the detail

app-modules/workflow/src/Models/WorkflowRunStep.php
```diff
@@ -0,0 +1,91 @@
...
+use Illuminate\Database\Eloquent\SoftDeletes;
+use OwenIt\Auditing\Contracts\Auditable;
+
+/**
+ * @mixin IdeHelperWorkflowRunStep
+ */
+class WorkflowRunStep extends BaseModel implements Auditable
+{
+    use SoftDeletes;
+    use AuditableTrait;
```
Orrison: This Model does not need to be `Auditable` since it is created via automated process.

app-modules/workflow/src/Models/WorkflowRunStep.php
```diff
@@ -0,0 +1,91 @@
...
+        'details_id',
+        'details_type',
+    ];
+
+    protected $casts = [
+        'execute_at' => 'datetime',
+        'dispatched_at' => 'datetime',
+        'succeeded_at' => 'datetime',
+        'last_failed_at' => 'datetime',
+        'details_type' => WorkflowActionType::class,
```
Orrison: Let's chat about this next week in person with the team. This leads me to believe we need to handle everything in an abstract details class rather than an ENUM like this.

app-modules/workflow/src/Models/Workflow.php
```diff
@@ -53,10 +53,27 @@ class Workflow extends BaseModel implements Auditable
...
+    public function hasBeenExecuted(): bool
+    {
+        return $this->workflowSteps->contains(function (WorkflowStep $step) {
+            $details = $step->currentDetails;
+
+            assert($details instanceof WorkflowDetails);
+
+            return $details->hasBeenExecuted();
+        });
+    }
```
Orrison: This seems strange to be going through the details at all. Especially when, looking at the details class, there you come back up through the step to here and over to the trigger.

Most likely it makes sense for this just to immediately check it's related trigger and see if there are any WorkflowRuns for that.

app-modules/workflow/src/Models/WorkflowDetails.php
```diff
@@ -0,0 +1,55 @@
...
+
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Workflow\Models;
+
+use App\Models\BaseModel;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+abstract class WorkflowDetails extends BaseModel
```
Orrison, line 56: We can discuss this as a team. But as I mentioned elsewhere. Looking at this all I am thinking we may be able to get rid of the ENUM and move all of it's information to this abstract. I am not sure though.

app-modules/workflow/src/Models/WorkflowCaseDetails.php
```diff
@@ -63,11 +61,25 @@ class WorkflowCaseDetails extends BaseModel implements Auditable, WorkflowAction
...
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Case)->where('details_id', $this->id)->first()->dispatched_at);
     }
```
Orrison: See my other comment in the workflow. I don't think this should happen through / here.

app-modules/workflow/src/Models/WorkflowCareTeamDetails.php
```diff
@@ -64,11 +62,25 @@ class WorkflowCareTeamDetails extends BaseModel implements Auditable, WorkflowAc
...
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::CareTeam)->where('details_id', $this->id)->first()->dispatched_at);
     }
```
Orrison: Same as others, I think this should be removed.

app-modules/workflow/src/Models/WorkflowEngagementEmailDetails.php
```diff
@@ -67,11 +65,25 @@ class WorkflowEngagementEmailDetails extends BaseModel implements Auditable, Wor
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::EngagementEmail)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowEngagementSmsDetails.php
```diff
@@ -65,11 +63,25 @@ class WorkflowEngagementSmsDetails extends BaseModel implements Auditable, Workf
...
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::EngagementSms)->where('details_id', $this->id)->first()->dispatched_at);
     }
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowEventDetails.php
```diff
@@ -58,11 +56,25 @@ class WorkflowEventDetails extends BaseModel implements Auditable, WorkflowActio
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Event)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowInteractionDetails.php
```diff
@@ -73,11 +71,25 @@ class WorkflowInteractionDetails extends BaseModel implements Auditable, Workflo
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Interaction)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowProactiveAlertDetails.php
```diff
@@ -66,11 +64,25 @@ class WorkflowProactiveAlertDetails extends BaseModel implements Auditable, Work
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::ProactiveAlert)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowDetails.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    abstract public function getType(): string;
+
+    abstract public function hasBeenExecuted(): bool;
```
Orrison: I think we should remove this and check this stuff directly in the workflow as I described there.

app-modules/workflow/src/Models/WorkflowRun.php
```diff
@@ -0,0 +1,85 @@
...
+use Illuminate\Database\Eloquent\SoftDeletes;
+use OwenIt\Auditing\Contracts\Auditable;
+
+/**
+ * @mixin IdeHelperWorkflowRun
+ */
+class WorkflowRun extends BaseModel implements Auditable
+{
+    use SoftDeletes;
+    use AuditableTrait;
```
Orrison: This doesn't need to be Auditable as it is created and only interacted with via automation

app-modules/workflow/src/Models/WorkflowRunStepRelated.php
```diff
@@ -0,0 +1,78 @@
...
+use Illuminate\Database\Eloquent\SoftDeletes;
+use OwenIt\Auditing\Contracts\Auditable;
+
+/**
+ * @mixin IdeHelperWorkflowRunStepRelated
+ */
+class WorkflowRunStepRelated extends BaseModel implements Auditable
+{
+    use SoftDeletes;
+    use AuditableTrait;
```
Orrison: This also does not need to be Auditable

app-modules/workflow/src/Models/WorkflowSubscriptionDetails.php
```diff
@@ -64,11 +62,25 @@ class WorkflowSubscriptionDetails extends BaseModel implements Auditable, Workfl
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Subscription)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowTagsDetails.php
```diff
@@ -64,11 +62,25 @@ class WorkflowTagsDetails extends BaseModel implements Auditable, WorkflowAction
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Tags)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Models/WorkflowTaskDetails.php
```diff
@@ -64,11 +62,25 @@ class WorkflowTaskDetails extends BaseModel implements Auditable, WorkflowAction
...
+
+        $workflowRun = WorkflowRun::whereWorkflowTriggerId($workflowStep->workflow->workflowTrigger->getKey())->first();
+
+        if (is_null($workflowRun)) {
+            return false;
+        }
+
+        $workflowRunSteps = $workflowRun->workflowRunSteps;
+
+        return ! is_null($workflowRunSteps->where('details_type', WorkflowActionType::Task)->where('details_id', $this->id)->first()->dispatched_at);
```
Orrison: Same as others I think this should be removed.

app-modules/workflow/src/Jobs/ExecuteWorkflowActionJob.php
```diff
@@ -0,0 +1,78 @@
...
+
+class ExecuteWorkflowActionJob implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+
+    public function handle(): void
+    {
+        $steps = WorkflowRunStep::where('execute_at', '<', now())->whereNull('dispatched_at');
```
Orrison:
```suggestion
        $steps = WorkflowRunStep::query()->where('execute_at', '<=', now())->whereNull('dispatched_at');
```

app-modules/workflow/src/Jobs/ExecuteWorkflowAction.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace AdvisingApp\Workflow\Jobs;
+
+use AdvisingApp\Workflow\Models\WorkflowRunStep;
+use Exception;
+use Illuminate\Bus\Batchable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Queue\Queueable;
+
+class ExecuteWorkflowAction implements ShouldQueue
```
Orrison: This job is not needed as all it is doing is dispatching the actual job that is supposed to be being executed. So I would remove this and just call the equivalent of `step->details_type->getActionExecutableJob($this->step)` in the other job

app-modules/workflow/src/Jobs/ExecuteWorkflowActionJob.php
```diff
@@ -0,0 +1,78 @@
...
+                        $step->succeeded_at = now();
+                        $step->save();
+                    });
+            } catch (Throwable $exception) {
+                $step->last_failed_at = now();
+                $step->save();
+
+                throw $exception;
+            }
+        });
```
Orrison: Looking at this now, I think we shouldn't us a batch for this.

CampaignActions use a batch because a single action there executes across multiple records. Where as this executes just one thing.

So I would change this to just dispatching the correct job for the step details type, and setting `dispatched_at` of the step.

Then the workflow job should handle setting `succeeded_at` if it completes or `last_failed_at` if it fails.

Orrison (changes requested):

app-modules/workflow/database/migrations/2025_07_23_224210_drop_workflow_step_id__and_make_appropriate_columns_nullable_from_workflow_case_details_table.php
rebecca-canyon: I see what you're saying about the utility of it, but I truly wasn't able to figure out how to be able to create either a Workflow Step or a Workflow Details with them both having non nullable foreign keys to each other.

I think the concept of recursive relationships was mentioned? Which unfortunately didn't seem to be what I'm doing here. And all I could find on circular relationships was "don't do them", so I figured that pivoting away from that might be wise.

app-modules/workflow/src/Jobs/ExecuteWorkflowActionOnEducatableJob.php
rebecca-canyon: I had another file that was `ExecuteWorkflowActionJob`. As I couldn't think of a better name for this file, I renamed the previous `ExecuteWorkflowActionJob` to `ExecuteWorkflowActionStepsJob` and this one to `ExecuteWorkflowActionJob`. Figured I ought to mention because I'm sure the git diff won't be super clear about that.

app-modules/workflow/src/Models/WorkflowCaseDetails.php
rebecca-canyon: Actually, on further thought, I dont think the whole `hasBeenExecuted` thing should be happening at all.

Since I was using Campaigns as something of a guideline here, and because I tunnel visioned on the experience of an individual, I thought it would make sense to bring the concept over and use whether or not the workflow has happened to gate whether or not it can be edited/deleted.
But since these are triggered on an individual based off of when they submit a form, and not just all at once, and there's nothing currently to see an individual's progress through a workflow, I don't think it actually makes sense here.

So I've just removed the whole concept.

app-modules/workflow/src/Models/WorkflowRunStep.php
Orrison: @rebecca-canyon as we discussed in person, let's move forward with removing the ENUM and moving it's functionality into the abstract `WorkflowDetails` class

Orrison (changes requested):
Added comment about, let's refactor the ENUM into the `WorkflowDetails` abstract class

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
```diff
@@ -0,0 +1,126 @@
...
+                        ]);
+
+                        $workflowTrigger->related()->associate($this->getOwnerRecord());
+                        $workflowTrigger->createdBy()->associate(auth()->user());
+
+                        $workflowTrigger->save();
+
+                        $workflow = new Workflow([
+                            'name' => 'Form Workflow',
+                            'is_enabled' => false,
```
Orrison, line 110: Is there any way that this get's enabled right now?

app-modules/form/src/Listeners/TriggerFormSubmissionWorkflows.php
```diff
@@ -0,0 +1,104 @@
...
+use AdvisingApp\Workflow\Models\Workflow;
+use AdvisingApp\Workflow\Models\WorkflowDetails;
+use AdvisingApp\Workflow\Models\WorkflowRun;
+use AdvisingApp\Workflow\Models\WorkflowRunStep;
+use AdvisingApp\Workflow\Models\WorkflowStep;
+use AdvisingApp\Workflow\Models\WorkflowTrigger;
+use Carbon\Carbon;
+use Illuminate\Contracts\Queue\ShouldQueue;
+
+class TriggerFormSubmissionWorkflows implements ShouldQueue
```
Orrison: I would change this here to `ShouldQueueAfterCommit` just in case we run into a DB transaction race condition.

```suggestion
class TriggerFormSubmissionWorkflows implements ShouldQueueAfterCommit
```

app-modules/workflow/src/Models/WorkflowDetails.php
```diff
@@ -0,0 +1,108 @@
...
+use AdvisingApp\Workflow\Filament\Blocks\WorkflowActionBlock;
+use AdvisingApp\Workflow\Jobs\ExecuteWorkflowActionJob;
+use App\Models\BaseModel;
+use App\Settings\LicenseSettings;
+use Filament\Forms\Components\Field;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+abstract class WorkflowDetails extends BaseModel
+{
+    abstract public function getType(): string;
```
Orrison: When we get a chance I would like to look into together what `getType` is used for

app-modules/workflow/src/Models/WorkflowRunStep.php
```diff
@@ -0,0 +1,87 @@
...
+        'details_id',
+        'details_type',
+    ];
+
+    protected $casts = [
+        'execute_at' => 'datetime',
+        'dispatched_at' => 'datetime',
+        'succeeded_at' => 'datetime',
+        'last_failed_at' => 'datetime',
+        'details_type' => WorkflowDetails::class,
```
Orrison: I don't think this is needed or even will work

Orrison (changes requested):

app-modules/form/src/Filament/Resources/FormResource/Pages/ManageFormWorkflows.php
rebecca-canyon: Yes; through the WorkflowResource.

app-modules/workflow/src/Models/WorkflowDetails.php
rebecca-canyon: A quick search shows me that it's actually not used- so I'm just going to remove it.

app-modules/workflow/src/Models/WorkflowRunStep.php
rebecca-canyon: Now that you point it out, I think you're right; I was probably thinking about the inheritance, and not that this would just resolve to a string / the class name.

Orrison (approved):

PR #1726, opened 2025-07-21
[ADVAPP-1697]: Implement smart prompts display in the institutional advisor on the new chat screen

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1697

### Technical Description

> Implement smart prompts display in the institutional advisor on the new chat screen

> No

> No

### DIFF
--- a/app-modules/ai/resources/views/components/assistant.blade.php
+++ b/app-modules/ai/resources/views/components/assistant.blade.php
@@ -536,11 +536,17 @@
                     @endif
                 </div>
 
-                <div
-                    class="flex flex-1 flex-col-reverse overflow-y-scroll rounded-xl border border-gray-950/5 text-sm shadow-sm dark:border-white/10 dark:bg-gray-800">
+                @php
+                    $isInstitutionalAdvisor = $this->thread->assistant->name === 'Institutional Advisor';
+                    $hasMessages = count($this->thread->messages) > 0;
+                    $shouldReverse = $isInstitutionalAdvisor && $hasMessages;
+                @endphp
 
-                    @if (count($this->thread->messages) == 0 && $this->thread->assistant->name === 'Institutional Advisor')
-                        @livewire('prompttabs', ['thread' => $this->thread], key('prompt-tabs-' . $this->thread->assistant->id))
+                <div
+                    class="{{ $shouldReverse ? 'flex-col-reverse' : 'flex-col' }} flex flex-1 overflow-y-scroll rounded-xl border border-gray-950/5 text-sm shadow-sm dark:border-white/10 dark:bg-gray-800">
+
+                    @if (!$hasMessages && $isInstitutionalAdvisor)
+                        @livewire('promptlibrarytabs', ['thread' => $this->thread, 'promptLibraryMode' => true], key('prompt-library-tabs-' . $this->thread->assistant->id))
                     @endif
 
                     <div
--- a/app-modules/ai/resources/views/components/assistant.blade.php
+++ b/app-modules/ai/resources/views/components/assistant.blade.php
@@ -538,6 +538,13 @@
 
                 <div
                     class="flex flex-1 flex-col-reverse overflow-y-scroll rounded-xl border border-gray-950/5 text-sm shadow-sm dark:border-white/10 dark:bg-gray-800">
+                    {{-- <div
+                    class="flex flex-col overflow-y-auto rounded-xl border border-gray-950/5 text-sm shadow-sm dark:border-white/10 dark:bg-gray-800"> --}}
+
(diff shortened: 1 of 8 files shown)


app-modules/ai/resources/views/components/assistant.blade.php
```diff
@@ -536,8 +536,19 @@ class="col-span-1 flex flex-col gap-2 overflow-hidden pt-3 lg:col-span-2 lg:pt-6
                     @endif
                 </div>
 
+                @php
+                    $isInstitutionalAdvisor = $this->thread->assistant->name === 'Institutional Advisor';
```
Orrison: Instead of checking for the name here you should check if the assistant `is_default`. It may not always have the same name but what you are trying to do here is see if it is the default advisor.

app-modules/ai/src/Enums/AiPromptTabs.php
```diff
@@ -0,0 +1,57 @@
...
+
+namespace AdvisingApp\Ai\Enums;
+
+use Filament\Support\Contracts\HasLabel;
+
+enum AiPromptTabs: string implements HasLabel
+{
+    case Newest = 'newest';
+
+    case MostLoved = 'most';
```
Orrison:
```suggestion
    case MostLoved = 'most_loved';
```

app-modules/assistant/src/Filament/Pages/InstitutionalAdvisor.php
```diff
@@ -56,6 +56,8 @@ class InstitutionalAdvisor extends Page
 
     public const APPLICATION = AiAssistantApplication::PersonalAssistant;
 
+    public mixed $promptLibraryMode = null;
```
Orrison: Could it make more sense for this to be type hinted to a `bool` defaulting to `false` and have it be called something more specific to what it does? Like `isSmartPromptsTypePreselected`?

app/Livewire/PromptLibraryTabs.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    use InteractsWithActions;
+    use InteractsWithForms;
+    use CanManagePromptLibrary;
+
+    public string $activeTab = AiPromptTabs::Newest->value;
+
+    public ?AiThread $thread;
+
+    public mixed $promptLibraryMode = null;
```
Orrison: Could it make more sense for this to be type hinted to a `bool` defaulting to `false` and have it be called something more specific to what it does? Like `isSmartPromptsTypePreselected`?

app/Livewire/PromptLibraryTabs.php
```diff
@@ -0,0 +1,75 @@
...
+
+    public function render(): View
+    {
+        $prompts = Prompt::query()
+            ->withCount('upvotes', 'uses')
+            ->when($this->activeTab === AiPromptTabs::Newest->value, fn (Builder $query) => $query->latest())
+            ->when($this->activeTab === AiPromptTabs::MostLoved->value, fn (Builder $query) => $query->orderByDesc('upvotes_count'))
+            ->when($this->activeTab === AiPromptTabs::MostViewed->value, fn (Builder $query) => $query->orderByDesc('uses_count'))
```
Orrison: Both of these queries may never trigger their empty state I don't think. I think you will need to keep the order you are doing but also add conditions to the query so that you are scoping to only prompts with upvotes and only prompts with uses, respectively.

app/Livewire/PromptLibraryTabs.php
```diff
@@ -0,0 +1,75 @@
...
+    public string $activeTab = AiPromptTabs::Newest->value;
+
+    public ?AiThread $thread;
+
+    public mixed $promptLibraryMode = null;
+
+    public function render(): View
+    {
+        $prompts = Prompt::query()
+            ->withCount('upvotes', 'uses')
```
Orrison, line 67: This is only supposed to display smart prompts so you should add that to this query as well.

Orrison (changes requested):

app-modules/assistant/src/Filament/Pages/InstitutionalAdvisor.php
payal-canyon: Hi @Orrison ,
I had considered defaulting it to `false`, but during testing, I noticed that clicking the `Prompt Library` button next to the `Send` button would automatically select the `Custom Prompt` type — which wasn’t ideal. That’s why I initially set it to `null` to avoid preselecting anything.

I’ve now updated it to default to `false` as suggested. If we decide to go back to null to prevent any preselection, I can easily update it — just let me know.
Thank You!

Orrison (approved):

PR #1729, opened 2025-07-21
[ADVAPP-1660]: Enhance admissions to automatically generate prospects who might not yet already exist

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1660

### Technical Description

> 
- Introduced a new migration to add 'generate_prospects' column to applications table.
- Implemented 'registerProspect' method in ApplicationWidgetController to handle prospect registration.
- Updated API routes to include prospect registration endpoint.
- Enhanced Application model to support 'generate_prospects' attribute.
- Added UI components in App.vue for prospect registration form.
- Created OnlineAdmissionGenerateProspect feature flag for conditional prospect generation.

> No.

> Yes.
> OnlineAdmissionGenerateProspect 


app-modules/application/database/migrations/2025_07_21_193542_add_generate_prospects_to_applications_table.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('applications', function (Blueprint $table) {
+            $table->boolean('generate_prospects')->default(false);
```
Orrison: There is precedent in all of our apps that boolean columns like this start with something like `is`, `allow`, `should`, etc.

Given that pattern and the context of the feature, I am thinking we should change this to `should_generate_prospects`

app-modules/application/src/Models/Application.php
```diff
@@ -54,6 +54,7 @@ class Application extends Submissible
         'primary_color',
         'rounding',
         'content',
+        'generate_prospects',
```
Orrison: This will need to change given the above.

app-modules/application/src/Models/Application.php
```diff
@@ -62,6 +63,7 @@ class Application extends Submissible
         'allowed_domains' => 'array',
         'is_wizard' => 'boolean',
         'rounding' => Rounding::class,
+        'generate_prospects' => 'boolean',
```
Orrison: This will need to change given the above.

app-modules/application/src/Filament/Resources/ApplicationResource/Pages/Concerns/HasSharedFormConfiguration.php
```diff
@@ -94,6 +95,11 @@ public function fields(): array
                 ->label('Multi-step form')
                 ->live()
                 ->disabled(fn (?Application $record) => $record?->submissions()->exists()),
+            Toggle::make('generate_prospects')
```
Orrison: This will need to change given the above.

app-modules/application/src/Http/Controllers/ApplicationWidgetController.php
```diff
@@ -86,17 +93,30 @@ public function requestAuthentication(Request $request, ResolveSubmissionAuthorF
         $author = $resolveSubmissionAuthorFromEmail($data['email']);
 
         if (! $author) {
-            throw ValidationException::withMessages([
-                'email' => 'A student with that email address could not be found. Please contact your system administrator.',
-            ]);
+            if (! $application->generate_prospects) {
```
Orrison: This will need to change given the above.

widgets/application/src/App.vue
```diff
@@ -294,6 +345,113 @@ async function authenticate(applicationData, node) {
...
+                            </div>
+                        </div>
+                        <div class="flex flex-wrap -mx-3 mb-6">
+                            <div class="w-full md:w-1/2 px-3 mb-6 md:mb-0">
+                                <FormKit
+                                    type="tel"
+                                    label="Mobile"
+                                    name="mobile"
+                                    placeholder="xxx-xxx-xxxx"
+                                    validation="required|length:0,255"
```
Orrison: I don't think we need this to be required as I don't think Prospects need a phone number. So if, so let's remove the required here.

Is it possible to add e164 validation here easily?

widgets/application/src/App.vue
```diff
@@ -294,6 +345,113 @@ async function authenticate(applicationData, node) {
...
+                            <div class="w-full md:w-1/2 px-3">
+                                <FormKit
+                                    type="text"
+                                    label="Postal"
+                                    name="postal"
+                                    validation="required|length:0,255"
+                                    validation-visibility="submit"
+                                />
+                            </div>
+                        </div>
```
Orrison: This address data is not required either.

Orrison (changes requested):

widgets/application/src/App.vue
ankit-canyon: Added !

widgets/application/src/App.vue
ankit-canyon: Hi @Orrison 

as we discussed, I compare it with prospect creation and added the same required fields here as well.

Thank you.

Orrison (approved):

PR #1742, opened 2025-07-24
[ADVAPP-1728]: Bug Fix: Attempts to Send Alert to a Deleted User

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1728

### Technical Description
- Add a scope to only pull back Subscriptions if they have a user.

- No

- No

### DIFF
--- a/app-modules/alert/tests/Tenant/AlertCreateTest.php
+++ b/app-modules/alert/tests/Tenant/AlertCreateTest.php
@@ -79,3 +79,25 @@ it('dispatches the proper notifications to subscribers on created', function () 
     Notification::assertSentTo($users, AlertCreatedNotification::class);
     Notification::assertSentTimes(AlertCreatedNotification::class, $student->subscriptions()->count());
 });
+
+it('only notifies active users of alerts', function () {
+    Notification::fake();
+
+    $deletedUser = User::factory()->licensed(LicenseType::cases())->create();
+
+    /** @var Student $student */
+    $student = Student::factory()->create();
+
+    $student->subscriptions()->create([
+        'user_id' => $deletedUser->id,
+    ]);
+
+    $deletedUser->delete();
+
+    Alert::factory()->create([
+        'concern_id' => $student->sisid,
+        'concern_type' => Student::class,
+    ]);
+
+    Notification::assertNotSentTo($deletedUser, AlertCreatedNotification::class);
+});
--- a/app-modules/alert/tests/Tenant/AlertCreateTest.php
+++ b/app-modules/alert/tests/Tenant/AlertCreateTest.php
@@ -81,6 +81,7 @@ it('dispatches the proper notifications to subscribers on created', function () 
 });
 
 it('only notifies active users of alerts', function () {
+    // @Todo: This tests works but we should change it to use the Listener instead of the Notification
     Notification::fake();
 
     $deletedUser = User::factory()->licensed(LicenseType::cases())->create();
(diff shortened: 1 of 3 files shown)


app-modules/alert/tests/Tenant/AlertCreateTest.php
```diff
@@ -79,3 +79,25 @@
     Notification::assertSentTo($users, AlertCreatedNotification::class);
     Notification::assertSentTimes(AlertCreatedNotification::class, $student->subscriptions()->count());
 });
+
+it('only notifies active users of alerts', function () {
```
Orrison, line 83: Though this does work. I think it would be more appropriate to be in a test for the Listener itself.

app-modules/notification/src/Models/Concerns/HasSubscriptions.php
```diff
@@ -46,6 +46,6 @@ trait HasSubscriptions
      */
     public function subscriptions(): MorphMany
     {
-        return $this->morphMany(Subscription::class, 'subscribable');
+        return $this->morphMany(Subscription::class, 'subscribable')->whereHas('user');
```
Orrison, line 49: @danharrin I am curious your opinion on this. Technically I think this would solve anywhere in the app where we are trying to retrieve subscriptions where the related User got deleted.

But it feels strange to modify the relationship directly rather than deal with this where the issue is happening.

I am unsure.

Orrison (changes requested):

app-modules/notification/src/Models/Concerns/HasSubscriptions.php
danharrin: I think its a good fix to be honest. A subscription doesn't make sense unless a user is involved, so it should never be returned.

My only concern is if we have any logic that automatically unsubscribes users after they are deleted - maybe this would result in the relationship not being able to be used, since the query would not return anything.

But I think this is good to go

Orrison: Yeah we don't have that right now. And if we did I guess we could chain on the relationship `->withTrashed()` or something like that to override it.

I'll add a ticket to the backlog to prune subscriptions where there is no User or Educatable

app-modules/alert/tests/Tenant/AlertCreateTest.php
Orrison: @krlmrr I am good with the other change and have resolved my comment. But I would still like us to move this test to one specific to the Listener.

krlmrr: Will do once I get to Denver.

app-modules/notification/src/Models/Concerns/HasSubscriptions.php
danharrin: I don't think there is an easy way to remove the scope once it's already applied, `withTrashed()` definitely wouldn't, however we could always define another relationship like `subscriptionsWithoutUsers` for this purpose.

Orrison: Yeah, that sounds right. Well since we don't need it now this is fine.

app-modules/alert/tests/Tenant/AlertCreateTest.php
krlmrr: @Orrison The todo has been added.

Orrison: Per discussion in person, we are going to move this to a later iteration.

PR #1743, opened 2025-07-25
[ADVAPP-1724][ADVAPP-1725]: Restrict the view of models if it is not configurable to Global Administrators, Restrict the view of the application dropdown in the custom advisor screens to Global Administrators

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1724
- https://canyongbs.atlassian.net/browse/ADVAPP-1725

### Technical Description

> Restrict the view of models if it is not configurable to Global Administrators and Restrict the view of the application dropdown in the custom advisor screens to Global Administrators.

> No.

> No.

### DIFF
--- a/app-modules/ai/tests/Tenant/Feature/Filament/Resources/AiAssistantResource/EditAiAssistantTest.php
+++ b/app-modules/ai/tests/Tenant/Feature/Filament/Resources/AiAssistantResource/EditAiAssistantTest.php
@@ -161,7 +161,7 @@ it('can edit a record', function () use ($licenses, $permissions) {
             'collection_name' => 'avatar',
         ]
     );
-})->only();
+});
 
 it('validates the inputs', function ($data, $errors) use ($licenses, $permissions) {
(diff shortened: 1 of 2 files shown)


app-modules/ai/tests/Tenant/Feature/Filament/Resources/AiAssistantResource/EditAiAssistantTest.php
```diff
@@ -149,7 +149,7 @@
     assertDatabaseHas(
         AiAssistant::class,
         $request->except([
-            'avatar',
+            'avatar', 'model',
```
Orrison: Does this fail without this added right now? If so, then I believe there is a problem. Because `model` should still be being populated and saved even if the User is not an admin.

It should probably default to what the default institutional model is.

Orrison (changes requested):

Orrison (approved):

PR #1752, opened 2025-08-01
[ADVAPP-1701]: Enhance analytics and implement granular filtering in student reports

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1701

### Technical Description

> Enhance analytics and implement granular filtering in student reports

> No

> No


krlmrr (approved):

app-modules/report/src/Filament/Widgets/Concerns/InteractsWithPageFilters.php
```diff
@@ -76,4 +78,30 @@ public function getMonthRange(Carbon $startDate, Carbon $endDate): Collection
...
+    {
+        if (blank($segmentId)) {
+            return;
+        }
+
+        $query->whereKey(
+            app(TranslateSegmentFilters::class)
+                ->execute($segmentId)
+                ->pluck($query->getModel()->getQualifiedKeyName()),
+        );
```
Orrison: I would modify this to intead simply pass the `$query` to the `applyFilterToQuery` method of `TranslateSegmentFilters`.

It is not scalable to run `pluck` and `whereKey` here. Segments can be huge so you may run out of allowed items within an `in` statement, memory, query length, etc.

app-modules/report/src/Filament/Widgets/MostRecentStudentsTable.php
```diff
@@ -78,21 +78,25 @@ public function table(Table $table): Table
...
+                    )
+                    ->when(
+                        $segmentId,
+                        fn (Builder $query) => $this->segmentFilter($query, $segmentId)
+                    )
+                    ->orderBy('created_at_source', 'desc')
+                    ->take(100);
+
+                return Student::query()
+                    ->whereIn($key, $subQuery)
```
Orrison: No need to change this here, I just wanted to point out that a `whereIn` DOES make sense here because we are specifically controlling this down to `100` records. So that makes sure we don't overload any of this.

Orrison (changes requested):
Didn't get a chance to review all of this yet, but there is a big scalability issue with the usage of the segment translator. It should be easy to fix by changing to use it's other method though. Let me know if you run into any issues with that.

app-modules/report/src/Filament/Widgets/Concerns/InteractsWithPageFilters.php
```diff
@@ -76,4 +78,31 @@ public function getMonthRange(Carbon $startDate, Carbon $endDate): Collection
...
+        if (blank($segmentId)) {
+            return;
+        }
+
+        $query->whereIn(
+            $query->getModel()->getQualifiedKeyName(),
+            app(TranslateSegmentFilters::class)
+                ->execute($segmentId)
+                ->pluck($query->getModel()->getKeyName())
+        );
```
Orrison: This was not the change I mentioned, it looks like you may have interpreted me saying `in` in my last message as, change this to a `whereIn` I was just saying that a pluck of keys and an `in` is not safe. Both a `whereKey` and `whereIn` are both an `in` statement.

You should not be using any from of an `in` for this. You should not use `execute` from `TranslateSegmentFilters` for this. You should simply pass the `$query` to the `applyFilterToQuery` method of TranslateSegmentFilters which will dynamically scope the query to the segment on the SQL side so you don't run into any issues with memory or query length by plucking hundreds of thousands of IDs from a real segment.

Orrison (changes requested):

Orrison (approved):

PR #1760, opened 2025-08-04
[ADVAPP-1726]: Issues with notifying Users of subscription created for them

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1726

### Technical Description
- Add an `if/return` if the subscribable is null.

- No

- No

### DIFF
--- a/app-modules/notification/src/Listeners/NotifyUserOfSubscriptionCreated.php
+++ b/app-modules/notification/src/Listeners/NotifyUserOfSubscriptionCreated.php
@@ -42,7 +42,6 @@ use AdvisingApp\Prospect\Models\Prospect;
 use AdvisingApp\StudentDataModel\Filament\Resources\StudentResource;
 use AdvisingApp\StudentDataModel\Models\Student;
 use Filament\Notifications\Notification;
-use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Contracts\Queue\ShouldQueueAfterCommit;
 use Illuminate\Support\HtmlString;
 
--- a/app-modules/notification/src/Listeners/NotifyUserOfSubscriptionCreated.php
+++ b/app-modules/notification/src/Listeners/NotifyUserOfSubscriptionCreated.php
@@ -51,7 +51,7 @@ class NotifyUserOfSubscriptionCreated implements ShouldQueue
     {
         $subscribable = $event->subscription->subscribable;
 
-        if ($subscribable === null) {
+        if ($subscribable == null) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


rebecca-canyon (dismissed):

Orrison (changes requested):
Though this would prevent the error from being thrown, this does nothing to resolve the issue at play here.

For some reason, the `$subscribable` is null when this is being run.

This Listener is fired when a Subscription is created for a Subscribable (Student or Prospect) and is queued.

Typically when you encounter things like this, it is a dispatch / database transaction race condition.
i.e. this is getting dispatched and then processed by a worker, before the code that created the Student or Prospect finishes and commits the transaction, saving the details to the database.

Please try and find the root of this issue and remove the current code which would just cover up it happening.
I would trace back from where this listener, to the event, to see where it is fired, AND/OR look to see if events/listeners have a way to ensure they only dispatch after DB transactions in the lifecycle commit, I think Laravel recently added a lot of helpers and interfaces for that.

app-modules/notification/src/Observers/SubscriptionObserver.php
```diff
@@ -39,8 +39,9 @@
 use AdvisingApp\Notification\Events\SubscriptionCreated;
 use AdvisingApp\Notification\Events\SubscriptionDeleted;
 use AdvisingApp\Notification\Models\Subscription;
+use Illuminate\Contracts\Events\ShouldHandleEventsAfterCommit;
 
-class SubscriptionObserver
+class SubscriptionObserver implements ShouldHandleEventsAfterCommit
```
Orrison: This is for sure the right direction. But I think I would recommend bringing the commit check down into the individual Event/Listener.

I believe this is an interface you can put on a queued listener that will accomplish this: https://laravel.com/docs/12.x/events#queued-event-listeners-and-database-transactions
But ONLY for that listener.

Which is think may be better than applying this to the entire observer. If we do so, there could be issue with like doing stuff in the `-ing` methods like filling something before it is created in the DB during `creating`

krlmrr: This should be good to go now.

app-modules/notification/src/Listeners/NotifyUserOfSubscriptionCreated.php
```diff
@@ -43,9 +43,10 @@
 use AdvisingApp\StudentDataModel\Models\Student;
 use Filament\Notifications\Notification;
 use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Contracts\Queue\ShouldQueueAfterCommit;
 use Illuminate\Support\HtmlString;
 
-class NotifyUserOfSubscriptionCreated implements ShouldQueue
+class NotifyUserOfSubscriptionCreated implements ShouldQueue, ShouldQueueAfterCommit
```
Orrison: I think you are supposed to replace `ShouldQueue` with `ShouldQueueAfterCommit`

PR #1767, opened 2025-08-08
[ADVAPP-1796]: Introduce dynamic user representation for "From Name" in outbound email and text messages

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1796

### Technical Description

Introduces new engagement settings (SES settings are not the right place to store this, in my opinion). Reads that setting in the notification and sets the from name. To avoid setting the from address too, I have overridden the `from()` method on the base message object to pass the correct address by default.

No

EngagementSettingsFeature

### DIFF
--- a/app-modules/engagement/src/Settings/EngagementSettings.php
+++ b/app-modules/engagement/src/Settings/EngagementSettings.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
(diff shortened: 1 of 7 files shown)


krlmrr (approved):

app-modules/engagement/src/Settings/EngagementSettings.php
```diff
@@ -0,0 +1,49 @@
...
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Engagement\Settings;
+
+use Spatie\LaravelSettings\Settings;
+
+class EngagementSettings extends Settings
+{
+    public bool $are_dynamic_engagements_enabled = false;
```
Orrison, line 43: I know this is a little long, but given what this does, would `are_dynamic_engagements_from_name_enabled` be more appropriate?

Or `is_dynamic_from_name_enabled`?

Orrison (changes requested):

app-modules/engagement/src/Settings/EngagementSettings.php
danharrin: As discussed, this was what product asked for

PR #1768, opened 2025-08-08
[ADVAPP-1804]: Enhance the SIS sync records tool to allow care team assignments

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1804

### Technical Description

Create a temporary table for care team assignments, insert them into the temporary table, then clear out any existing assignments and insert them in the same transaction.

No

No

### DIFF
--- a/app-modules/student-data-model/src/Actions/CompleteStudentDataImport.php
+++ b/app-modules/student-data-model/src/Actions/CompleteStudentDataImport.php
@@ -76,6 +76,17 @@ class CompleteStudentDataImport
                 DB::statement('drop table "students"');
                 DB::statement("alter table \"import_{$import->studentsImport->getKey()}_students\" rename to \"students\"");
 
+                DB::table('care_teams')
+                    ->where('educatable_type', 'student')
+                    ->delete();
+
+                DB::statement("insert into care_teams
+                    (id, user_id, educatable_id, educatable_type, created_at, updated_at, care_team_role_id)
+                    select id, user_id, educatable_id, educatable_type, created_at, updated_at, care_team_role_id
+                    from \"import_{$import->studentsImport->getKey()}_care_teams\"");
+
+                DB::statement("drop table \"import_{$import->studentsImport->getKey()}_care_teams\"");
+
                 if ($import->emailAddressesImport) {
                     DB::statement('drop table "student_email_addresses"');
                     DB::statement("alter table \"import_{$import->emailAddressesImport->getKey()}_email_addresses\" rename to \"student_email_addresses\"");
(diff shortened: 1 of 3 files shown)


app-modules/student-data-model/src/Actions/CompleteStudentDataImport.php
```diff
@@ -76,6 +76,16 @@ public function execute(
                 DB::statement('drop table "students"');
                 DB::statement("alter table \"import_{$import->studentsImport->getKey()}_students\" rename to \"students\"");
 
+                DB::statement("insert into care_teams
+                    (id, user_id, educatable_id, educatable_type, created_at, updated_at, care_team_role_id)
+                    select id, user_id, educatable_id, educatable_type, created_at, updated_at, care_team_role_id
+                    from \"import_{$import->studentsImport->getKey()}_care_teams\"
+                    on conflict (user_id, educatable_id, educatable_type) do update
+                    set care_team_role_id = excluded.care_team_role_id, updated_at = excluded.updated_at
+                    where excluded.updated_at > care_teams.updated_at");
```
Orrison, line 85: bueno

app-modules/student-data-model/src/Filament/Imports/StudentImporter.php
```diff
@@ -239,4 +306,53 @@ public function getJobBatchName(): ?string
...
+                continue;
+            }
+
+            $user = User::query()
+                ->where('email', $this->data[$careTeamColumn])
+                ->first();
+
+            if (! $user) {
+                continue;
+            }
```
Orrison, line 331: The ticket does say that the row should fail if the email address is unmatched. Is that possible here?

app-modules/student-data-model/src/Filament/Imports/StudentImporter.php
```diff
@@ -239,4 +306,53 @@ public function getJobBatchName(): ?string
...
+
+            if (filled($this->data[$careTeamRoleColumn])) {
+                $careTeamRole = CareTeamRole::query()
+                    ->where(new Expression('lower(name)'), Str::lower($this->data[$careTeamRoleColumn]))
+                    ->first();
+            }
+
+            if (! ($careTeamRole ?? null)) {
+                $defaultCareTeamRole ??= CareTeamRoleType::studentDefault();
```
Orrison, line 344: The null assignment here doesn't make sense since `$defaultCareTeamRole` has not been defined yet at this point.

Orrison (changes requested):

app-modules/student-data-model/src/Filament/Imports/StudentImporter.php
danharrin: There is a `Rule::exists()`, this is just a safety check.

app-modules/student-data-model/src/Filament/Imports/StudentImporter.php
danharrin: It is in a loop, it can be defined by a previous iteration of the loop. It is so we do not need to run a query to fetch the default for each item in the loop.

Orrison: Ah, I see good point.

Orrison (approved):

PR #1772, opened 2025-08-11
[ADVAPP-1792]: Clean up ProjectPageFeature

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1792

### Technical Description
- Remove Feature
- Remove Migration
- Remove implimentation

- No

- No


app-modules/project/src/Filament/Resources/ProjectResource.php
```diff
@@ -61,7 +60,7 @@ public static function canAccess(): bool
         /** @var User $user */
         $user = auth()->user();
 
-        return ProjectPageFeature::active() && $user->can('project.view-any');
+        return $user->can('project.view-any');
     }
```
Orrison: I think it actually would just be fine in this file to just remove this override of `canAccess` since the parent would be checking that same permission anyways.

Orrison (changes requested):

Orrison (approved):

krlmrr:
> Karl it looks like this we effected by the GitHub outages. The only way to get it to trigger CI checks again is to push something up. So I would recommend rebasing the branch onto main and force pushing up to get it to re-trigger.
> 
> Once you have done so and CI is passing, please re-request my review.

Will do.

Orrison (approved):

PR #1794, opened 2025-08-18
[ADVAPP-1816]: Cleanup ProjectFileFeature

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1816

### Technical Description
- Remove Feature
- Remove Migration 
- Remove implementation

- No

- No


app-modules/project/src/Filament/Resources/ProjectResource/Pages/ManageFiles.php
```diff
@@ -67,13 +66,6 @@ public static function getNavigationLabel(): string
         return 'Files';
     }
 
-    public static function canAccess(array $arguments = []): bool
-    {
-        $user = auth()->user();
-
-        return ProjectFileFeature::active() && $user->can('viewAny', [ProjectFile::class, $arguments['record']]);
-    }
```
Orrison: This `canAccess` actually does have a specific override in the `$user->can('viewAny', [ProjectFile::class, $arguments['record']])` that should be retained.

In fact, you can see that a test is failing here because this was removed.

Restore this override and keep the `$user->can('viewAny', [ProjectFile::class, $arguments['record']])` part.

Orrison (changes requested):

app-modules/project/src/Filament/Resources/ProjectResource/Pages/ManageFiles.php
krlmrr: I caught that, I was actually in the middle of fixing it.

PR #1796, opened 2025-08-18
[ADVAPP-1775]: Ensure that Subscriptions take into account soft deleted Users and/or the Educatables to prevent unnecessary exceptions

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1775

### Technical Description

Ensure that users and educatables are not deleted before creating a subscription between them.

No

No


rebecca-canyon:
@Orrison Gotcha. In that case, I believe that returning in the places where I've put in exceptions ought to do it.

Most likely, I would see this being an issue with Campaigns, and eventually Workflows; but I think it would technically be possible for a User to try and subscribe (or bulk subscribe) to an Educatable, and for one or the other to be deleted before submission.

PR #1798, opened 2025-08-19
[ADVAPP-1828]: Transition institutional advisor and custom advisors to websockets experience for improved streaming

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1828

### Technical Description

Replaces streamed responses with web sockets for all advisors.

No.

No.


Orrison (changes requested):
This all looks pretty good. Tested locally everything seemed to be working well.

But, and I am unsure if these changes caused this issue, I noticed that with the continue generating functionality something seems wrong.

For example, I asked it to write me a long story. It started off well, but then ran out of the token limit and add the `...` and the continue generating button.
But when I pressed continue generating, instead of continuing the story, it acted as if I had asked it to write me a story again and asked me again what gener I would like.

I have attached a screenshot of this in the ticket.

Can we check to see if you can replicate that and if it could be a result of changes here?

danharrin:
@Orrison I tested the feature quite a bit while developing, and only saw that once or twice. it is a random chance that the AI might not understand the request to continue the previous response, but it does work and is not a result of this PR. The old implementation had the same random issue. I attached a video to Jira to show you.

PR #1799, opened 2025-08-19
[ADVAPP-1825]: Introduce a file upload form field type

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1825

### Technical Description

> Introduce a file upload form field type

> No

> No

### DIFF
--- a/app-modules/form/resources/views/blocks/submissions/upload.blade.php
+++ b/app-modules/form/resources/views/blocks/submissions/upload.blade.php
@@ -1,0 +1,17 @@
+<x-form::blocks.field-wrapper
+    class="py-3"
+    :$label
+    :$isRequired
+>
+@dd($response)
+    {{-- @if (filled($response ?? null))
+        <a
+            href="{{ $response }}"
+            target="_blank"
+        >
+            {{ $response }}
+        </a>
+    @else
+        <span class="text-gray-500">No response</span>
+    @endif --}}
+</x-form::blocks.field-wrapper>
--- a/app-modules/form/resources/views/blocks/submissions/upload.blade.php
+++ b/app-modules/form/resources/views/blocks/submissions/upload.blade.php
@@ -37,23 +37,18 @@
     :$isRequired
 >
     <div class="flex flex-col space-y-2">
-        @foreach ($response as $file)
-            @php
-                $url = Storage::disk('s3')->temporaryUrl($file['path'], now()->addMinutes(5), [
-                    'ResponseContentDisposition' => 'attachment; filename="' . basename($file['path']) . '"',
-                ]);
-            @endphp
-            <a
-                class="inline-flex items-center space-x-1 text-sm text-primary-600 hover:underline"
-                href="{{ $url }}"
-            >
-                <span>{{ $file['originalFileName'] }}</span>
-                <x-heroicon-s-arrow-down-tray class="h-4 w-4" />
-            </a>
-        @endforeach
+        @if ($media && count($media) > 0)
+            @foreach ($media as $item)
+                <a
+                    class="inline-flex items-center space-x-1 text-sm text-primary-600 hover:underline"
+                    href="{{ $item['temporary_url'] }}"
+                >
+                    <span>{{ $item['name'] }}</span>
+                    <x-heroicon-s-arrow-down-tray class="h-4 w-4" />
+                </a>
+            @endforeach
+        @else
+            <span class="text-gray-500">No response</span>
+        @endif
     </div>
-
-    @if (blank($response ?? null))
-        <span class="text-gray-500">No response</span>
-    @endif
 </x-form::blocks.field-wrapper>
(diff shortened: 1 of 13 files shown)


widgets/form/src/FormKit/Inputs/Upload.vue
```diff
@@ -0,0 +1,197 @@
...
+                }),
+            );
+            load();
+            return;
+        }
+
+        const index = fileIndexCounter.value++;
+        const { get } = consumer();
+        try {
+            // const data = await get(props.context.uploadUrl, {
```
danharrin: Please remove dead code

app-modules/form/src/Filament/Blocks/UploadFormFieldBlock.php
danharrin, line 1: Currently there is no way to switch the input from "single" to "multiple" and control the max number of files. As per the ticket:

> Add in the form block, the ability to define how many files can be uploaded. Max value is 5

app-modules/form/resources/views/blocks/submissions/upload.blade.php
```diff
@@ -0,0 +1,60 @@
...
+                <span>{{ $file['originalFileName'] }}</span>
+                <x-heroicon-s-arrow-down-tray class="h-4 w-4" />
+            </a>
+        @endforeach
+    </div>
+
+    @if (blank($response ?? null))
+        <span class="text-gray-500">No response</span>
+    @endif
+
```
danharrin: Please remove this new line

app-modules/form/src/Filament/Blocks/UploadFormFieldBlock.php
amit-canyon: @danharrin   Could you please suggest a way to pass data? I'm having trouble getting it to the form block.

I have another ticket to add an upload block in the admissions application. If the requirements are the same, we can reuse the existing `UploadFormFieldBlock`. However, if the upload requirements differ, we should either reuse it with configurable parameters (if possible) or create a separate upload block specifically for the admissions form. here is the data requirement:

danharrin:
Please re-request my review when you have addressed what we discussed about storing the files permanently after the form is submitted

amit-canyon:
> Please re-request my review when you have addressed what we discussed about storing the files permanently after the form is submitted

Yes, @danharrin  working on it. Thanks

app-modules/form/src/Actions/ProcessSubmissionField.php
```diff
@@ -53,6 +56,35 @@ public function __invoke(Submission $submission, string $fieldId, mixed $respons
             'response' => $response,
         ]);
 
+        $field = FormField::find($fieldId);
```
danharrin: I think the pivot would already be loaded if you use `$submission->fields()->find($fieldId)`, so then the extra query for `$formFieldSubmission` could be avoided?

app-modules/form/src/Actions/ProcessSubmissionField.php
```diff
@@ -53,6 +56,35 @@ public function __invoke(Submission $submission, string $fieldId, mixed $respons
...
+
+        if ($field && $field->type === 'upload' && is_array($response)) {
```
danharrin: Instead of the `'upload'` string, maybe use `UploadFormFieldBlock::type()`?

app-modules/form/src/Actions/ProcessSubmissionField.php
```diff
@@ -53,6 +56,35 @@ public function __invoke(Submission $submission, string $fieldId, mixed $respons
...
+                ->fields()
+                ->where('field_id', $fieldId)
+                ->latest('created_at')
+                ->first()
+                ->pivot;
+
+            foreach ($response as $file) {
+                $key = ltrim($file['path'], '/');
+
+                $url = Storage::disk('s3')->temporaryUrl(
```
danharrin: Instead of generating a temporary URL and then adding it to the media collection, does `$formFieldSubmission->copyMedia($key)` work? Maybe it doesn't, but that would avoid it being downloaded.

app-modules/form/src/Actions/ProcessSubmissionField.php
```diff
@@ -53,6 +56,35 @@ public function __invoke(Submission $submission, string $fieldId, mixed $respons
...
+                    [
+                        'ResponseContentDisposition' => 'attachment; filename="' . basename($file['originalFileName'] ?? $key) . '"',
+                    ]
+                );
+
+                $formFieldSubmission
+                    ->addMediaFromUrl($url)
+                    ->usingFileName($file['originalFileName'] ?? basename($key))
+                    ->toMediaCollection('files', 's3-public');
+            }
```
danharrin, line 83: After uploading the media, I think you need to delete the old temporary file, and then save the new URL to the `FormFieldSubmission`?

@Orrison should the permanent files be stored privately instead of publicly?

Orrison: They should be private

amit-canyon: image

amit-canyon: I have changed '`s3-public` to `s3`. @Orrison

danharrin (approved):

PR #1800, opened 2025-08-19
[ADVAPP-1770]: Add Navigation Badge for Admissions

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1770

### Technical Description
- Add Navigation Badge for Admissions

- No

- No

### DIFF
--- a/app-modules/application/src/Providers/ApplicationServiceProvider.php
+++ b/app-modules/application/src/Providers/ApplicationServiceProvider.php
@@ -36,20 +36,20 @@
 
 namespace AdvisingApp\Application\Providers;
 
-use Filament\Panel;
-use Illuminate\Support\Facades\Event;
-use Illuminate\Support\ServiceProvider;
 use AdvisingApp\Application\ApplicationPlugin;
-use AdvisingApp\Application\Models\Application;
-use AdvisingApp\Application\Models\ApplicationStep;
-use AdvisingApp\Application\Models\ApplicationField;
-use Illuminate\Database\Eloquent\Relations\Relation;
-use AdvisingApp\Application\Models\ApplicationSubmission;
-use AdvisingApp\Application\Models\ApplicationAuthentication;
-use AdvisingApp\Application\Models\ApplicationSubmissionState;
 use AdvisingApp\Application\Events\ApplicationSubmissionCreated;
 use AdvisingApp\Application\Listeners\ClearApplicationSubmissionCountCache;
 use AdvisingApp\Application\Listeners\NotifySubscribersOfApplicationSubmission;
+use AdvisingApp\Application\Models\Application;
+use AdvisingApp\Application\Models\ApplicationAuthentication;
+use AdvisingApp\Application\Models\ApplicationField;
+use AdvisingApp\Application\Models\ApplicationStep;
+use AdvisingApp\Application\Models\ApplicationSubmission;
+use AdvisingApp\Application\Models\ApplicationSubmissionState;
+use Filament\Panel;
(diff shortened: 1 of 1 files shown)


app-modules/application/src/Filament/Resources/ApplicationResource/Pages/ManageApplicationSubmissions.php
```diff
@@ -189,4 +190,22 @@ public function table(Table $table): Table
...
+        $item = parent::getNavigationItems($urlParameters)[0];
+
+        $ownerRecord = $urlParameters['record'];
+
+        $formSubmissionsCount = Cache::tags('{form-submission-count}')
+            ->remember(
+                "form-submission-count-{$ownerRecord->getKey()}",
+                now()->addMinutes(5),
+                fn (): int => $ownerRecord->submissions()->count(),
+            );
```
Orrison: Because these are for applications and not forms I would change the tag and the key prefix to be `application-submission-count` so it doesn't get mixed up with the form stuff.

I also noticed that this cache is not cleared when I submit an application, so I have to wait 5 minutes to see this updated value.

This was solved in forms via the Listener `app-modules/form/src/Listeners/ClearFormFormSubmissionCountCache.php`. There should be a similar Event for Applications you can add a Listener for.
I think it's `ApplicationSubmissionCreated`

krlmrr: Thanks for the advice! It should be implemented now.

app-modules/application/src/Filament/Resources/ApplicationResource/Pages/ManageApplicationSubmissions.php
```diff
@@ -189,4 +190,22 @@ public function table(Table $table): Table
...
+
+    public static function getNavigationItems(array $urlParameters = []): array
+    {
+        $item = parent::getNavigationItems($urlParameters)[0];
+
+        $ownerRecord = $urlParameters['record'];
+
+        $applicationSubmissionsCount = Cache::tags('{application-submission-count}')
+            ->remember(
+                "applciation-submission-count-{$ownerRecord->getKey()}",
```
Orrison:
```suggestion
                "application-submission-count-{$ownerRecord->getKey()}",
```

app-modules/application/src/Listeners/ClearApplicationSubmissionCountCache.php
```diff
@@ -0,0 +1,51 @@
...
+use AdvisingApp\Application\Events\ApplicationSubmissionCreated;
+use Illuminate\Support\Facades\Cache;
+
+class ClearApplicationSubmissionCountCache
+{
+    public function handle(ApplicationSubmissionCreated $event): void
+    {
+        Cache::tags('{application-submission-count}')
+            ->forget(
+                "applciation-submission-count-{$event->submission->submissible->getKey()}"
```
Orrison:
```suggestion
                "application-submission-count-{$event->submission->submissible->getKey()}"
```

app-modules/application/src/Listeners/ClearApplicationSubmissionCountCache.php
Orrison, line 1: Make sure to also register this to the appropriate event in the ApplicationServiceProvider.

You can see how that is done with the equivalent for Forms.

We have avoided Event/Listener auto discovery due to performance issues.

Orrison (changes requested):

Orrison (approved):

PR #1801, opened 2025-08-19
[ADVAPP-1829]: Create Resource, Model, Migrations and Policy

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1829

### Technical Description

>

>

>

### DIFF
--- a/app-modules/ai/database/migrations/2025_08_19_181044_seed_ai_data_advisor_permissions.php
+++ b/app-modules/ai/database/migrations/2025_08_19_181044_seed_ai_data_advisor_permissions.php
@@ -1,0 +1,40 @@
+<?php
+
+use CanyonGBS\Common\Database\Migrations\Concerns\CanModifyPermissions;
+use Illuminate\Database\Migrations\Migration;
+
+return new class extends Migration
+{
+    use CanModifyPermissions;
+
+    /** @var array<string, string> */
+    private array $permissions = [
+        'ai-data-advisors.view-any' => 'Data Advisor',
+        'ai-data-advisors.*.view' => 'Data Advisor',
+        'ai-data-advisors.create' => 'Data Advisor',
+        'ai-data-advisors.*.update' => 'Data Advisor',
+        'ai-data-advisors.*.delete' => 'Data Advisor',
+        'ai-data-advisors.*.restore' => 'Data Advisor',
+        'ai-data-advisors.*.force-delete' => 'Data Advisor',
+    ];
+
+    /** @var array<string> */
+    private array $guards = [
+        'web',
+        'api',
+    ];
+
+    public function up(): void
+    {
+        foreach ($this->guards as $guard) {
+            $this->createPermissions($this->permissions, $guard);
+        }
+    }
+
+    public function down(): void
+    {
+        foreach ($this->guards as $guard) {
+            $this->deletePermissions(array_keys($this->permissions), $guard);
+        }
+    }
+};
--- a/app-modules/ai/src/Filament/Resources/DataAdvisorResource.php
+++ b/app-modules/ai/src/Filament/Resources/DataAdvisorResource.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace AdvisingApp\Ai\Filament\Resources;
+
+use Filament\Forms;
+use Filament\Tables;
+use Filament\Forms\Form;
+use Filament\Tables\Table;
+use Filament\Resources\Resource;
+use AdvisingApp\Ai\Models\DataAdvisor;
+use Filament\Tables\Actions\EditAction;
+use Illuminate\Database\Eloquent\Builder;
+use Filament\Tables\Actions\BulkActionGroup;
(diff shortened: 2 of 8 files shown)


app-modules/ai/database/migrations/2025_08_19_181044_seed_ai_data_advisor_permissions.php
Orrison, line 1: Let's update this permission migration to have the prefix `seed_permissions_ ` rather then just `seed` to match others.

app-modules/ai/src/Filament/Resources/DataAdvisorResource.php
```diff
@@ -0,0 +1,92 @@
...
+
+class DataAdvisorResource extends Resource
+{
+    protected static ?string $model = DataAdvisor::class;
+
+    protected static ?string $navigationIcon = 'heroicon-o-rectangle-stack';
+
+    protected static ?string $navigationGroup = 'Artificial Intelligence';
+
+    protected static ?int $navigationSort = 35;
```
Orrison: We have typically incremented by 10s. So it may be best to follow that and have this be `40`

app-modules/ai/src/Models/DataAdvisor.php
```diff
@@ -0,0 +1,47 @@
...
+
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Ai\Models;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\SoftDeletes;
+
+class DataAdvisor extends Model
```
Orrison: Let's add the Auditing interface and trait to this so we don't forget

app-modules/ai/src/Policies/DataAdvisorPolicy.php
```diff
@@ -0,0 +1,100 @@
...
+        return $authenticatable->canOrElse(
+            abilities: 'ai-data-advisors.view-any',
+            denyResponse: 'You do not have permission to view data advisors.'
+        );
+    }
+
+    public function view(Authenticatable $authenticatable, DataAdvisor $dataAdvisor): Response
+    {
+        return $authenticatable->canOrElse(
+            abilities: ["ai-data-advisors.{$dataAdvisor->getKey()}.view"],
```
Orrison: Here and in the other places in this file please change `{$dataAdvisor->getKey()}` to just the wildcard `*`. We don't check specific record access anymore as it wasn't efficient.

app-modules/ai/src/Policies/DataAdvisorPolicy.php
```diff
@@ -0,0 +1,100 @@
...
+            denyResponse: 'You do not have permission to view data advisors.'
```
Orrison:
```suggestion
            denyResponse: 'You do not have permission to view this data advisor.'
```

app-modules/ai/src/Policies/DataAdvisorPolicy.php
```diff
@@ -0,0 +1,100 @@
...
+            abilities: 'ai-data-advisors.create',
+            denyResponse: 'You do not have permission to create data advisors.'
+        );
+    }
+
+    public function update(Authenticatable $authenticatable, DataAdvisor $dataAdvisor): Response
+    {
+        return $authenticatable->canOrElse(
+            abilities: ["ai-data-advisors.{$dataAdvisor->getKey()}.update"],
+            denyResponse: 'You do not have permission to create data advisors.'
```
Orrison:
```suggestion
            denyResponse: 'You do not have permission to update this data advisor.'
```

app-modules/ai/src/Filament/Resources/DataAdvisorResource/Pages/EditDataAdvisor.php
```diff
@@ -0,0 +1,56 @@
...
+class EditDataAdvisor extends EditRecord
+{
+    use EditPageRedirection;
+
+    protected static string $resource = DataAdvisorResource::class;
+
+    protected function getHeaderActions(): array
+    {
+        return [
+            Actions\DeleteAction::make(),
```
Orrison: Please fully import that class.

Orrison (changes requested):
Interesting, we have not done this before. Precedently we would create a stand-alone Filament pages for this that we would delete later.

Especially since we have not been given much detail or done any planning for what a Data Advisor could be, it is possible we wouldn't have it be it's own Model. And though not a huge deal, because you are creating the migration now this now means that the columns we add later will start off at the end.

I think this is okay though, if we go a different route we can just delete the Model and the table later. But there is some feedback to be addressed.

Orrison (changes requested):
Linting and tests for this are failing. Looks mostly to be due to the wrong classes being used for auditing in the new Model.

Take a look at other Models with Auditing for guidance on what to do here.

PR #1802, opened 2025-08-19
[ADVAPP-1740]: Update the SSO toggle functionality within the multifactor authentication section of Profile Settings

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1740

### Technical Description

Update label from 'SSO Enabled' to 'Product MFA Disabled'

No

No

### DIFF
--- a/app-modules/multifactor-authentication/src/Livewire/MultifactorAuthenticationManagement.php
+++ b/app-modules/multifactor-authentication/src/Livewire/MultifactorAuthenticationManagement.php
@@ -187,7 +187,7 @@ class MultifactorAuthenticationManagement extends Component implements HasAction
             Toggle::make('is_sso_enabled')
                 ->helperText('Your account is configured using enterprise single sign-on (SSO) so multifactor authentication is handled by
         your institution, rather than by this product.')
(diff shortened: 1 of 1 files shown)


app-modules/multifactor-authentication/src/Livewire/MultifactorAuthenticationManagement.php
```diff
@@ -187,7 +187,7 @@ public function getFormSchema(): array
             Toggle::make('is_sso_enabled')
                 ->helperText('Your account is configured using enterprise single sign-on (SSO) so multifactor authentication is handled by
         your institution, rather than by this product.')
-                ->label('SSO Enabled')
+                ->label('Product MFA Disabled')
```
Orrison: This label only needs to be changed if the User logs in via SSO. Which is indicated by the User having `is_external` being `true`.

So I would change this to a function and check if the Users `is_external` attribute and do this new wording if `true` and the old wording if not.

Orrison (changes requested):

Orrison (approved):

PR #1807, opened 2025-08-20
[ADVAPP-1757]: Introduce the ability to associate tasks with a project

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1757

### Technical Description

Associate tasks with projects

No

AssociateTasksWithProjectsFeature feature flag

### DIFF
--- a/app-modules/project/src/Filament/Resources/ProjectResource/Pages/ManageTasks.php
+++ b/app-modules/project/src/Filament/Resources/ProjectResource/Pages/ManageTasks.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace AdvisingApp\Project\Filament\Resources\ProjectResource\Pages;
+
+use AdvisingApp\Project\Filament\Resources\ProjectResource;
+use App\Features\AssociateTasksWithProjectsFeature;
+use Filament\Actions;
+use Filament\Resources\Pages\ManageRelatedRecords;
+use Filament\Tables\Actions\AssociateAction;
+use Filament\Tables\Actions\BulkActionGroup;
+use Filament\Tables\Actions\DissociateAction;
+use Filament\Tables\Actions\DissociateBulkAction;
+use Filament\Tables\Columns\TextColumn;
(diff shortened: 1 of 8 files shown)


app-modules/project/src/Filament/Resources/ProjectResource/Pages/ManageTasks.php
```diff
@@ -0,0 +1,94 @@
...
+    public static function getNavigationLabel(): string
+    {
+        return 'Tasks';
+    }
+
+    public static function canAccess(array $parameters = []): bool
+    {
+        $user = auth()->user();
+
+        return AssociateTasksWithProjectsFeature::active() && $user->can(['project.*.view']) && parent::canAccess($parameters);
```
Orrison: I am not 100% sure here, but I think you can get rid of `$user->can(['project.*.view'])` as I think the parent Filament method should check if you can view the parent record of this `ManageRelatedRecords` page. Which should be validated by the access testing in your test.

If that doesn't work then it may make more sense to do `$user->can(['view', $this->getOwnerRecord())`

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ManageTasksTest.php
```diff
@@ -0,0 +1,166 @@
...
+        'record' => $project->getRouteKey(),
+    ])
+        ->assertCanSeeTableRecords($project->tasks);
+});
+
+it('does not list tasks already associated with a project in task search', function () {
+    asSuperAdmin();
+
+    $project1 = Project::factory()->create();
+    $project2 = Project::factory()->create();
```
Orrison: Let's no use numbers in variable definitions, let's do like `projectOne` instead.

Orrison (changes requested):

Orrison (approved):

PR #1816, opened 2025-08-22
[ADVAPP-1835]: Introduce initial styled baseline for embedded QnA Advisor

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1835

### Technical Description

> Introduce initial styled baseline for embedded QnA

> No

> No

### DIFF
--- a/widgets/qna-advisor/src/App.vue
+++ b/widgets/qna-advisor/src/App.vue
@@ -36,9 +36,8 @@ import Echo from 'laravel-echo';
 import Pusher from 'pusher-js/dist/web/pusher';
 import { defineProps, onMounted, onUnmounted, ref } from 'vue';
 import headshotAgent from '../../../resources/images/canyon-ai-headshot.jpg?url';
-import userAvatar from '../public/images/user-default-avatar.svg?url';
 import loadingSpinner from '../public/images/loading-spinner.svg?url';
-
+import userAvatar from '../public/images/user-default-avatar.svg?url';
 
 const props = defineProps(['url']);
 const sendMessageUrl = ref(null);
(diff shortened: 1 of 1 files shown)


widgets/qna-advisor/src/App.vue
```diff
@@ -145,23 +153,80 @@ async function sendMessage() {
...
-        <div v-if="isLoading">
-            <p>AI is typing...</p>
-        </div>
-
-        <div>
-            <textarea v-model="message" placeholder="Ask your question..." :disabled="isLoading"></textarea>
-            <button @click="sendMessage" :disabled="isLoading || !message.trim()">Send</button>
+        <div class="relative h-screen" v-if="sendMessageUrl === null">
+            <div class="absolute top-1/2 left-1/2 -translate-x-1/2 -translate-y-1/2 text-center">
+                <div role="status">
```
Orrison: Is this `role` needed for something? If not, then it can be removed.

widgets/qna-advisor/src/App.vue
Orrison, line 1: Some elements you have applied `dark` classes to. Therefore if your system preferences on your computer are set to dark-mode then some dark-mode CSS is currently applied.

Given this is the case, please switch to dark mode on your local system temporary and fix any styling inconsistencies.

Image

public/images/loading-spinner.svg
Orrison, line 1: Can we try and move this to the widget itself and pull it in via import? So that it is bundled into the widget?

public/images/user-default-avatar.svg
Orrison, line 1: Can we try and move this to the widget itself and pull it in via import? So that it is bundled into the widget?

PR #1821, opened 2025-08-26
[ADVAPP-1783]: Implement a workflow step for sending emails in admissions and forms

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1783

### Technical Description

> Implement a workflow step for sending emails in admissions and forms

> No

> No

### DIFF
--- a/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
+++ b/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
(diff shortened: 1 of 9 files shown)


app-modules/workflow/database/factories/WorkflowRunStepFactory.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * @extends Factory<WorkflowRunStep>
+ */
+class WorkflowRunStepFactory extends Factory
+{
+    public function definition(): array
+    {
+        return [
+            'workflow_run_id' => WorkflowRun::factory(),
+            'details_type' => WorkflowEngagementEmailDetails::class,
```
Orrison: All of these workflow detail type models (and all Models) should be in the morph map and referred to as such. Please let me know if this is not the case and what it would take to fix that.

payal-canyon: Hey @Orrison !
Sure. I’ve updated the `WorkflowRunStepFactory` so that all detail models are resolved via the morph map. By default, it randomly assigns one of the supported detail factories, and for tests we can explicitly provide a model using the `withDetails()` helper. This ensures that `details_type` is always stored using the model’s morph class, as suggested. Please let me know if you’d like me to make any changes to this approach.

Thank You!

danharrin:
@Orrison the field prefix stuff indeed needs to be removed, but it's an issue with the underlying workflow system, not this block specifically. This block is just following the existing framework.

payal-canyon:
> @Orrison the field prefix stuff indeed needs to be removed, but it's an issue with the underlying workflow system, not this block specifically. This block is just following the existing framework.



> I have one note so far please do re require my review when that is addressed.
> 
> I would also like @danharrin to take a look as I believe we intended to get rid of the `$fieldPrefix` stuff in the block here

Hey @Orrison and @danharrin!
I just tested it without the prefix, and it’s working fine. I think we’re good to remove it.

Thank You!

danharrin:
@payal-canyon have you tested both creation and editing of a block? I think it is there to support the differences between repeaters and relation managers

payal-canyon:
> @payal-canyon have you tested both creation and editing of a block? I think it is there to support the differences between repeaters and relation managers

Hey @danharrin !
I’ve re-added the prefix as discussed.

PR #1832, opened 2025-08-29
[ADVAPP-1867]: Add a merge value across all case type templates for contact name

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1867

### Technical Description

> Add a merge value across all case type templates for contact name

> No

> No

### DIFF
--- a/app-modules/prospect/src/Filament/Resources/ProspectResource/RelationManagers/CasesRelationManager.php
+++ b/app-modules/prospect/src/Filament/Resources/ProspectResource/RelationManagers/CasesRelationManager.php
@@ -43,7 +43,6 @@ use AdvisingApp\CaseManagement\Filament\Resources\CaseResource\Pages\CreateCase;
 use AdvisingApp\CaseManagement\Filament\Resources\CaseResource\Pages\ViewCase;
 use AdvisingApp\CaseManagement\Models\CaseModel;
 use AdvisingApp\CaseManagement\Models\CasePriority;
-use AdvisingApp\Division\Models\Division;
 use App\Filament\Tables\Columns\IdColumn;
 use Filament\Forms\Form;
 use Filament\Infolists\Infolist;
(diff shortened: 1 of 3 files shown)


app-modules/prospect/src/Filament/Resources/ProspectResource/RelationManagers/CasesRelationManager.php
```diff
@@ -99,7 +100,14 @@ public function table(Table $table): Table
             ])
             ->headerActions([
                 CreateAction::make()
-                    ->modalHeading('Create new case'),
+                    ->modalHeading('Create new case')
+                    ->mutateFormDataUsing(function (array $data): array {
```
danharrin: Can you please explain this addition?

danharrin (changes requested):

app-modules/prospect/src/Filament/Resources/ProspectResource/RelationManagers/CasesRelationManager.php
jayushi-canyon: @danharrin while creating a case divison_id should be filled automatically if there is only 1 division but while working I found out that, that functionality was not working as expected and division_id was going null and giving error. Therefore, I tried various ways to solve that and added this logic during createAction and that seemed to have solved the issue.

danharrin: Hmm. I think this should probably be in an observer. What do you think @Orrison?

Orrison: @danharrin I agree that I would make this fix in an observer for cases

danharrin (approved):

PR #1840, opened 2025-09-01
[ADVAPP-1866]: Enhance interactions by providing configurability to institutions around which metadata they would like to collect and display

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1866

### Technical Description

> Enhance interactions by providing configurability to institutions around which metadata they would like to collect and display

> No

> Yes. Feature Flag: `InteractionMetadataFeature`

### DIFF
--- a/app-modules/interaction/database/migrations/2025_09_01_114712_create_interaction_management_settings.php
+++ b/app-modules/interaction/database/migrations/2025_09_01_114712_create_interaction_management_settings.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Advising App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 use Spatie\LaravelSettings\Exceptions\SettingAlreadyExists;
 use Spatie\LaravelSettings\Migrations\SettingsMigration;
 
--- a/app-modules/interaction/resources/views/filament/pages/list-interaction-drivers.blade.php
+++ b/app-modules/interaction/resources/views/filament/pages/list-interaction-drivers.blade.php
@@ -1,3 +1,36 @@
+{{--
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Advising App™ is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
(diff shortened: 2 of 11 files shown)


app-modules/interaction/database/migrations/2025_09_01_114712_create_interaction_management_settings.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->migrator->add('interaction_management.is_outcome_required', true);
+
+        $this->migrator->add('interaction_management.is_relation_enabled', true);
+        $this->migrator->add('interaction_management.is_relation_required', true);
+
+        $this->migrator->add('interaction_management.is_status_enabled', true);
+        $this->migrator->add('interaction_management.is_status_required', true);
+
+        $this->migrator->add('interaction_management.is_type_enabled', true);
+        $this->migrator->add('interaction_management.is_type_required', true);
```
Orrison: If you take a look at the other settings migrations. Each of these adds unfortunately needs to be surrounded by a `try/catch` catching the exception that it already exists so that it doesn't break the migration if someone went to the settings page and saved before this ran.

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
```diff
@@ -49,10 +59,86 @@
...
+
+    public function mount(): void
+    {
+        if (InteractionMetadataFeature::active()) {
+            $settings = $this->getSettings();
+
+            $this->data = [
+                'is_driver_enabled' => $settings->is_driver_enabled,
+                'is_driver_required' => $settings->is_driver_required,
+            ];
```
danharrin: Just to be safe, you should use `$this->form->fill()` instead of directly setting the data property to ensure all the hooks run

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
```diff
@@ -49,10 +59,86 @@
 use Filament\Tables\Table;
 use Illuminate\Database\Eloquent\Builder;
 
-class ListInteractionDrivers extends ListRecords
+class ListInteractionDrivers extends ListRecords implements HasForms
 {
+    use InteractsWithForms;
```
danharrin: I think the base page class has this interface and trait already

danharrin (changes requested):

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
payal-canyon: Hey @danharrin !
I understand your point, but if we remove both `HasForms` and `InteractsWithForms`, the following error occurs:

`AdvisingApp\Interaction\Filament\Resources\InteractionInitiativeResource\Pages\ListInteractionInitiatives::form(): 
Argument #1 ($form) must be of type Filament\Forms\Form, Filament\Infolists\Infolist given, 
called in /var/www/html/vendor/filament/infolists/src/Concerns/InteractsWithInfolists.php on line 55`

From what I see, `ListRecords` extends `Page`, and `Page` extends `BasePage.` This means we can safely remove `HasForms`, but we still need to explicitly add `InteractsWithForms`. Without it, the above error occurs. When I add `InteractsWithForms`, the issue is **resolved**.

if you still want me to remove `InteractsWithForms`? If so, what would be the recommended way to handle this error?

Thank You!

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
```diff
@@ -49,10 +59,100 @@
...
+    public function mount(): void
+    {
+        if (InteractionMetadataFeature::active()) {
+            $this->fillForm();
+        }
+    }
+
+    public function fillForm(): void
+    {
+        $this->callHook('beforeFill');
```
danharrin: You don't need to call these hooks, they are just for the Filament framework really :)

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
danharrin: Ah that's weird, don't worry about it for now then.

app-modules/interaction/src/Filament/Resources/InteractionDriverResource/Pages/ListInteractionDrivers.php
payal-canyon: @danharrin  Oops! I thought we needed to include those as well. Out of curiosity, I have a question: I noticed that in other pages like `AmazonS3` and similar ones, we call `$this->callHook()`. Could you explain why it’s being called in those files within the Filament framework?

danharrin: Yeah, those pages shouldn't have them either :) Its for this feature, but we don't need it on custom pages - https://filamentphp.com/docs/4.x/resources/editing-records#lifecycle-hooks

PR #1842, opened 2025-09-02
[ADVAPP-1830]: Introduce the ability to generate an image inline in the institutional advisor and custom advisors by selecting an image button that can turned on or off

danharrin:

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1830

### Technical Description

Uses the non-streaming Responses API with the "image generation" tool to generate images based on a prompt. The feature can be enabled per-model by providing an image generation model within the same region as the text deployment. It also needs to be enabled per-chat using the new button in the UI.

No

No

### DIFF
--- a/app-modules/ai/src/Jobs/Advisors/RetryAdvisorMessage.php
+++ b/app-modules/ai/src/Jobs/Advisors/RetryAdvisorMessage.php
@@ -68,6 +68,7 @@ class RetryAdvisorMessage implements ShouldQueue
         protected AiThread $thread,
         protected string $content,
         protected array $files = [],
+        protected bool $hasImageGeneration = false,
     ) {}
 
     public function handle(): void
@@ -101,7 +102,8 @@ class RetryAdvisorMessage implements ShouldQueue
         try {
             $stream = $aiService->retryMessage(
                 message: $message,
-                files: $this->files
+                files: $this->files,
+                hasImageGeneration: $this->hasImageGeneration,
             );
         } catch (Throwable $exception) {
             report($exception);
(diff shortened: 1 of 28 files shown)


app-modules/ai/src/Jobs/Advisors/RetryAdvisorMessage.php
```diff
@@ -132,6 +137,14 @@ public function handle(): void
                 continue;
             }
 
+            if ($chunk instanceof Image) {
+                $media = $this->thread->addMediaFromBase64($chunk->content)
+                    ->usingFileName(Str::random() . '.' . $chunk->format)
+                    ->toMediaCollection('generated_images', diskName: 's3-public');
```
Orrison, line 143: Let's check with Product in the ticket if there is going to be an issue with these being stored in a public bucket.

app-modules/ai/src/Jobs/Advisors/SendAdvisorMessage.php
```diff
@@ -153,6 +158,14 @@ public function handle(): void
                 continue;
             }
 
+            if ($chunk instanceof Image) {
+                $media = $this->thread->addMediaFromBase64($chunk->content)
+                    ->usingFileName(Str::random() . '.' . $chunk->format)
+                    ->toMediaCollection('generated_images', diskName: 's3-public');
```
Orrison, line 164: Let's check with Product in the ticket if there is going to be an issue with these being stored in a public bucket.

Orrison (changes requested):
This all looks good. But as discussed please check with Product about the image visibility in the comments of the ticket. If this works fine then please re-request my review. If it doesn't let me know.

PR #1843, opened 2025-09-03
[ADVAPP-1760]: Some users are experiencing issues when configuring email templates for case types

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1760

### Technical Description

> Some users are experiencing issues when configuring email templates for case types

> No

> No

### DIFF
--- a/app-modules/case-management/src/Filament/Resources/CaseTypeResource/Pages/ManageCaseTypeEmailTemplate.php
+++ b/app-modules/case-management/src/Filament/Resources/CaseTypeResource/Pages/ManageCaseTypeEmailTemplate.php
@@ -52,6 +52,7 @@ use Filament\Resources\Pages\EditRecord;
 use FilamentTiptapEditor\TiptapEditor;
 use Livewire\Attributes\Computed;
 use Livewire\Attributes\Locked;
+use Log;
 
 /** @property-read ?CaseTypeEmailTemplate $template */
 class ManageCaseTypeEmailTemplate extends EditRecord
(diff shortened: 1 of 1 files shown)


app-modules/case-management/src/Filament/Resources/CaseTypeResource/Pages/ManageCaseTypeEmailTemplate.php
```diff
@@ -164,7 +164,7 @@ protected function getEmailTemplateFormSchema(CaseTypeEmailTemplateRole $role):
                     'description',
                     'type',
                 ])
-                ->blocks($role === CaseTypeEmailTemplateRole::Customer ? [] : [
+                ->blocks([
```
danharrin, line 167: Can you please explain what the error was and why this was the correct fix?

danharrin (changes requested):

app-modules/case-management/src/Filament/Resources/CaseTypeResource/Pages/ManageCaseTypeEmailTemplate.php
jayushi-canyon: @danharrin When previously i created this functionality I got confused in a functionality and this conditions for customers. But now when I cross verified again with Aiding app feature I realised that this condition was not necessary and it was causing issue.

The issue was that in the seed file for case_type_email_template table buttonblocks were added in DB, but when creating the form due to this conditions blocks were missing in frontend and conditions were not matched which was throwing error.

danharrin (approved):

PR #1846, opened 2025-09-03
[ADVAPP-1857]: Enable integration with stock media sources to support visual enhancements

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1857

### Technical Description

Create Stock Media module and settings, and a page to manage said settings.

No

StockMediaFeature feature flag

### DIFF
--- a/app-modules/stock-media/src/Filament/Pages/ManageStockMediaSettings.php
+++ b/app-modules/stock-media/src/Filament/Pages/ManageStockMediaSettings.php
@@ -84,7 +84,7 @@ class ManageStockMediaSettings extends SettingsPage
 
     protected function mutateFormDataBeforeSave(array $data): array
     {
-      if (filled($data['provider'] ?? null)) {
+        if (filled($data['provider'] ?? null)) {
             $data['provider'] = StockMediaProvider::parse($data['provider']);
         }
 


app-modules/stock-media/src/Settings/StockMediaSettings.php
```diff
@@ -0,0 +1,63 @@
...
+{
+    public bool $active = false;
+
+    public ?string $provider = null;
+
+    public ?string $pexels_api_key = null;
+
+    public static function group(): string
+    {
+        return 'stock-media';
```
danharrin: Please make the settings group use `_` instead of `-` like most of the rest of them (I know there are some inconsistencies in some)

app-modules/stock-media/src/Settings/StockMediaSettings.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace AdvisingApp\StockMedia\Settings;
+
+use Spatie\LaravelSettings\Settings;
+
+class StockMediaSettings extends Settings
+{
+    public bool $active = false;
+
+    public ?string $provider = null;
```
danharrin: Should the type of this one be an enum instead of a string?

Also, do we need `active`? If there is no provider selected, I guess that is the "inactivate" state covered already.

rebecca-canyon: Having `active` as a setting was one of the requirements from Product. So while you're right that this could be made to work without it easily, I think it should stay.

danharrin: Sorry, I did check but missed that originally.

Can you rename it to `is_active` please? The label in the UI should still be `Active` though

app-modules/stock-media/src/Filament/Pages/ManageStockMediaSettings.php
```diff
@@ -0,0 +1,86 @@
...
+            ->schema([
+                Toggle::make('is_active')
+                    ->label('Active')
+                    ->inline(false)
+                    ->live(),
+                Select::make('provider')
+                    ->options(StockMediaProvider::class)
+                    ->enum(StockMediaProvider::class)
+                    ->visible(fn (Get $get) => $get('is_active'))
+                    ->afterStateHydrated(fn (Select $component) => $component->state(StockMediaProvider::Pexels))
```
danharrin: Can you explain why this line is included please?

rebecca-canyon: That's to ensure that the Pexels option is selected, since the field is currently disabled.

danharrin: Hmm I'm not sure why we should disable it in this case. Once the user selects "active" I think it's quite normal for them to then select "Pexels" before seeing the API key field.

app-modules/stock-media/src/Filament/Pages/ManageStockMediaSettings.php
```diff
@@ -0,0 +1,86 @@
...
+                    ->disabled(true)
+                    ->dehydrated(),
+                TextInput::make('pexels_api_key')
+                    ->label('API Key')
+                    ->visible(fn (Get $get) => $get('is_active') && $get('provider') === StockMediaProvider::Pexels)
```
danharrin: I think you also need to account for `$get('provider')` to be the `StockMediaProvider::Pexels->value` as well as `StockMediaProvider::Pexels`, as that can happen when the enum select is used

danharrin (changes requested):

app-modules/stock-media/src/Filament/Pages/ManageStockMediaSettings.php
rebecca-canyon: Fair; I had it in my head that having the field both preselected and disabled had been asked for, but I can't find any evidence of that now.

danharrin (changes requested):
When testing this, activating the integration, and then setting the provider, I got this error. I think you need to use the `mutateFormDataBeforeSave()` hook on the page to ensure that it is an enum instance before saving

Image

danharrin (approved):

PR #1858, opened 2025-09-05
[ADVAPP-1871]: Some users have reported that multiple workflow steps do not happen relative to each other as expected

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1871

### Technical Description

> Some users have reported that multiple workflow steps do not happen relative to each other as expected

> No

> Yes. Feature Flag: `WorkflowSequentialExecutionFeature`


app-modules/workflow/database/migrations/2025_09_05_133146_add_offset_minutes_and_previous_workflow_step_id_to_workflow_run_steps_table.php
danharrin, line 1: Can you rename `offset_minutes` to `delay_minutes` to match the `workflow_steps` table please?

app-modules/application/src/Listeners/TriggerApplicationSubmissionWorkflows.php
```diff
@@ -98,20 +104,15 @@ public function handle(ApplicationSubmissionCreated $event): void
         }
     }
 
-    private function getStepScheduledAt(WorkflowStep $workflowStep, ApplicationSubmissionCreated $event): Carbon
+    private function getStepScheduledAt(WorkflowStep $workflowStep, ApplicationSubmissionCreated $event, int $index): ?Carbon
```
danharrin: Instead of passing the index in, let's avoid calling the function altogether unless the index is 0?

app-modules/workflow/src/Concerns/SchedulesNextWorkflowStep.php
```diff
@@ -0,0 +1,51 @@
...
+use AdvisingApp\Workflow\Services\SequentialWorkflowStepScheduler;
+
+trait SchedulesNextWorkflowStep
+{
+    protected function markStepCompletedAndScheduleNext(WorkflowRunStep $workflowRunStep): void
+    {
+        $workflowRunStep->succeeded_at = now();
+        $workflowRunStep->saveOrFail();
+
+        app(SequentialWorkflowStepScheduler::class)->scheduleNextSteps($workflowRunStep);
```
danharrin: Considering `SequentialWorkflowStepScheduler` is just used here, can we just move everything it does into this method instead? And then remove the extra class?

danharrin (changes requested):
Can you introduce a feature flag for the new columns please?

app-modules/workflow/src/Concerns/SchedulesNextWorkflowStep.php
```diff
@@ -0,0 +1,62 @@
...
+        $workflowRunStep->saveOrFail();
+
+        $nextSteps = WorkflowRunStep::query()
+            ->where('workflow_run_id', $workflowRunStep->workflow_run_id)
+            ->where('previous_workflow_run_step_id', $workflowRunStep->id)
+            ->whereNull('dispatched_at')
+            ->get();
+
+        foreach ($nextSteps as $nextStep) {
+            $executeAt = Carbon::now()->addMinutes($nextStep->delay_minutes);
```
danharrin: For consistency with line 46, maybe make this a `now()` instead of `Carbon::now()`

app-modules/workflow/src/Jobs/CareTeamWorkflowActionJob.php
```diff
@@ -95,6 +99,13 @@ public function handle(): void
                     $workflowRunStepRelated->save();
                 });
 
+            if (WorkflowSequentialExecutionFeature::active()) {
+                $this->markStepCompletedAndScheduleNext($this->workflowRunStep);
```
danharrin: Every time this method is called, you pass in `$this->workflowRunStep`. Maybe the param can be removed, and you just access `$this->workflowRunStep` from inside the trait method? Then there is less code to write for each action job.

danharrin (approved):

PR #1861, opened 2025-09-05
[ADVAPP-1886]: Introduce a four step wizard for logging a new interaction for students and prospects

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1886

### Technical Description

Change interaction creation to use wizard for individual interactions, bulk actions, and campaign journey steps.

No

No

### DIFF
--- a/app-modules/interaction/src/Filament/Resources/InteractionResource/Pages/CreateInteraction.php
+++ b/app-modules/interaction/src/Filament/Resources/InteractionResource/Pages/CreateInteraction.php
@@ -60,10 +60,12 @@ use Filament\Forms\Components\MorphToSelect;
 use Filament\Forms\Components\Select;
 use Filament\Forms\Components\Textarea;
 use Filament\Forms\Components\TextInput;
+use Filament\Forms\Components\Wizard\Step;
 use Filament\Forms\Form;
 use Filament\Forms\Get;
 use Filament\Forms\Set;
 use Filament\Resources\Pages\CreateRecord;
+use Filament\Resources\Pages\CreateRecord\Concerns\HasWizard;
 use Filament\Resources\Pages\ManageRelatedRecords;
 use Filament\Resources\RelationManagers\RelationManager;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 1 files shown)


app-modules/interaction/src/Filament/Resources/InteractionResource/Pages/CreateInteraction.php
```diff
@@ -269,4 +251,27 @@ private function getSettings(): InteractionManagementSettings
 
         return $this->settings;
     }
+
+    private function getEndDateTime(Get $get, Set $set): void
```
danharrin: Can you keep this named `calculateEndDateTime` since `get` methods usually return something, and this one actually has an effect

app-modules/interaction/src/Filament/Resources/InteractionResource/Pages/CreateInteraction.php
```diff
@@ -60,205 +60,187 @@
...
+                        ->live(onBlur: true)
+                        ->afterStateUpdated(fn (Get $get, Set $set) => $this->getEndDateTime($get, $set)),
+                    TextInput::make('duration')
+                        ->label('Duration (Minutes)')
+                        ->integer()
+                        ->minValue(0)
+                        ->required()
+                        ->dehydrated(false)
+                        ->live(onBlur: true)
+                        ->afterStateUpdated(fn (Get $get, Set $set) => $this->getEndDateTime($get, $set)),
```
danharrin: Does this work and pass the parameters through correctly?

```suggestion
                        ->afterStateUpdated($this->calculateEndDateTime(...)),
```

rebecca-canyon: From my manual testing, it does.

danharrin: I think that is probably the route we should go to avoid the parameter repetition then  👍

rebecca-canyon: Ah, apologies, I completely misread what you had said.

In trying this morning, I haven't had much success with that; I could refactor `calculateEndDateTime` to take the specific keys that it needs and return `end_datetime` and set that within the `afterStateUpdated` closure, but I don't think I can just have the function call. If nothing else, I get a linting error for using the results of a void method.

PR #1865, opened 2025-09-09
[ADVAPP-1887]: Introduce a three step wizard for sending a new message

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1887

### Technical Description

Change the relationship manager on an educatable's holistic view page, the actions from clicking an educatable's email or phone number, and the creation of a new message in the unified inbox to use a wizard.

No

No


app-modules/engagement/src/Filament/Actions/SendEngagementAction.php
```diff
@@ -112,227 +113,231 @@ protected function setUp(): void
...
-                                $educatable = match ($get('recipient_type')) {
-                                    'student' => Student::find($get('recipient_id')),
-                                    'prospect' => Prospect::find($get('recipient_id')),
-                                    default => null,
-                                };
-
-                                if (! $educatable?->phoneNumbers()->where('can_receive_sms', true)->exists()) {
-                                    $set('channel', 'email');
-                                }
+            ->form(function (Form $form) {
```
danharrin: I think you should use `steps()` here

app-modules/engagement/src/Filament/Actions/RelationManagerSendEngagementAction.php
```diff
@@ -327,16 +340,7 @@ protected function setUp(): void
             ->modalCloseButton(false)
             ->closeModalByClickingAway(false)
             ->closeModalByEscaping(false)
-            ->createAnother(false)
-            ->modalCancelAction(false)
-            ->extraModalFooterActions([
```
danharrin, line 332: I think this custom "cancel" action needs to stay somehow, otherwise is the TipTap editor bug still there when you close and reopen the modal?

app-modules/engagement/src/Filament/Actions/RelationManagerSendEngagementAction.php
```diff
@@ -231,13 +244,13 @@ protected function setUp(): void
                             ->hidden(fn (Get $get): bool => $get('channel') === NotificationChannel::Sms->value)
                             ->helperText('You can insert recipient or your information by typing {{ and choosing a merge value to insert.')
                             ->columnSpanFull(),
-                        EngagementSmsBodyInput::make(context: 'create', form: $form),
```
danharrin, line 234: Have a look inside the `EngagementSmsBodyInput` and see how `$form` is just used to determine the value of `showMergeTagsInBlocksPanel`? I think we should still pass it in if we can

danharrin (changes requested):

app-modules/engagement/src/Filament/Actions/RelationManagerSendEngagementAction.php
rebecca-canyon: I tried a couple of times and I wasn't able to recreate that bug on my end after removing the custom cancel action. I was also unsuccessful in getting the `extraModalFooterActions` to work with the wizard, or with trying to overwrite the behavior of either the close or cancel buttons.

app-modules/engagement/src/Filament/Actions/RelationManagerSendEngagementAction.php
rebecca-canyon: If `$form` is provided, it returns the opposite of whether or not it's a `RelationManager`. And since this is a `RelationManager`, it would just return false anyways.

app-modules/engagement/src/Filament/Actions/SendEngagementAction.php
rebecca-canyon: Using `steps()` here breaks sending an engagement from an individual view page when clicking on their email / phone number. Something about the `EducatableSelect` I think?

danharrin:
Updated the labels from "Send your email or text" to "Send your message" to reduce consumed space in the wizard header so it doesn't start scrolling.

PR #1870, opened 2025-09-09
[ADVAPP-1736]: Improve student holistic profile by displaying the SMS Opt Out field in the student's profile information

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1736

### Technical Description

> Improve student holistic profile by displaying the SMS Opt Out field in the student's profile information

> No

> Yes. Feature Flag: 


app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/EditStudent.php
```diff
@@ -191,6 +199,32 @@ public function form(Form $form): Form
...
+                                        $record = $component->getContainer()->getRecord();
+
+                                        if ($record instanceof StudentPhoneNumber && $record->exists) {
+                                            $key = $record->id;
+                                        } else {
+                                            $key = $component->getContainer()->getStatePath();
+                                        }
+
+                                        $this->smsOptOutStates[$key] = (bool) $state;
+                                    })
```
danharrin: Instead of doing this logic here, I would probably use the `mutateRelationshipData` hooks before fill and before save. https://filamentphp.com/docs/3.x/forms/fields/repeater#mutating-related-item-data-before-filling-the-field

Before fill, we can set the value to be true or false based on if the number is in the opted out table.

Before save, we can save the number to the table, or delete it from the table if it has changed.

Then we can remove the `smsOptOutStates` property for tracking this, since the information is already available in the `$data` of each hook method.

app-modules/student-data-model/src/Actions/ToggleSmsOptOut.php
```diff
@@ -0,0 +1,59 @@
...
+    public function execute(StudentPhoneNumber $phoneNumber, bool $optOut): void
+    {
+        $e164Number = PhoneNumberNormalizer::toE164($phoneNumber->number);
+
+        if (! $e164Number) {
+            return;
+        }
+
+        if ($optOut) {
+            SmsOptOutPhoneNumber::updateOrCreate(['number' => $e164Number]);
```
danharrin: If the record already exists, I would probably not update it, since that would affect the timestamp. We can do an `exists()` check before creating it to avoid this.

app-modules/engagement/src/Filament/Actions/SendEngagementAction.php
```diff
@@ -184,6 +256,15 @@ protected function setUp(): void
                                         NotificationChannel::Sms => $educatable?->phoneNumbers()
                                             ->where('can_receive_sms', true)
                                             ->get()
+                                            ->filter(function (StudentPhoneNumber | ProspectPhoneNumber $phoneNumber) {
```
danharrin: Instead of filtering after the phone numbers have been fetched, you could use a `whereDoesntHave()` to scope the results to only those that aren't related to an opt out record.

The opt out records could be associated with a phone number using a `BelongsTo` Eloquent relationship, using `number` as the foreign key.

app-modules/student-data-model/src/Support/PhoneNumberNormalizer.php
```diff
@@ -0,0 +1,70 @@
...
+namespace AdvisingApp\StudentDataModel\Support;
+
+use App\Settings\ImportSettings;
+use libphonenumber\NumberParseException;
+use libphonenumber\PhoneNumberFormat;
+use libphonenumber\PhoneNumberUtil;
+
+class PhoneNumberNormalizer
+{
+    public static function toE164(string $phoneNumber): ?string
```
danharrin: Please correct me if I'm wrong, but I think we already have a validation rule so that only E164 phone numbers can be stored already. In which case, I don't think we need to normalise the numbers before checking if they exist in the opt out table. That will allow us to use relationships to check if numbers are opted out too, which will be faster and reduce N+1 queries.

danharrin (changes requested):

danharrin (dismissed):
Apologies, a couple of PRs have changed the actions since you last pushed, and there are now merge conflicts

app-modules/student-data-model/src/Filament/Resources/StudentResource/Pages/CreateStudent.php
```diff
@@ -210,7 +212,25 @@ public function form(Form $form): Form
                                     ->label('Can receive SMS messages')
                                     ->columnSpanFull()
                                     ->default(true),
+                                Checkbox::make('sms_opt_out_phone_number')
```
danharrin, line 214: One small thing, I would put this checkbox on the same row as the "Can recieve" checkbox so there is less vertical space consumed

danharrin (approved):

PR #1872, opened 2025-09-09
[ADVAPP-1791]: Introduce a workflow step type task for forms and admissions

jayushi-canyon:
- https://canyongbs.atlassian.net/browse/ADVAPP-1791

### Technical Description

> Introduce a workflow step type task for forms and admissions

> No

> No


app-modules/task/src/Notifications/TaskAssignedToUserNotification.php
```diff
@@ -56,7 +57,7 @@ class TaskAssignedToUserNotification extends Notification implements ShouldQueue
     use Queueable;
 
     public function __construct(
-        public Task $task,
+        public WorkflowTaskDetails| Task $task,
```
danharrin: Can you please explain why this needs to accept a new model? Once a workflow step runs, it should create a Task, and then that task should be used to send the notification.

danharrin (approved):

PR #1875, opened 2025-09-10
Km/advapp 1890

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-

### Technical Description

>

>

>


app-modules/ai/database/migrations/2025_09_10_174316_create_confidential_prompt_teams_table.php
```diff
@@ -0,0 +1,61 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('confidential_prompt_teams', function (Blueprint $table) {
+            $table->id();
```
Orrison: We use uuids even for pivots

app-modules/ai/database/migrations/2025_09_10_174316_create_confidential_prompt_teams_table.php
```diff
@@ -0,0 +1,61 @@
...
+
+            $table->foreignUuid('prompt_id')
+                ->constrained('prompts');
+
+            $table->foreignUuid('team_id')
+                ->constrained('teams');
```
Orrison: Since this is a pivot you can make both of these cascade on delete

app-modules/ai/database/migrations/2025_09_10_174324_create_confidential_prompt_users_table.php
```diff
@@ -0,0 +1,61 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('confidential_prompt_users', function (Blueprint $table) {
+            $table->id();
```
Orrison: We use uuids even for pivots

app-modules/ai/database/migrations/2025_09_10_174324_create_confidential_prompt_users_table.php
```diff
@@ -0,0 +1,61 @@
...
+
+            $table->foreignUuid('prompt_id')
+                ->constrained('prompts');
+
+            $table->foreignUuid('user_id')
+                ->constrained('users');
```
Orrison: Since this is a pivot you can make both of these cascade on delete.

app-modules/ai/src/Filament/Resources/PromptResource/Pages/CreatePrompt.php
```diff
@@ -82,7 +86,31 @@ public function form(Form $form): Form
...
+                                ->relationship('confidentialAccessTeams', 'name')
+                                ->preload()
+                                ->label('Teams')
+                                ->multiple()
+                                ->exists('teams', 'id')
+                                ->visible(fn (Get $get) => $get('is_confidential')),
+                        ])
+                            ->columns(2)
+                            ->columnSpanFull()
+                            ->visible(fn (Get $get) => ConfidentialPromptsFeature::active() && $get('is_smart') == false),
```
Orrison:
```suggestion
                            ->visible(fn (Get $get) => ConfidentialPromptsFeature::active() && $get('is_smart') === false),
```

app-modules/ai/src/Filament/Resources/PromptResource/Pages/EditPrompt.php
```diff
@@ -87,6 +91,32 @@ public function form(Form $form): Form
                             ])
                             ->grouped()
```
Orrison, line 94: You may want to make this `live` like you did in `create`

app-modules/ai/src/Filament/Resources/PromptResource/Pages/EditPrompt.php
```diff
@@ -87,6 +91,32 @@ public function form(Form $form): Form
...
+                                ->multiple()
+                                ->exists('teams', 'id')
+                                ->visible(fn (Get $get) => $get('is_confidential')),
+                        ])
+                            ->columns(2)
+                            ->columnSpanFull()
+                            ->visible(fn (Get $get) => ConfidentialPromptsFeature::active()
+                                && auth()->user()->hasRole(Authenticatable::SUPER_ADMIN_ROLE)
+                                || auth()->user()->id == $get('user_id')
+                                && $get('is_smart') == false),
```
Orrison: Please make a habit of using strict comparison. It is only niche cases where loose comparison is necessary imo.

```suggestion
                                || auth()->user()->id === $get('user_id')
                                && $get('is_smart') === false),
```

app-modules/ai/src/Filament/Resources/PromptResource/Pages/EditPrompt.php
```diff
@@ -87,6 +91,32 @@ public function form(Form $form): Form
...
+                                ->label('Teams')
+                                ->multiple()
+                                ->exists('teams', 'id')
+                                ->visible(fn (Get $get) => $get('is_confidential')),
+                        ])
+                            ->columns(2)
+                            ->columnSpanFull()
+                            ->visible(fn (Get $get) => ConfidentialPromptsFeature::active()
+                                && auth()->user()->hasRole(Authenticatable::SUPER_ADMIN_ROLE)
+                                || auth()->user()->id == $get('user_id')
```
Orrison: I am not sure this `$get('user_id')` call works? `$get` only retrieves values from this form, not from the record. You will probably need to load in the record and check with that.

You should have tests for this functionality and access, which would have caught this.

app-modules/ai/src/Models/Scopes/ConfidentialPromptScope.php
```diff
@@ -0,0 +1,66 @@
...
+                        ->orWhereHas('confidentialAccessTeams', function (Builder $query) {
+                            $query->whereHas('users', function (Builder $query) {
+                                $query->where('users.id', auth()->id());
+                            });
+                        })
+                        ->orWhereHas('confidentialAccessUsers', function (Builder $query) {
+                            $query->where('users.id', auth()->id());
+                        });
+                });
+        });
```
Orrison, line 64: This seems alright but will need to be updated for `Project`s and should be tested to ensure it is working.

app-modules/ai/src/Models/ConfidentialPromptTeam.php
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+namespace AdvisingApp\Ai\Models;
+
+use AdvisingApp\Team\Models\Team;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\Pivot;
+
+class ConfidentialPromptTeam extends Pivot
+{
```
Orrison, line 45: This should have UUIDs

app-modules/ai/src/Models/ConfidentialPromptUser.php
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+namespace AdvisingApp\Ai\Models;
+
+use App\Models\User;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\Pivot;
+
+class ConfidentialPromptUser extends Pivot
+{
```
Orrison, line 45: This should have UUIDs

Orrison (changes requested):
This is also missing the ability to associate `Project`s to have confidential access.

app-modules/ai/src/Models/Scopes/ConfidentialPromptScope.php
Orrison: Per our conversation this morning, no need to deal with Project right now.

But please add a test for this global scope ensuring that it works.

Orrison (changes requested):
One note.

It could also be good to add protection at the observer level of a prompt during `updating` to throw an exception if there is an authed user who is not the creator or super admin. To protect against anyone changing the value that shouldn't

app-modules/ai/src/Observers/PromptObserver.php
```diff
@@ -37,11 +37,22 @@
...
     }
+
+    public function updating(Prompt $prompt): void
+    {
+        if (auth()->user()->id === $prompt->user_id || auth()->user()->hasRole(Authenticatable::SUPER_ADMIN_ROLE)) {
+            return;
+        }
+
+        throw new Exception('You do not have permission to update this prompt.');
+    }
```
Orrison: This is not what I was intending. You would need to first check if there is an authed User.

If there is then check if any of the confidentiality attributes are being changed (dirty) if they are then throw an exception if the user is not someone that can do so.

Given that this is not being done elsewhere and just would have been a nice to have and it is near deployment. Please just remove / revert this.

Orrison (changes requested):

PR #1878, opened 2025-09-11
[ADVAPP-1779]: Workflow Step editing does not load in current data, making editing not possible

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1779

### Technical Description

> Workflow Step editing does not load in current data, making editing not possible.
This PR fixes issues with Automatic Assignment in Case workflows for both creation and editing.

> CaseBlock.php
Creation Fix: In transformAssignmentData(), 'automatic' is converted to null before saving, preventing errors when creating workflows.
Editing Fix: generateFields() and editFields() now include 'automatic' in the dropdown when applicable, ensuring consistent behavior in edit forms.

> WorkflowStepsRelationManager.php
Editing Fix: fillForm() reconstructs edit form data (splitting delay_minutes) and calls beforeUpdate(), which applies the assignment fix during edits.

> WorkflowCaseDetails.php
Creation/Editing Fix: In toArray(), if no user is assigned and assignment type is not None, 'assigned_to_id' is set to 'automatic', so the edit form shows the correct value.

> No

> No

### DIFF
--- a/app-modules/workflow/src/Filament/Blocks/WorkflowActionBlock.php
+++ b/app-modules/workflow/src/Filament/Blocks/WorkflowActionBlock.php
@@ -103,7 +103,5 @@ abstract class WorkflowActionBlock extends Block
      * Prepare WorkflowDetails for form population during editing.
      * Allows blocks to load necessary relationships before toArray() is called.
      */
-    public function prepareForEdit(WorkflowDetails $details): void
-    {
-    }
+    public function prepareForEdit(WorkflowDetails $details): void {}
 }


payal-canyon:
**Note:** In this PR, I’m also fixing a bug related to `Automatic Assignment` in `Case` workflows. When creating a workflow and selecting `Automatic Assignment` in the `Assign Case To` dropdown, saving after filling in the details results in an error. The same issue occurs while editing. This PR includes a fix for that as well.

Thank You!

app-modules/workflow/src/Filament/Resources/WorkflowResource/RelationManagers/WorkflowStepsRelationManager.php
```diff
@@ -167,6 +167,10 @@ public function table(Table $table): Table
                     ->fillForm(function (WorkflowStep $record): array {
                         assert($record->currentDetails instanceof WorkflowDetails);
 
+                        if ($record->currentDetails instanceof WorkflowCaseDetails) {
```
danharrin: Ideally, logic related to specific blocks (such as Case) should not leak into the general workflow module code. Ideally, classes that are specific to that block such as `CaseBlock` should handle this. Do we need to introduce new methods to the Block interface and then call them from here, so that blocks are able to hook into the saving process?

app-modules/workflow/src/Filament/Blocks/CaseBlock.php
```diff
@@ -227,4 +229,45 @@ public static function type(): string
...
+     * @return array<string, mixed>
+     */
+    public function beforeUpdate(array $data): array
+    {
+        return $this->transformAssignmentData($data);
+    }
+
+    public function prepareForEdit(WorkflowDetails $details): void
+    {
+        if ($details instanceof WorkflowCaseDetails) {
```
danharrin: Is there a situation where this is `false`? If not, `assert()` is probably better

danharrin (changes requested):

danharrin (approved):

PR #1895, opened 2025-09-18
[ADVAPP-1921]: Introduce insert stock photo tip tap extension

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1921

### Technical Description

Adds a new tool to editors for inserting stock images. Introduces an API endpoint to search for these using the configured integration.

No

No

### DIFF
--- a/app-modules/engagement/src/Filament/Resources/EmailTemplateResource/Pages/CreateEmailTemplate.php
+++ b/app-modules/engagement/src/Filament/Resources/EmailTemplateResource/Pages/CreateEmailTemplate.php
@@ -71,7 +71,7 @@ class CreateEmailTemplate extends CreateRecord
                         'recipient email',
                         'recipient preferred name',
                     ])
-                    ->profile('email')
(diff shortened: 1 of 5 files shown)


app-modules/stock-media/routes/web.php
```diff
@@ -0,0 +1,40 @@
...
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use AdvisingApp\StockMedia\Http\Controllers\StockImagesController;
+use Illuminate\Support\Facades\Route;
+
+Route::post('/api/stock-images', StockImagesController::class)->middleware(['signed'])->name('api.stock-images');
```
Orrison, line 40: This has a route prefix of `api`, yet is in the `web` routes file. Is there a reason for that and are we sure that's what we want to do?

Is there any reason this can't be in `api.php` in this same directory?

Orrison: Discussed in a call. All good

PR #1901, opened 2025-09-19
[ADVAPP-1945]: Introduce a Projects report placeholder under Project Management in the report library

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1945

### Technical Description

> Introduce a Projects report placeholder under Project Management in the report library

> No

> No


Orrison (changes requested):
Hmm, it seems we keep mixing back and forth changing the sort order to be fully sequential rather then by 10s.

We never officially made any decision to change anything so let's make sure navigation sort is done via increments of 10.

PR #1919, opened 2025-09-25
[ADVAPP-1959]: Improve the unified inbox by adding an status as to whether the message is from a prospect or a student

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1959

### Technical Description

Add relations column to unified inbox to display if the sender/recipient is a student or prospect

No

No

### DIFF
--- a/app-modules/engagement/src/Filament/Pages/Inbox.php
+++ b/app-modules/engagement/src/Filament/Pages/Inbox.php
@@ -49,10 +49,8 @@ use Filament\Tables\Actions\ViewAction;
 use Filament\Tables\Columns\TextColumn;
 use Filament\Tables\Concerns\InteractsWithTable;
 use Filament\Tables\Contracts\HasTable;
-use Filament\Tables\Filters\Filter;
 use Filament\Tables\Filters\SelectFilter;
 use Filament\Tables\Table;
-use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Str;
 
 class Inbox extends Page implements HasTable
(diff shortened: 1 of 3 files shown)


app-modules/engagement/src/Filament/Pages/SentItems.php
```diff
@@ -94,6 +94,9 @@ public function table(Table $table): Table
                 TextColumn::make('status')
                     ->badge()
                     ->state(fn (Engagement $record) => EngagementDisplayStatus::getStatus($record)),
+                TextColumn::make('recipient_type')
+                    ->label('Relation')
+                    ->formatStateUsing(fn (Engagement $record) => ucwords($record->recipient_type)),
```
Orrison: As the ticket is written I think there is an chance this isn't being asked for in the Sent tab.

I believe it was just asking for the Inbox itself, especially given that it says "from a student or prospect"

app-modules/engagement/src/Filament/Pages/Inbox.php
```diff
@@ -92,6 +92,9 @@ public function table(Table $table): Table
                     ->badge(),
                 TextColumn::make('status')
                     ->badge(),
+                TextColumn::make('sender_type')
+                    ->label('Relation')
+                    ->formatStateUsing(fn (EngagementResponse $record) => ucwords($record->sender_type)),
```
Orrison: This is missing being sortable as well as this table is missing a filter being added as requested.

Please also add a test for the filter once it is done, I don't think a test file exists for the class yet so you will need to add one. But feel free to only add the test for your filter, other tests could be added in scope later.

Orrison (changes requested):

Orrison (approved):

PR #1935, opened 2025-09-30
[ADVAPP-1960]: Introduce hyperlinking of student and prospect names in the unified inbox to enable direct navigation to individual records

krlmrr:
### Ticket(s) or GitHub Issue
- https://canyongbs.atlassian.net/browse/ADVAPP-1960

### Technical Description
- Add a redirect to the from column.

- No 

- No

### DIFF
--- a/app-modules/engagement/resources/views/filament/pages/view-engagement-response.blade.php
+++ b/app-modules/engagement/resources/views/filament/pages/view-engagement-response.blade.php
@@ -9,22 +9,22 @@
     Notice:
 
     - You may not provide the software to third parties as a hosted or managed
-    service, where the service provides users with access to any substantial set of
-    the features or functionality of the software.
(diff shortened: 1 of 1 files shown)


app-modules/engagement/src/Filament/Pages/Inbox.php
```diff
@@ -102,6 +104,21 @@ public function table(Table $table): Table
                         return (($record->sender instanceof Student) || ($record->sender instanceof Prospect))
                             ? $record->sender->full_name
                             : null;
+                    })
+                    ->action(function (EngagementResponse $record) {
```
Orrison: I believe you can and probably should use `url()` instead of `action()` here.

That works better for this scenario, is a Filament default functionality, and has other modifications out of the box we can do to it in the future if we wanted.

Orrison (approved):

PR #1969, opened 2025-10-07
[ADVAPP-1780]: Improve performance of workflows for case execution by implementing unit tests

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-1780

### Technical Description

Create factory, test for workflow case steps

No

No

### DIFF
--- a/app-modules/workflow/database/factories/WorkflowCaseDetailsFactory.php
+++ b/app-modules/workflow/database/factories/WorkflowCaseDetailsFactory.php
@@ -58,7 +58,7 @@ class WorkflowCaseDetailsFactory extends Factory
             'division_id' => Division::factory(),
             'status_id' => CaseStatus::factory(),
             'priority_id' => CasePriority::factory(),
-            'assigned_to_id' => fn(array $attributes) => User::factory()->for(Team::factory()->hasAttached(CasePriority::find($attributes['priority_id'])->type, relationship: 'manageableCaseTypes')),
+            'assigned_to_id' => fn (array $attributes) => User::factory()->for(Team::factory()->hasAttached(CasePriority::find($attributes['priority_id'])->type, relationship: 'manageableCaseTypes')),
             'close_details' => $this->faker->sentence(),
             'res_details' => $this->faker->sentence(),
         ];


app-modules/workflow/database/factories/WorkflowCaseDetailsFactory.php
```diff
@@ -0,0 +1,68 @@
...
+     * @return array<string, mixed>
+     */
+    public function definition(): array
+    {
+        $priority = CasePriority::inRandomOrder()->first() ?? CasePriority::factory()->create();
+
+        return [
+            'division_id' => Division::inRandomOrder()->first()->id ?? Division::factory(),
+            'status_id' => CaseStatus::inRandomOrder()->first()->id ?? CaseStatus::factory(),
+            'priority_id' => $priority->getKey(),
```
Orrison: I would like us to start getting rid of these `inRandomOrder()->first()` checks. They were not a good plan when we first started making these.

Can you please replace each of these with JUST the factory call, you also shouldn't need to `create` the factory, just give it the factory and let Laravel handle the creation.

You should then also probably be able to inline priority.

app-modules/workflow/tests/Tenant/CaseWorkflowActionJobTest.php
```diff
@@ -0,0 +1,103 @@
...
+
+    $cases = $student->cases()->get();
+
+    expect($cases)->toHaveCount(1)
+        ->and($cases->first()->division->getKey())->toEqual($division->getKey())
+        ->and($cases->first()->status->getKey())->toEqual($caseStatus->getKey())
+        ->and($cases->first()->priority->getKey())->toEqual($casePriority->getKey())
+        ->and($cases->first()->assignments()->first()->user_id)->toEqual($assignedTo->getKey())
+        ->and($cases->first()->close_details)->toEqual($closeDetails)
+        ->and($cases->first()->res_details)->toEqual($resDetails);
```
Orrison, line 101: I think we should also assert that the case is properly related to the workflow run step via `WorkflowRunStepRelated`

Orrison (changes requested):

app-modules/workflow/database/factories/WorkflowCaseDetailsFactory.php
rebecca-canyon: Priority is pulled out to be used for the `assigned_to_id` creation. So I think that inlining it would make that assignment more complicated / not worth it.

app-modules/workflow/database/factories/WorkflowCaseDetailsFactory.php
```diff
@@ -0,0 +1,68 @@
...
+     */
+    public function definition(): array
+    {
+        $priority = CasePriority::factory()->create();
+
+        return [
+            'division_id' => Division::factory(),
+            'status_id' => CaseStatus::factory(),
+            'priority_id' => $priority->getKey(),
+            'assigned_to_id' => User::factory()->for(Team::factory()->hasAttached($priority->type, relationship: 'manageableCaseTypes')),
```
Orrison: I saw your comment in regards to `$priority` being needed for `assigned_to_id`.

Given that, you still can, and probably should, inline `priority_id` and for `'assigned_to_id'` put it in a closure that accepts `$attributes` so you can pull in the `priority_id` from there, I believe that should resolve to something by then.

PR #1972, opened 2025-10-08
[ADVAPP-2041]: Empty email address values causing errors to be thrown in blade rendering

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2041

### Technical Description

Allow for auth check to pass students

No

No


danharrin (changes requested):
Only prospects need to be passed into the policy method, not students. If a student is passed in, we don't use it, so we may as well pass null as we did before.

What is the object that we call `getKey()` on? It doesn't look like the original issue is related to the changes you made, but I might be wrong so please explain if that is the case.

PR #2386, opened 2026-03-20
[ADVAPP-2424]: Enhance the text editor for Engagements, Email, and SMS

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2424

### Technical Description

> Replace TipTap editor with Filament `RichEditor` for engagement, workflow, and campaign email flows. Integrate `SpatieMediaLibraryFileAttachmentProvider` so all uploaded images are stored in the media library instead of directly to S3. Rename `campaign_action_id` to a generic source morphTo relation on engagements, allowing both campaign actions and workflow steps to be linked as the engagement source. Create `EngagementFileAttachmentProvider` to resolve media from the source model when rendering engagement emails. Update AI draft actions to output `RichEditor` JSON schema via `RichContentRenderer` instead of raw HTML.

> No

> Yes. Temporary Data Migration: `app-modules/engagement/database/migrations/2026_03_24_192248_tmp_data_reset_oversized_image_dimensions_in_engagement_tables.php` AND
`app-modules/campaign/database/migrations/2026_04_09_105554_tmp_data_rename_media_collection_data_body_to_body_for_campaign_actions.php`
> In `tests/TenantMigrationTests.php`, we need to remove the test case related to the temporary migration (please do not remove the entire file).

---


danharrin (changes requested):
As discussed yesterday, there is some extra work required here to clean up image handling in workflows/campaigns

Orrison (changes requested):
@payal-canyon since @danharrin is OOO, please re-request my review when his feedback is addressed.

app-modules/engagement/src/Actions/CreateEngagementBatch.php
```diff
@@ -60,17 +60,12 @@ public function execute(EngagementCreationData $data): void
         $engagementBatch->successful_engagements = 0;
 
         DB::transaction(function () use ($engagementBatch, $data) {
+            $engagementBatch->body = $data->body;
```
Orrison: Just curious why here, and in app-modules/engagement/src/Actions/CreateEngagement.php, do we set `$engagementBatch->body = $data->body;` here instead of in the group about like it was done in this Olympus version of this PR?

app-modules/engagement/src/Filament/Pages/Inbox.php
```diff
@@ -126,8 +126,8 @@ public function table(Table $table): Table
                     ->sortable(),
                 TextColumn::make('subject')
                     ->description(
-                        fn (EngagementResponse $record): ?string => filled($body = $record->getBodyMarkdown())
-                            ? Str::limit(strip_tags($body), 50)
+                        fn (EngagementResponse $record): ?string => filled($body = $record->getBody())
+                            ? Str::limit(html_entity_decode(strip_tags($body), ENT_QUOTES | ENT_HTML5, 'UTF-8'), 50)
```
Orrison, line 134: Hmm, `EngagementResponse` does have a `subject`.

I think if we should change this to check the type, (like in the type column above). If it is SMS we can do what we are doing here.
If it is email then we should do what we are doing here, but get the content from the subject. (unless the subject is empty, then do this

app-modules/engagement/src/Models/EmailTemplate.php
```diff
@@ -71,4 +75,18 @@ public function user(): BelongsTo
...
+    {
+        $this->registerRichContent('content')
+            ->fileAttachmentsDisk('s3-public')
+            ->fileAttachmentProvider(SpatieMediaLibraryFileAttachmentProvider::make())
+            ->mergeTags([
+                'recipient first name' => '{{ recipient first name }}',
+                'recipient last name' => '{{ recipient last name }}',
+                'recipient full name' => '{{ recipient full name }}',
+                'recipient email' => '{{ recipient email }}',
+                'recipient preferred name' => '{{ recipient preferred name }}',
```
Orrison, line 90: I noticed the Olympus version here doesn't include the brackets. Is this correct? Or do we need to fix the Olympus one?

app-modules/engagement/src/Filament/Resources/EmailTemplates/Pages/CreateEmailTemplate.php
```diff
@@ -62,18 +66,24 @@ public function form(Schema $schema): Schema
...
+                    ->fileAttachmentsDisk('s3-public')
                     ->mergeTags($mergeTags = [
                         'recipient first name',
                         'recipient last name',
                         'recipient full name',
                         'recipient email',
                         'recipient preferred name',
                     ])
-                    ->tools(['bold', 'italic', 'small', 'link', 'color', '|', 'heading', 'bullet-list', 'ordered-list', 'hr', 'media', 'stock-image', '|', 'clear-formatting'])
+                    ->toolbarButtons([['bold', 'italic', 'small', 'link', 'textColor'], ['h1', 'h2', 'h3', 'bulletList', 'orderedList', 'horizontalRule', 'attachFiles', 'stockImage'], ['mergeTags']])
```
Orrison: I think you will need to filter `stockImage` out of this list if `getStockImagePlugin` returns `null`. This currently throws an error if you visit this page and it is not active.

You will need this change in the edit page as well, and anywhere else we are applying `stockImage` to the toolbar

Orrison (dismissed):

app-modules/engagement/src/Models/EmailTemplate.php
Orrison: After testing I will say I think this seems to be working. Do you think we need to change this to match in Olympus?

app-modules/engagement/src/Actions/CreateEngagementBatch.php
payal-canyon: Yes! good catch!

app-modules/engagement/src/Models/EmailTemplate.php
payal-canyon: @Orrison , Yeah, I was a bit confused as well when I first saw this. I checked with Dan, and he mentioned:

“Ideally, we should use the {{ }} format wherever we don’t have actual data to pass. For email templates, we don’t usually render merge tags, so it doesn’t make much difference there. However, in cases like engagement batches, where we preview the email, it’s helpful for users to see the {{ }} placeholders.”

So based on that, we should use {{ }} when actual data isn’t available.

app-modules/engagement/src/Filament/Resources/EmailTemplates/Pages/CreateEmailTemplate.php
payal-canyon: Good catch!

app-modules/engagement/src/Models/EmailTemplate.php
Orrison: Got it, I'll make a PR in Olympus to update that then. Thanks!

danharrin (changes requested):
As discussed, the relationship between engagement and workflow step needs to be established so we can fetch the media from the correct model when sending the engagement

app-modules/campaign/src/Models/CampaignAction.php
danharrin, line 1: I don't think I feel comfortable about overriding internal model methods like `getAttribute()`, but I understand why you did it. I think instead, we need to rethink how we handle the `$fieldPrefix` across blocks for campaigns. This has been a goal long term for the campaign module anyway, so it would be good to get it out of the way. I think that instead of adding the `$fieldPrefix` to the name of the fields, we should just wrap the fields in a `Group` or similar with a `->statePath($fieldPrefix)`. This way, the field names are correct and resolve to rich content attributes properly, but the state is still in the right place. Do you think this would work? It might also need applying to workflows, I am not sure.

danharrin: If we do this, we may need a data migration to fix the name of the media collection on these models, I just saw in other parts of the code you have the media collection with a `data.` prefix?

app-modules/workflow/src/Models/WorkflowEngagementEmailDetails.php
```diff
@@ -89,4 +94,33 @@ public function getActionExecutableJob(WorkflowRunStep $workflowRunStep): Execut
     {
         return new EngagementEmailWorkflowActionJob($workflowRunStep);
     }
+
+    public function getEditFields(): array
```
danharrin: I have seen this method to loop through fields and set the model in a few places, is there a particular reason why we can't hard-code the `->model()` call on that `RichEditor::make()` instead of doing this?

danharrin (changes requested):

danharrin (commented):
This PR is ready, but I am waiting on a new Filament release to fix some spacing issues with merge tags before updating this PR and merging.

danharrin (dismissed):

danharrin (dismissed):

danharrin (dismissed):

danharrin (approved):

PR #2427, opened 2026-04-03
[ADVAPP-2508]: Introduce the ability to opt out of email messages

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2508

### Technical Description

> Create EmailType Enums consisting value Marketing and Transactional.
> When any Engagement or campaign Emails are sent and unsubscribe link is created which will be sent only when Email Type is Marketing.
> When user clicks on unsubscribe link he/she will be opted out of the email.
> For now, we don't have toggles for email type so by default each email will be 'Transactional'.

> No

> EmailTypeFeature

---


app-modules/engagement/src/Actions/CreateEngagement.php
```diff
@@ -60,7 +62,7 @@ public function execute(EngagementCreationData $data, bool $notifyNow = false):
         $engagement->channel = $data->channel;
         $engagement->subject = $data->subject;
         $engagement->scheduled_at = $data->scheduledAt;
-
+        $engagement->email_type = EmailTypeFeature::active() ? EmailType::from($data->emailType) : EmailType::Transactional;
```
Orrison: If the feature is not active then I think the column would not exist, right?

I feel that may mean this would error? Let's make it so `email_type` is ONLY set if the feature is active

app-modules/engagement/src/Actions/CreateEngagementBatch.php
```diff
@@ -58,6 +60,7 @@ public function execute(EngagementCreationData $data): void
         $engagementBatch->total_engagements = $data->recipient->count();
         $engagementBatch->processed_engagements = 0;
         $engagementBatch->successful_engagements = 0;
+        $engagementBatch->email_type = EmailTypeFeature::active() ? EmailType::from($data->emailType) : EmailType::Transactional;
```
Orrison: Same as the other message

app-modules/engagement/src/Jobs/CreateBatchedEngagement.php
```diff
@@ -97,6 +99,7 @@ public function handle(): void
         $engagement->subject = $this->engagementBatch->subject;
         $engagement->body = $this->engagementBatch->body;
         $engagement->scheduled_at = $this->engagementBatch->scheduled_at;
+        $engagement->email_type = EmailTypeFeature::active() ? $this->engagementBatch->email_type : EmailType::Transactional;
```
Orrison: Same as above

app-modules/engagement/src/Notifications/EngagementNotification.php
```diff
@@ -59,14 +62,19 @@
...
     use Queueable;
 
     public function __construct(
         public Engagement $engagement
     ) {}
 
+    public function getEmailType(): string
+    {
+        return EmailTypeFeature::active() ? $this->engagement->email_type->value : EmailType::Transactional->value;
+    }
```
Orrison: This and the interface should return the actual `EmailType` enum instance, not the string value.

If the string value is needed by what calls this then it can do so

app-modules/engagement/tests/Tenant/Feature/Actions/CreateEngagementTest.php
```diff
@@ -39,10 +39,13 @@
 use AdvisingApp\Engagement\Models\Engagement;
 use AdvisingApp\Engagement\Notifications\EngagementNotification;
 use AdvisingApp\Engagement\Tests\Tenant\RequestFactories\CreateEngagementRequestFactory;
+use AdvisingApp\Notification\Enums\EmailType;
 use Illuminate\Support\Facades\Notification;
 
 use function Pest\Laravel\assertDatabaseCount;
 
+use Symfony\Component\Mime\Email;
```
Orrison: Idk if I am somehow missing something. But is this being used anywhere in this file? If not it should be removed.

app-modules/notification/src/Http/Controllers/UnsubscribeController.php
```diff
@@ -0,0 +1,79 @@
...
+        if ($request->isMethod('HEAD')) {
+            return response()->noContent();
+        }
+
+        $email = $request->query('email');
+
+        EmailAddressOptInOptOut::firstOrCreate(
+            ['address' => $email],
+            ['status' => EmailAddressOptInOptOutStatus::OptedOut],
+        )->update(['status' => EmailAddressOptInOptOutStatus::OptedOut]);
```
Orrison: I think you may want `updateOrCreate` here. It would make it so you don't need the chained `update` call

app-modules/notification/src/Notifications/Channels/MailChannel.php
```diff
@@ -100,6 +103,19 @@ public function send($notifiable, Notification $notification): ?SentMessage
             'recipient_address' => is_array($recipientAddress) ? null : $recipientAddress,
         ]);
 
+        $emailType = $notification instanceof HasEmailType
+        ? $notification->getEmailType()
+        : EmailType::Transactional->value;
+
+        $emailMessage->email_type = EmailType::from($emailType);
```
Orrison: Like I mentioned in another comment. This will be handled better if `HasEmailType` actually returns an `EmailType` instance and not the string version

app-modules/notification/src/Notifications/Contracts/HasEmailType.php
```diff
@@ -0,0 +1,42 @@
...
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Notification\Notifications\Contracts;
+
+interface HasEmailType
+{
+    public function getEmailType(): string;
```
Orrison: This return-type should be the actual enum

app-modules/notification/src/Support/UnsubscribeUrl.php
```diff
@@ -0,0 +1,62 @@
...
+     * @return string The signed unsubscribe URL
+     */
+    public static function generate(string $email, int $expirationMinutes = 43200): string
+    {
+        return URL::signedRoute(
+            'unsubscribe',
+            ['email' => $email],
+            now()->addMinutes($expirationMinutes)->toImmutable(),
+        );
+    }
```
Orrison: We should not have these expire. The the 30 day default here is within what is legally the minimum, most systems DO NOT expire them. Let's do so as well.

Orrison (changes requested):

app-modules/notification/src/Notifications/Channels/MailChannel.php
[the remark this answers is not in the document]
Orrison: I am pretty sure you can ignore this as I think we DO call `$emailMessage->save()` shortly after this. But could be good to confirm

app-modules/notification/src/Http/Controllers/UnsubscribeController.php
```diff
@@ -0,0 +1,79 @@
...
+use Illuminate\Contracts\View\View;
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Illuminate\Support\Facades\URL;
+
+class UnsubscribeController extends Controller
+{
+    public function show(Request $request): View
+    {
+        $email = $request->query('email');
```
Orrison, line 55: We should add validation to this. We can do the request validation in this controller instead of within a custom request class.

But we should validate the `email` is required and that it should be an `email`. We will want to do the same for the store below

Orrison (changes requested):
This is looking great @jayushi-canyon just one more change and I think we should be good to go.

Orrison (approved):

PR #2429, opened 2026-04-06
[ADVAPP-2547]: Introduce a filter in support programs by contact person

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2547

### Technical Description

> Introduce a filter in support programs by contact person

> No

> NO

---

### DIFF
--- a/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
+++ b/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
@@ -121,7 +121,17 @@ class ListBasicNeedsPrograms extends ListRecords
                         ->pluck('contact_person', 'contact_person')
                         ->all())
                     ->searchable()
-                    ->optionsLimit(40),
+                    ->getSearchResultsUsing(fn (string $search): array => BasicNeedsProgram::query()
+                        ->whereNotNull('contact_person')
+                        ->whereRaw('LOWER(contact_person) LIKE ?', ['%' . mb_strtolower($search) . '%'])
+                        ->distinct()
+                        ->orderBy('contact_person')
+                        ->limit(40)
+                        ->pluck('contact_person', 'contact_person')
+                        ->all())
+                    ->getOptionLabelUsing(fn (string | int | null $value): ?string => filled($value)
+                        ? BasicNeedsProgram::query()->where('contact_person', $value)->value('contact_person')
+                        : null),
             ], layout: FiltersLayout::BeforeContent)
             ->recordActions([
                 ViewAction::make(),
--- a/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
+++ b/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
@@ -129,9 +129,7 @@ class ListBasicNeedsPrograms extends ListRecords
                         ->limit(40)
                         ->pluck('contact_person', 'contact_person')
                         ->all())
-                    ->getOptionLabelUsing(fn (string | int | null $value): ?string => filled($value)
-                        ? BasicNeedsProgram::query()->where('contact_person', $value)->value('contact_person')
-                        : null),
+                    ->getOptionLabelUsing(fn (string | int | null $value): ?string => $value),
             ], layout: FiltersLayout::BeforeContent)
             ->recordActions([
                 ViewAction::make(),
--- a/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
+++ b/app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
@@ -37,6 +37,7 @@
 namespace AdvisingApp\BasicNeeds\Filament\Resources\BasicNeedsPrograms\Pages;
 
 use AdvisingApp\BasicNeeds\Filament\Resources\BasicNeedsPrograms\BasicNeedsProgramResource;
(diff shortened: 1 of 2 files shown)


app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
```diff
@@ -110,6 +111,17 @@ public function table(Table $table): Table
...
+                    ->label('Contact Person')
+                    ->options(fn (): array => BasicNeedsProgram::query()
+                        ->whereNotNull('contact_person')
+                        ->distinct()
+                        ->orderBy('contact_person')
+                        ->limit(40)
+                        ->pluck('contact_person', 'contact_person')
+                        ->all())
+                    ->searchable()
+                    ->optionsLimit(40),
```
Orrison: Having `options` set here can be good. But this configuration where you just have `searchable` will mean that ONLY these `40` options will be available and the search will only work across those `40` results on the frontend.

You will need to also include a `getSearchResultsUsing()` and `getOptionLabelUsing()`: https://filamentphp.com/docs/4.x/forms/select#returning-custom-search-results

And I think you can get rid of `optionsLimit`, I don't think it will be needed with this setup.

app-modules/basic-needs/tests/Tenant/BasicNeedsProgram/BasicNeedsProgramTest.php
Orrison, line 1: Pared with my other comment, I would add a test ensuring that searching is returning the results you would expect. I would even seed a bunch (more than 40) and ensure that the search returns results outside of your originally loaded 40

Orrison (dismissed):

app-modules/basic-needs/tests/Tenant/BasicNeedsProgram/BasicNeedsProgramTest.php
mittal-canyon: Added suggested testcase changes.

app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
mittal-canyon: Added suggested changes

app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
```diff
@@ -110,6 +111,27 @@ public function table(Table $table): Table
...
+                    ->getSearchResultsUsing(fn (string $search): array => BasicNeedsProgram::query()
+                        ->whereNotNull('contact_person')
+                        ->whereRaw('LOWER(contact_person) LIKE ?', ['%' . mb_strtolower($search) . '%'])
+                        ->distinct()
+                        ->orderBy('contact_person')
+                        ->limit(40)
+                        ->pluck('contact_person', 'contact_person')
+                        ->all())
+                    ->getOptionLabelUsing(fn (string | int | null $value): ?string => filled($value)
+                        ? BasicNeedsProgram::query()->where('contact_person', $value)->value('contact_person')
```
danharrin: This code is finding a program with the matching `contact_person` and then returning the `contact_person`, you could just return `$value` and not make a DB query at all

app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
```diff
@@ -110,6 +111,27 @@ public function table(Table $table): Table
...
+                        ->limit(40)
+                        ->pluck('contact_person', 'contact_person')
+                        ->all())
+                    ->searchable()
+                    ->getSearchResultsUsing(fn (string $search): array => BasicNeedsProgram::query()
+                        ->whereNotNull('contact_person')
+                        ->whereRaw('LOWER(contact_person) LIKE ?', ['%' . mb_strtolower($search) . '%'])
+                        ->distinct()
+                        ->orderBy('contact_person')
+                        ->limit(40)
```
danharrin, line 129: I think the limit we aim to use is 50 so please update both places that use that

danharrin (changes requested):

app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
mittal-canyon: In ticket description it's been told to go with 40 limit and in above comment also Kevin has mentioned 40 limit   so i have added like that, let me know if still need to do the limit changes.

app-modules/basic-needs/src/Filament/Resources/BasicNeedsPrograms/Pages/ListBasicNeedsPrograms.php
mittal-canyon: Resolved

app-modules/basic-needs/tests/Tenant/BasicNeedsProgram/BasicNeedsProgramTest.php
danharrin, line 1: The final two tests in this file don't test anything, you copied the queries into the test instead of testing the app code. I would suggest deleting them

danharrin (changes requested):

app-modules/basic-needs/tests/Tenant/BasicNeedsProgram/BasicNeedsProgramTest.php
mittal-canyon: I have updated test cases.

danharrin (approved):

PR #2438, opened 2026-04-08
[ADVAPP-2550]: Some users have reported that an error is thrown when you try and edit a campaign journey step of the type Task

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2550

### Technical Description

> Some users have reported that an error is thrown when you try and edit a campaign journey step of the type Task

> No

> No

---


app-modules/campaign/src/Filament/Blocks/TaskBlock.php
Orrison, line 1: For the things you added a custom `getSearchResultsUsing` to I think you need the `getOptionLabelUsing` as well: https://filamentphp.com/docs/4.x/forms/select#returning-custom-search-results

Orrison (changes requested):
Can you add some tests ensuring this is fixed in the form block and that the Job is doing what we expect now please?

Orrison (approved):

PR #2445, opened 2026-04-10
[ADVAPP-2541] : Improve campaigns by introducing the ability to archive campaigns

monali-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2541

### Technical Description

- Enabled archiving capability on campaign model via CanBeArchived
- Added campaign policy authorization for archive action (using existing delete permission pattern): CampaignPolicy
- Added archive button on edit page with dynamic label
- List page now excludes archived campaigns when feature is active

> No

- CampaignArchivingFeature
---


Orrison (changes requested):
We recently (today as of writing this) made an update to our `common` package adding support for the archive methods in PHPStan.

Could you please replicate using that and removing the phpstan-ignore's added here: https://github.com/canyongbs/aidingapp/pull/1000

Orrison:
@monali-canyon just a heads up that https://github.com/canyongbs/aidingapp/pull/1000 like I mentioned in my other comment has now merged in. So the most recent version of `common` upgraded to in that PR can be upgraded to here so that we can get rid of / not have to do all the PHPStan ignores for the archive methods / scopes

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -76,6 +79,17 @@ public function form(Schema $schema): Schema
     protected function getHeaderActions(): array
     {
         return [
+            ArchiveAction::make()
+                ->label(fn (Campaign $record): string => (bool) $record->getAttribute('enabled') ? 'Disable and Archive' : 'Archive')
```
Orrison: Why are we using `getAttribute` and casting to `(bool)` here (and everywhere else)?

Would `$record->enabled` not be sufficient? It is already cast to bool by the Models casts and should be available.

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -76,6 +79,17 @@ public function form(Schema $schema): Schema
...
+                ->modalHeading(fn (Campaign $record): string => (bool) $record->getAttribute('enabled') ? 'Disable and Archive Campaign' : 'Archive Campaign')
+                ->modalSubmitActionLabel(fn (Campaign $record): string => (bool) $record->getAttribute('enabled') ? 'Disable and Archive' : 'Archive')
+                ->hidden(fn (): bool => ! CampaignArchivingFeature::active())
+                ->action(function (Campaign $record): void {
+                    if ($record->enabled) {
+                        $record->forceFill(['enabled' => false])->save();
```
Orrison: Why call `forceFill` here? `enabled` is a fillable attribute

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -76,6 +79,17 @@ public function form(Schema $schema): Schema
...
+                    }
+                    $record->archive();
```
Orrison: Since you are making potentially multiple database changes here we would want to happen all at once or not at all, I would surround this all in a DB transaction / `try/catch` setup.

Also, just confirming whether this `action()` override will still result in the proper success / failure notifications? If not, we need to make sure that is added

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/EditCampaignTest.php
```diff
@@ -0,0 +1,124 @@
...
+use function Tests\asSuperAdmin;
+
+test('archive action is visible on edit page', function () {
+    asSuperAdmin();
+
+    $campaign = Campaign::factory()->enabled()->create();
+
+    livewire(EditCampaign::class, ['record' => $campaign->getRouteKey()])
+        ->assertActionVisible('archive');
+});
```
Orrison, line 70: This should assert it is visible once a user is built up to have the correct access needed. It is often best to do that 1-by-1.

What I mean is, create an initial User with no permissions or anything, assert they don't have access (or in this case, that they don't see the action) add one of the THINGS (permissions, settings, etc.) needed for access, assert they still don't see it, then eventually when they have all they should need assert access.

For all other tests not directly "access" related you can use a super admin, for brevity and ease. But access tests should ensure what is needed to HAVE access.

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/EditCampaignTest.php
```diff
@@ -0,0 +1,124 @@
...
+
+test('archive action exists on edit page', function () {
+    asSuperAdmin();
+
+    $campaign = Campaign::factory()->enabled()->create();
+
+    livewire(EditCampaign::class, ['record' => $campaign->getRouteKey()])
+        ->assertActionExists('archive');
+});
```
Orrison: I don't think this is really needed with the other access test

Orrison (changes requested):
Looking pretty good. I have some comments, there is also some merge conflicts to resolve

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
monali-canyon: $record->enabled is sufficient but it was showing linting error. I have fixed it now

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/EditCampaignTest.php
monali-canyon: Okay I have changed as your suggestions.

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
Orrison: That was most likely caused by your database and/or helper files being out of date. Make sure your primary Tenant has all migrations ran and then run `composer generate-helper-files` and it should clear up things like that by making sure the proper phpstan helper files are in place

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -76,6 +82,36 @@ public function form(Schema $schema): Schema
...
+                    } catch (Throwable $exception) {
+                        report($exception);
+
+                        Notification::make()
+                            ->danger()
+                            ->title('Failed to archive campaign')
+                            ->body($exception->getMessage())
+                            ->send();
+
+                        throw $exception;
```
Orrison: You should not `throw` the exception here. The point is to capture it and `report()` so that it goes through our error handling, and then show a handled error message to the user, instead of the exception bubbling up for the User to see.

Orrison (changes requested):

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
monali-canyon: Should I simply remove  `throw $exception;` ?

Orrison: Yes

Orrison (approved):

PR #2460, opened 2026-04-16
[ADVAPP-2534]: Some users are experiencing issues where email previews in the unified inbox and messages in the holistic profile show unexpected content

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2534

### Technical Description

> Some users are experiencing issues where email previews in the unified inbox and messages in the holistic profile show unexpected content

> No

> No

---

### DIFF
--- a/app-modules/engagement/src/Filament/Pages/Inbox.php
+++ b/app-modules/engagement/src/Filament/Pages/Inbox.php
@@ -130,15 +130,11 @@ class Inbox extends Page implements HasTable
                             return $record->subject;
                         }
 
-                        return filled($body = $record->getBody())
-                            ? Str::limit(html_entity_decode(strip_tags($body), ENT_QUOTES | ENT_HTML5, 'UTF-8'), 50)
-                            : null;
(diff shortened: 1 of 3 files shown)


danharrin (changes requested):
I don't think using regex to remove the markdown is the best approach here. Can you explain why the markdown syntax is present in the first place please?

amit-canyon:
> I don't think using regex to remove the markdown is the best approach here. Can you explain why the markdown syntax is present in the first place please?

@danharrin 

I need to remove Markdown-specific patterns such as `[link](https://example.com)` and `image` from the content.

While HTML tags are handled using strip_tags, there isn’t a built-in way to strip Markdown syntax. To handle this, I implemented a regex-based solution to clean these patterns from the description field.

If there’s a more appropriate way to handle removing this type of Markdown content from the table description, please let me know and I’ll proceed with that approach.

Thanks!

Orrison:
@danharrin has discovered this seems to no longer be an issue in Production, it seems it was fixed when we upgraded the editor / rendering code recently.

I am going to close this branch and the ticket out. If it is reported again we can look into it further.

PR #2462, opened 2026-04-16
[ADVAPP-2586] : Some users have reported that attempting to view a Task that is not related to a Student or Prospect throws errors

monali-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2586

### Technical Description

- When concern is null, the license check is skipped entirely (the user still needs the general ability permission).
- update — Same guard added on the license check, plus changed $task->concern->student_id to $task->concern?->student_id on the converted-Prospect check to prevent a secondary null error.

> No

> No
---


app-modules/task/src/Policies/TaskPolicy.php
```diff
@@ -67,7 +67,7 @@ public function viewAny(Authenticatable $authenticatable): Response
 
     public function view(Authenticatable $authenticatable, Task $task): Response
     {
-        if (! $authenticatable->hasLicense($task->concern->getLicenseType())) {
+        if ($task->concern && ! $authenticatable->hasLicense($task->concern->getLicenseType())) {
```
payal-canyon: `hasLicense(null)` returns `true` (early return on `blank($type)`). So the verbose `$task->concern && ! $authenticatable->hasLicense(...)` is unnecessary. Replace each policy check with:

```javascript
if (! $authenticatable->hasLicense($task->concern?->getLicenseType())) {
    return Response::deny(...);
}
```

Apply this to all 5 methods in this TaskPolicy (view, update, delete, restore, forceDelete).

For the Prospect-converted check on [TaskPolicy.php:94] the `?->` is correct - keep that line.

app-modules/project/src/Filament/Resources/Projects/Pages/ManageTasks.php
```diff
@@ -188,20 +188,17 @@ public function table(Table $table): Table
                                             ->default('Unassigned'),
                                         TextEntry::make('concern.full_name')
                                             ->label('Related To')
-                                            ->getStateUsing(
-                                                fn (Task $record): ?string => match ($record->concern->getMorphClass()) {
-                                                    Student::class => $record->concern->full_name,
-                                                    Prospect::class => $record->concern->full_name,
-                                                    default => null,
+                                            ->url(function (Task $record): ?string {
+                                                if ($record->concern instanceof Student) {
```
payal-canyon: let's replace the if/if/return null with `match (true) { instanceof … }` for parity with the rest of the file.

payal-canyon (changes requested):

payal-canyon (changes requested):
I noticed a few files that don’t seem related to this task. Could you please review and revert those changes, and then re-request my review?
Thanks!

payal-canyon (approved):

PR #2463, opened 2026-04-16
[ADVAPP-2542] & [ADVAPP-2543]: Introduce book with Round Robin & Availability option for group meetings

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2542
- https://canyongbs.atlassian.net/browse/ADVAPP-2543

### Technical Description

Introduces a round robin option for assigning group meetings, as per the existing round robin features present in Cases and Aiding App. Refactors the booking controller and tests to allow for these more complex options to be added without bloating existing files.

No

`app/Features/BookingGroupRoundRobinFeature.php`

---


danharrin:
I have decided to push Availability into this PR as well as Round Robin, since it should not hinder the review process too much (the logic and tests driving Availability mode are in their own files, not mixed with the rest of the files), and we need to test/deploy both together on staging so it should save time.

Orrison (dismissed):

Orrison (changes requested):
Two issues found during testing we will need to fix:

1. In the booking page UI when a month has no availability the calendar locks up and you can't move between months anymore
2. Round Robin should select the next person in the group that has availability, if the next person does NOT have availability then move on to the next, unless ALL members do not have avaiability

Orrison (approved):

PR #2470, opened 2026-04-20
[ADVAPP-2579]: Enhance enterprise AI features by including web search when helpful

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2579

### Technical Description

> Enable web_search tool on OpenAI Responses for enterprise AI. Adds the Azure OpenAI web_search tool alongside the existing file_search tool across all Responses API call sites in BaseOpenAiService (stream, streamRaw, sendNewMessage). Unlike file_search, web_search is unconditional and does not depend on a vector store being present.

> No

> No

---

### DIFF
--- a/app-modules/integration-open-ai/src/Services/BaseOpenAiService/Concerns/InteractsWithResearchRequests.php
+++ b/app-modules/integration-open-ai/src/Services/BaseOpenAiService/Concerns/InteractsWithResearchRequests.php
@@ -76,10 +76,13 @@ trait InteractsWithResearchRequests
                 ]),
             ]),
             providerOptions: [
-                'tools' => [[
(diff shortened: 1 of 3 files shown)


app-modules/integration-open-ai/src/Services/BaseOpenAiService/Concerns/InteractsWithResearchRequests.php
danharrin, line 1: I think we should avoid updating research here as the original ticket doesn't include it

danharrin (changes requested):

danharrin (approved):

PR #2498, opened 2026-04-30
[ADVAPP-2615]: Fix portal sidebar position

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2615

### Technical Description

> Fixed the resource hub portal sidebar so it spans the full viewport height and remains visible while scrolling the main content area.

> No

> No

---


danharrin (changes requested):
On both mobile and desktop, the logo should stick to the top of the sidebar and not scroll out of view

nency-canyon:
> On both mobile and desktop, the logo should stick to the top of the sidebar and not scroll out of view

Updated the logo to stick to the top of the sidebar in mobile and desktop. Please Verify, Thanks.

danharrin (changes requested):
Let's keep the topbar that contains the sidebar open button always sticky on mobile so that it cannot be scrolled off of the screen. Otherwise the user needs to scroll to the top of the page each time to open the sidebar on mobile.

nency-canyon:
> Let's keep the topbar that contains the sidebar open button always sticky on mobile so that it cannot be scrolled off of the screen. Otherwise the user needs to scroll to the top of the page each time to open the sidebar on mobile.

You're right, thanks for catching that. I initially applied sticky to the outer page wrapper, which made the sticky scope incorrect and could cause the behavior to break when scrolling into later sections. I corrected it by applying sticky only to the mobile topbar/button wrapper and keeping the page content in a normal scrolling container.
This is now fixed.

danharrin (approved):

PR #2505, opened 2026-05-04
[ADVAPP-2403]: Enhance admissions workflow by expanding triggers to have additional granularity related to entering and exiting an admissions stage

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2403

### Technical Description

> Enhance admissions workflow by expanding triggers to have additional granularity related to entering and exiting an admissions stage

> No

> Yes - it's documented in `.cleanup-tasks/2026_05_05_admissions_stage_workflow_triggers_feature.md`

---


database/migrations/2026_05_05_124659_data_activate_admissions_stage_workflow_triggers_feature.php
danharrin, line 1: I think this should be merged into the main migration & put in a transaction with it

app-modules/application/src/Filament/Resources/Applications/Pages/ManageApplicationWorkflows.php
```diff
@@ -62,12 +69,72 @@ public static function getNavigationLabel(): string
...
+    }
+
+    public function getTabs(): array
+    {
+        if (! AdmissionsStageWorkflowTriggersFeature::active()) {
+            return [];
+        }
+
+        $states = ApplicationSubmissionState::query()
+            ->withCount('workflowTriggers')
```
danharrin: I think this should be scoped to the current application, right? `->withCount(['workflowTriggers' => fn ($query) => $query->where('related_id', $this->getOwnerRecord()->getKey())])`

danharrin (changes requested):
Thanks for the thorough implementation here. The event-driven approach is solid. I do want to raise an architectural concern though, because I think it crosses a module boundary in a way that'll cause pain down the road.

The workflow module is designed to be a generic, reusable system across the whole app. This PR introduces application-domain knowledge directly into it in three places:

1. `workflow_triggers` migration `application_submission_state_id` FK
The workflow module's migration now creates a hard foreign key pointing at `application_submission_states`. This couples the workflow module's schema to the admissions module. If tomorrow case management or surveys wanted the same Enter/Exit trigger concept, we'd need another domain-specific FK column on the same table. The fix here is a polymorphic pair — `sub_related_type` / `sub_related_id` — so the `WorkflowTrigger` can point at any model as its "sub related" record without the workflow module needing to know what that model is.

2. `WorkflowTrigger` model `applicationSubmissionState()` relationship
Same problem at the Eloquent layer. `app-modules/workflow/src/Models/WorkflowTrigger.php` now imports `AdvisingApp\Application\Models\ApplicationSubmissionState` directly. The workflow module's core model should never reference a specific domain model. A generic `subRelated(): MorphTo` replaces this cleanly.

3. `WorkflowResource::form()` application-specific fields
This is the most visible violation. The generic workflow edit form now contains a Stage selector and Enter/Exit radio that are conditionally hidden behind `$record?->workflowTrigger?->related_type === 'application'`. The `WorkflowResource` should have no knowledge of admissions, states, etc. These fields should really live elsewhere in my opinion.

danharrin (approved):

PR #2509, opened 2026-05-04
[ADVAPP-2623]: Enhance the ability to add details when emails are actioned

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2623
 
### Technical Description

> Enhance the ability to add details when email/sms status changes to actioned.

> No

> Yes, `EngagementResponseMarkAsActionedFeature`

---

### DIFF
--- a/app-modules/engagement/src/Filament/Actions/EngagementResponseMarkAsActionedAction.php
+++ b/app-modules/engagement/src/Filament/Actions/EngagementResponseMarkAsActionedAction.php
@@ -48,7 +48,6 @@ use Filament\Forms\Components\Textarea;
 use Filament\Notifications\Notification;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Facades\DB;
-use Illuminate\Support\Str;
 
 class EngagementResponseMarkAsActionedAction extends Action
 {
@@ -58,16 +57,7 @@ class EngagementResponseMarkAsActionedAction extends Action
 
         $this
             ->label('Mark as Actioned')
-            ->modalHeading(function (?Model $record): string {
-                $engagementResponse = $this->getEngagementResponse($record);
-                $type = match ($engagementResponse?->type) {
-                    EngagementResponseType::Email => 'Email',
-                    EngagementResponseType::Sms => 'Text',
-                    default => null,
-                };
-
-                return 'Mark ' . $type . ' as Actioned';
-            })
+            ->modalHeading(fn (?Model $record): string => 'Mark ' . $this->resolveHeadingTypeLabel($record) . ' as Actioned')
             ->modalDescription(function (?Model $record): string {
                 return 'When you action ' . $this->resolveTypeLabel($record) . ', you are indicating that you have taken all necessary steps to respond to this ' . $this->resolveSenderLabel($record) . '. Please describe below what steps you have taken.';
             })
@@ -101,7 +91,7 @@ class EngagementResponseMarkAsActionedAction extends Action
                 });
 
                 Notification::make()
-                    ->title(Str::ucfirst($this->resolveTypeLabel($record)) . ' marked as actioned')
+                    ->title($this->resolveHeadingTypeLabel($record) . ' marked as actioned')
                     ->success()
                     ->send();
             });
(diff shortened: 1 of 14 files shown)


app-modules/engagement/src/Filament/Actions/EngagementResponseMarkAsActionedAction.php
```diff
@@ -0,0 +1,163 @@
...
+            $record instanceof HolisticEngagement && $record->record instanceof EngagementResponse => $record->record,
+            $record instanceof Timeline && $record->timelineable instanceof EngagementResponse => $record->timelineable,
+            default => null,
+        };
+    }
+
+    public function resolveEngagementResponse(?Model $record): ?EngagementResponse
+    {
+        return $this->getEngagementResponse($record);
+    }
```
payal-canyon, line 150: `resolveTypeLabel` and `resolveSenderLabel` currently duplicate logic that can be simplified.
Both methods should delegate to `getEngagementResponse()` and then call their respective helpers, instead of maintaining separate internal match logic.
Also, `resolveEngagementResponse()` is just a one-line alias of `getEngagementResponse()` and doesn’t add any value-so it can be removed.
Remove `resolveEngagementResponse()`

```javascript
public function getTypeLabel(?EngagementResponse $engagementResponse): ?string
{
    return match ($engagementResponse?->type) {
        EngagementResponseType::Email => 'an email',
        EngagementResponseType::Sms => 'a text',
        default => null,
    };
}

public function resolveTypeLabel(?Model $record): ?string
{
    return $this->getTypeLabel($this->getEngagementResponse($record));
}

public function getSenderLabel(?EngagementResponse $engagementResponse): ?string
{
    return match (true) {
        $engagementResponse?->sender instanceof Student => 'student',
        $engagementResponse?->sender instanceof Prospect => 'prospect',
        default => null,
    };
}

public function resolveSenderLabel(?Model $record): ?string
{
    return $this->getSenderLabel($this->getEngagementResponse($record));
}

public function getEngagementResponse(?Model $record): ?EngagementResponse
{
    return match (true) {
        $record instanceof EngagementResponse => $record,
        $record instanceof HolisticEngagement && $record->record instanceof EngagementResponse => $record->record,
        $record instanceof Timeline && $record->timelineable instanceof EngagementResponse => $record->timelineable,
        default => null,
    };
}
```

payal-canyon (changes requested):

app-modules/engagement/src/Filament/Actions/EngagementResponseMarkAsActionedAction.php
```diff
@@ -0,0 +1,163 @@
...
+            });
+    }
+
+    public function getTypeLabel(?EngagementResponse $engagementResponse): ?string
+    {
+        return match ($engagementResponse?->type) {
+            EngagementResponseType::Email => 'an email',
+            EngagementResponseType::Sms => 'a text',
+            default => null,
+        };
```
payal-canyon, line 106: Let’s remove the articles `a` and `an`. so, this matches the ticket description exactly.

payal-canyon (changes requested):

app-modules/engagement/src/Filament/Actions/EngagementResponseMarkAsActionedAction.php
amit-canyon: @payal-canyon  Articles already added in description that's why I added here, Please see: 
`When you action an {email/text},`

payal-canyon: Yeah, I see what you’re saying. The sample title in the description is `Mark {Email/Text} as Actioned`, but the actual description uses “a”/“an”. That’s why I suggested it earlier.
But it’s fine-let’s leave it as is.

amit-canyon: Ahh, got your point. Thanks!

payal-canyon (approved):

PR #2516, opened 2026-05-07
[ADVAPP-2625]: Enhance the ability to use institutional data by allowing more granular settings for Employee Advisor

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2625

### Technical Description

> Enhance the ability to use institutional data by allowing more granular settings for Employee Advisor

> No

> Yes,  `AiAssistantResourceHubCategoryFeature`

---


Orrison:
Changed review to @danharrin as he is most familiar with this here and he wrote the tech plan for it

danharrin (dismissed):
Can you please name all tables / new classes / foreign keys with the "employee advisor" terminology instead of "ai assistant"? This will make the future ticket to rename this feature easier.

danharrin (changes requested):
If you enable `has_resource_hub_knowledge`, then save the assistant, wait for its vector store to be created, then change one of the settings like resource hub article access level or the categories, it doesn't seem to update the knowledge. I think the `AiAssistantObserver` needs to be updated to handle changes to the `access` attribute, and maybe a new observer needs to be created for the new model as well so that the parent advisor's vector store is synced again.

danharrin (approved):

PR #2518, opened 2026-05-08
[ADVAPP-2631]: Some users have reported that labels for mapped blocks do not appear when creating forms

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2631

### Technical Description

Adds an additional label element to this field in the FormKit schema.

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/form/resources/views/blocks/previews/educatable-address.blade.php
+++ b/app-modules/form/resources/views/blocks/previews/educatable-address.blade.php
@@ -33,16 +33,32 @@
 --}}
 <x-form::blocks.field-wrapper :$label :$isRequired :description="$description ?? null">
     <div class="space-y-2">
-        <x-form::blocks.previews.field />
-        <x-form::blocks.previews.field />
-        <x-form::blocks.previews.field />
-        <div class="grid grid-cols-2 gap-2">
+        <x-form::blocks.field-wrapper label="Address Line 1">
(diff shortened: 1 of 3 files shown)


payal-canyon (changes requested):
Looks good overall - just one small addition. Similar to the fix we applied for the `Address`, we also need to apply the same fix for the `Name` field.

While testing locally, I noticed that `EducatableNameFormFieldBlock` has the exact same issue. We should apply the same fix there as well.

danharrin:
I did think about that but I'm not sure that is the desired experience @payal-canyon, I think the existing name fields are quite self-explanatory and given that was not reported by product I think we should leave it.

payal-canyon:
> I did think about that but I'm not sure that is the desired experience @payal-canyon, I think the existing name fields are quite self-explanatory and given that was not reported by product I think we should leave it.

I agree with what you’re trying to say. However, the product team already mentioned this in the ticket description: `Make sure all mapped blocks output their label in the form rendering.`

Also, in the attached screenshot they included `YOUR NAME` similar to `YOUR ADDRESS`. Based on that, I assumed this change was expected as part of the ticket. If you still feel that we do not need to add this, then let me know and I will approve the PR and merge it into main.

Thank You!

payal-canyon (approved):

PR #2519, opened 2026-05-08
[Chore]: Remove Octane

Orrison:
### Ticket(s) or GitHub Issue

- N/A

### Technical Description

This PR makes a few changes to our underlying deployment strategy to primarily remove Laravel Octane make improvements related to that, and change a few other configurations for improvement of the architecture.

- Removes Laravel Octane and replaces it with PHP-FPM
- Tunes php.ini, nginx, and PHP-FPM setting for better performance. Primarily increasing cache sizes, enabling JIT, etc.
- Removes the old `webuser` user configuration and properly configures `www-data` default nginx user. Also makes it so all Docker containers run as that use instead of `root` for security. Because of this all ports now run `8080` and `8443` instead of `80` and `443`. No changes needed locally for the ports, but there will be a small time period before I update the other apps to do this as well where advising app commands will need to call `./pls` since it's `pls` file is slightly different to account for the user change.
- Updates Modular to v3 to enable proper caching of the modular code
- Fixes some tests

NOTE: This should NOT be merged in till https://github.com/canyongbs/devops/pull/90 is merged in and this PR is updated to switch back to devops main. Additionally, this branch will need to be rebased onto `main` again since it currently extends the last real release tag so that it could be deployed to staging for testing without releasing new changes to staging yet.

Discussed offline

### Cleanup Tasks

N/A

---


danharrin (changes requested):
The code changes look sane and work locally, awaiting the devops changes if we are going ahead with removing Octane.

danharrin (approved):

PR #2530, opened 2026-05-13
[ADVAPP-2635]: Rename QnA Advisors to Customer Advisors in code and database tables / columns

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2635

### Technical Description

Create migration to rename tables, foreign columns, and permissions from QnA Advisors to Customer Advisors; update tests and policies in respect; update names of models, filament pages, etc

No

### Cleanup Tasks

`cleanup-tasks/2026_05_12_rename_qna_advisors_feature.md`

---


app-modules/ai/src/Providers/AiServiceProvider.php
```diff
@@ -98,11 +98,11 @@ public function boot(): void
...
-            'qna_advisor' => QnaAdvisor::class,
-            'qna_advisor_category' => QnaAdvisorCategory::class,
-            'qna_advisor_file' => QnaAdvisorFile::class,
-            'qna_advisor_link' => QnaAdvisorLink::class,
-            'qna_advisor_question' => QnaAdvisorQuestion::class,
+            'customer_advisor' => CustomerAdvisor::class,
+            'customer_advisor_category' => CustomerAdvisorCategory::class,
+            'customer_advisor_file' => CustomerAdvisorFile::class,
+            'customer_advisor_link' => CustomerAdvisorLink::class,
+            'customer_advisor_question' => CustomerAdvisorQuestion::class,
```
Orrison, line 105: One thing that can be a little tricky about renaming these is that we need to be sure that we don't have any polymorphic relationships that COULD have referenced these old names previously. Or else it would now fail to resolve those records.

If we want to be careful, you can keep the old key names, referencing the new class names, AFTER the definitions of the new string / class names. Like this:

```php
'customer_advisor' => CustomerAdvisor::class,
'customer_advisor_category' => CustomerAdvisorCategory::class,
'customer_advisor_file' => CustomerAdvisorFile::class,
'customer_advisor_link' => CustomerAdvisorLink::class,
'customer_advisor_question' => CustomerAdvisorQuestion::class,
'qna_advisor' => CustomerAdvisor::class, // LEGACY, kept for backward compatibility
'qna_advisor_category' => CustomerAdvisorCategory::class, // LEGACY, kept for backward compatibility
'qna_advisor_file' => CustomerAdvisorFile::class, // LEGACY, kept for backward compatibility
'qna_advisor_link' => CustomerAdvisorLink::class, // LEGACY, kept for backward compatibility
'qna_advisor_question' => CustomerAdvisorQuestion::class, // LEGACY, kept for backward compatibility
```

This way, when NEW records are written, it will find the first morphmap record referencing `CustomerAdvisor` and save with `customer_advisor`, but then on READ out of the DB it will be able to still map `qna_advisor` to the new class FQCN

You also can get rid of / NOT do the legacy backwards compatibility items if you feel strongly that there are none, or you have a migration that renames all references to it in an pertinent DB column.

app-modules/ai/routes/widgets.php
```diff
@@ -34,42 +34,42 @@
...
 Route::middleware([
     'api',
     EncryptCookies::class,
-    QnaAdvisorWidgetCors::class,
+    CustomerAdvisorWidgetCors::class,
 ])
-    ->name('widgets.ai.qna-advisors.')
-    ->prefix('widgets/ai/qna-advisors')
+    ->name('widgets.ai.customer-advisors.')
+    ->prefix('widgets/ai/customer-advisors')
```
Orrison, line 139: There are existing widget embeds in the wild now that would be broken by renaming this with no redirect / fallbacks for the old names.

If you can please add some strategy for this that is simple and works with what we have here already.

I am not sure the best way to handle it off the top of my head. But let me know if you want to pair on it.

Orrison (changes requested):

app-modules/ai/routes/widgets.php
Orrison, line 1: Rather than actually writing out the same routes, you could define all the grouped routes in like a variable stored function and call it for both. That way this is smaller and easier to maintain. This is very rough, but something like:

```php
$registerAdvisorRoutes = function () {
    Route::prefix('api/{advisor}')
        ->name('api.')
        ->middleware([
            EnsureCustomerAdvisorEmbedIsEnabled::class,
            EnsureCustomerAdvisorRequestComingFromAuthorizedDomain::class,
        ])
        ->group(function () {
            Route::get('/', CustomerAdvisorResourcesController::class)->name('assets');
            Route::post('/entry', ShowAdvisorController::class)->name('entry');
            // ... all the other routes ...
        });

    Route::get('{file?}', CustomerAdvisorResourceController::class)
        ->where('file', '(.*)')
        ->name('asset');
};

// Primary routes
Route::middleware(['api', EncryptCookies::class, CustomerAdvisorWidgetCors::class])
    ->name('widgets.ai.customer-advisors.')
    ->prefix('widgets/ai/customer-advisors')
    ->group($registerAdvisorRoutes);

// Legacy support
Route::middleware(['api', EncryptCookies::class, CustomerAdvisorWidgetCors::class])
    ->name('widgets.ai.qna-advisors.')
    ->prefix('widgets/ai/qna-advisors')
    ->group($registerAdvisorRoutes);
```

Orrison (dismissed):

payal-canyon (changes requested):
Hey @rebecca-canyon !
It looks like this PR currently has merge conflicts. Could you please resolve them and re-request my review once that's done? Thanks!

app-modules/ai/src/Settings/AiCustomerAdvisorSettings.php
```diff
@@ -53,6 +54,6 @@ class AiQnaAdvisorSettings extends Settings
 
     public static function group(): string
     {
-        return 'ai-qna-advisor';
+        return RenameQnaAdvisorsFeature::active() ? 'ai-custom-advisor' : 'ai-qna-advisor';
     }
```
payal-canyon: I think we missed the `er` here. This should be `customer`, not `custom`, so let's add the missing `er`.

app-modules/ai/database/migrations/2026_05_12_222040_rename_qna_advisors_table_and_columns_to_customer_advisors.php
```diff
@@ -0,0 +1,133 @@
...
+            Schema::rename('qna_advisor_threads', 'customer_advisor_threads');
+            Schema::rename('qna_advisors', 'customer_advisors');
+
+            Schema::table('customer_advisor_categories', function (Blueprint $table) {
+                $table->renameColumn('qna_advisor_id', 'customer_advisor_id');
+            });
+
+            DB::table('settings')
+                ->where('group', 'ai-qna-advisor')
+                ->update(['group' => 'ai-customer-advisor']);
```
payal-canyon, line 87: After these setting renames, we may need to update the `portal_authentications` table as well.
Could you confirm whether this data exists there? If it does, let's handle the rename in the migration and revert it in the `down()` method too.
```javascript
DB::table('portal_authentications')
    ->where('portal_type', 'qna_advisor_widget')
    ->update(['portal_type' => 'customer_advisor_widget']);
```

payal-canyon (changes requested):

payal-canyon (dismissed):

payal-canyon (approved):

PR #2533, opened 2026-05-14
[ADVAPP-2652]: Ensure that resource hub vector stores contain the most up to date version of the resource hub articles at all times

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2652

### Technical Description

Ensure that the resource hub vector stores are updated when the article category is updated, and when internal articles are created immediately upload them to vector stores.

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/resource-hub/src/Observers/ResourceHubArticleObserver.php
+++ b/app-modules/resource-hub/src/Observers/ResourceHubArticleObserver.php
@@ -44,20 +44,12 @@ class ResourceHubArticleObserver
 {
     public function created(ResourceHubArticle $article): void
     {
-        if (! $article->public) {
-            return;
-        }
-
         SyncResourceHubArticlesToAssistantVectorStores::dispatch();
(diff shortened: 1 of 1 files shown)


app-modules/resource-hub/src/Observers/ResourceHubArticleObserver.php
```diff
@@ -44,20 +44,12 @@ class ResourceHubArticleObserver
 {
     public function created(ResourceHubArticle $article): void
     {
-        if (! $article->public) {
-            return;
-        }
```
Orrison, line 49: I don't think we want this removed here. And if we are removing it because when edited articles that are not public are synced I think this reveals a deeper issue.

I don't think we want non-public articles synced to the AI vector store.

Non-public articles don't show in the resources hub, so doing this would mean an individual could use the AI to get details on articles they are not supposed to see?

Orrison (changes requested):

Orrison (approved):

PR #2534, opened 2026-05-14
[ADVAPP-2405]: Improve custom advisors by adding a question-and-answer pair functionality

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2405

### Technical Description

Copies the Q&A feature from customer advisors to employee advisors using the same implementation code.

No

### Cleanup Tasks

Created

---

### DIFF
--- a/app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
+++ b/app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
@@ -17,7 +17,7 @@
       in the software, and you may not remove or obscure any functionality in the
       software that is protected by the license key.
     - You may not alter, remove, or obscure any licensing, copyright, or other notices
-      of the licensor in the software. Any use of the licensor's trademarks is subject
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
       to applicable law.
     - Canyon GBS Inc. respects the intellectual property rights of others and expects the
       same in return. Canyon GBS® and Advising App® are registered trademarks of
--- a/app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
+++ b/app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Advising App® is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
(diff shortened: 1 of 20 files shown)


app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $qnaSection = $this->generateQnaSection($assistant);
+
+        return <<<END
+
+        # Instructions
+        {$instructions}
+
+        {$qnaSection}
+        END;
```
payal-canyon, line 63: In `GetEmployeeAdvisorQnaInstructions::execute()`, we wrap the instructions in the # Instructions heredoc even when there are no Q&A pairs to add. Since the feature flag check in `BaseOpenAiService` isn't scoped to anything, every assistant with no Q&A configured still gets its prompt re-shaped the moment the migration activates the flag - even ones that don't use Q&A. Can we just return `$assistant->instructions` as-is when `generateQnaSection()` is empty? That way only assistants with actual Q&A content get the new format.
WDYT?

app-modules/ai/src/Models/EmployeeAdvisorCategory.php
```diff
@@ -0,0 +1,75 @@
...
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use Illuminate\Database\Eloquent\SoftDeletes;
+use OwenIt\Auditing\Auditable as AuditableTrait;
+use OwenIt\Auditing\Contracts\Auditable;
+
+/**
+ * @mixin IdeHelperEmployeeAdvisorCategory
+ */
+class EmployeeAdvisorCategory extends BaseModel implements Auditable
+{
```
payal-canyon, line 50: I noticed that the QnA version caches the generated instructions for an hour and uses observers on the category/question models to clear the cache on save/delete.
In the new implementation, caching doesn’t seem to be added yet, and the new models also don’t have observers. Because of that, we may end up hitting the DB and rebuilding the prompt on every chat message.
Was this intentionally skipped for now, or should we consider porting that behavior as well to keep it aligned with the QnA implementation?

payal-canyon (changes requested):

app-modules/ai/src/Models/EmployeeAdvisorCategory.php
danharrin: I intentionally skipped its to be honest, hitting the DB each time is not heavy work and means we don't have to handle cache invalidation for all tables that hold data used in the instructions.

app-modules/ai/src/Actions/GetEmployeeAdvisorQnaInstructions.php
danharrin: Done

payal-canyon (approved):

PR #2538, opened 2026-05-15
[ADVAPP-2650]: Set and enforce a default datetime format across the project and in Filament components

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2650

### Technical Description

> Set and enforce a default datetime format across the project and in Filament components

> No

### Cleanup Tasks

> No

---


resources/views/filament/widgets/welcome-widget.blade.php
danharrin, line 1: I think this is a place we should keep the old format as it is the user dashboard where a longer format date would be appropriate.

danharrin (changes requested):

danharrin (approved):

PR #2542, opened 2026-05-18
[ADVAPP-2670]: Improve performance and response time for embedded customer advisors (chatbots)

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2670

### Technical Description

Fixes streaming in responses for customer chatbots. Adds response time logging. Decreases chunk size. Always disables reasoning for customer advisors and integrated AI like Draft With AI.

Uses the background queue for sending AI messages after the response is sent, in a new PHP process. The background queue did not originally work due to missing middleware in the Kernel. Then I discovered that it also wasn't tenant-aware, so I implemented a custom connection which makes it tenant-aware.

Also fixes customer advisor previews which were broken by the Octane removal.

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/ai/resources/js/chat.js
+++ b/app-modules/ai/resources/js/chat.js
@@ -261,7 +261,11 @@ document.addEventListener('alpine:init', () => {
 
                         if (!this.hasSetUpNewMessageForResponse) {
                             console.log('First chunk of AI response received at:', new Date().toISOString());
-                            console.log('That response time was: ', (new Date() - this.messageSentAt) / 1000, 'seconds');
+                            console.log(
+                                'That response time was: ',
+                                (new Date() - this.messageSentAt) / 1000,
+                                'seconds',
+                            );
 
                             if (!this.isCompletingPreviousResponse) {
                                 this.messages.push({
--- a/app-modules/ai/resources/js/chat.js
+++ b/app-modules/ai/resources/js/chat.js
@@ -112,6 +112,7 @@ document.addEventListener('alpine:init', () => {
             csrfToken,
             lockedMessage: null,
             clickHandler: null,
+            pendingFinishedEvent: null,
 
             init: async function () {
                 this.clickHandler = async (event) => {
@@ -170,6 +171,10 @@ document.addEventListener('alpine:init', () => {
 
                 setInterval(() => {
                     if (/^\s*$/.test(this.pendingResponse)) {
+                        if (this.pendingFinishedEvent) {
+                            this.finalizeResponse(this.pendingFinishedEvent);
+                            this.pendingFinishedEvent = null;
+                        }
                         return;
                     }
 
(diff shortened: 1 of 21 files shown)


app-modules/ai/resources/js/chat.js
```diff
@@ -260,6 +265,13 @@ document.addEventListener('alpine:init', () => {
                         this.startResponseTimeout();
 
                         if (!this.hasSetUpNewMessageForResponse) {
+                            console.log('First chunk of AI response received at:', new Date().toISOString());
+                            console.log(
+                                'That response time was: ',
+                                (new Date() - this.messageSentAt) / 1000,
+                                'seconds',
+                            );
```
Orrison: Here, and anywhere else we are doing this, let's hide this a little bit by ONLY outputting the `(new Date() - this.messageSentAt) / 1000` call.

That way it won't be clear to someone just looking, what this is for.

app-modules/ai/resources/js/chat.js
```diff
@@ -446,6 +440,9 @@ document.addEventListener('alpine:init', () => {
             },
 
             sendMessage: async function (prompt = null) {
+                console.log('Message sent at:', new Date().toISOString());
```
Orrison: We can just remove this here, per my other comment

widgets/qna-advisor/src/App.ce.vue
```diff
@@ -193,6 +262,10 @@
     async function sendMessage() {
         if (!sendMessageUrl.value || !message.value.trim() || !threadId.value) return;
 
+        console.log('Message sent at:', new Date().toISOString());
```
Orrison: Let's remove this log

widgets/qna-advisor/src/App.ce.vue
```diff
@@ -368,26 +442,31 @@
...
+                    resetWordQueue();
                     isLoading.value = false;
                     return;
                 }
 
                 if (data.content) {
-                    currentResponse.value += data.content;
+                    if (!hasLoggedFirstChunk) {
+                        console.log('First chunk of AI response received at:', new Date().toISOString());
+                        console.log('That response time was: ', (new Date() - messageSentAt) / 1000, 'seconds');
```
Orrison: Like my other comments, let's just output `new Date() - messageSentAt) / 1000` to the log`

Orrison (changes requested):
I tested this locally and it seems to work well and is pretty fast.

I did want to ask if there are going to be any issues with automatic re-tries with switching to background?

danharrin:
For employee advisors, there are retries controlled by the frontend instead of the queue, and the customer advisors did not have retries as a feature before this point.

Orrison (approved):

PR #2552, opened 2026-05-19
[ADVAPP-2666]: Enhance forms and admissions workflows by introducing support for tags

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2666

### Technical Description

Implement workflow tags block. Allow users to select tags for both students and prospects, to be applied as appropriate

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
+++ b/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
@@ -42,6 +42,7 @@ use AdvisingApp\Workflow\Models\WorkflowEngagementSmsDetails;
 use AdvisingApp\Workflow\Models\WorkflowProactiveConcernDetails;
 use AdvisingApp\Workflow\Models\WorkflowRun;
 use AdvisingApp\Workflow\Models\WorkflowRunStep;
+use AdvisingApp\Workflow\Models\WorkflowTagsDetails;
 use Exception;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
@@ -58,6 +59,7 @@ class WorkflowRunStepFactory extends Factory
             WorkflowProactiveConcernDetails::factory(),
             // WorkflowCareTeamDetails::factory(),
             WorkflowCaseDetails::factory(),
+            WorkflowTagsDetails::factory(),
             // WorkflowTaskDetails::factory(),
             // Add more as factories are created
         ];
--- a/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
+++ b/app-modules/workflow/database/factories/WorkflowRunStepFactory.php
@@ -74,6 +74,7 @@ class WorkflowRunStepFactory extends Factory
                     WorkflowProactiveConcernDetails::class,
                     // WorkflowCareTeamDetails::class,
                     WorkflowCaseDetails::class,
+                    WorkflowTagsDetails::class,
                     // WorkflowTaskDetails::class,
                 ];
 
--- a/app-modules/workflow/src/Filament/Blocks/TagsBlock.php
+++ b/app-modules/workflow/src/Filament/Blocks/TagsBlock.php
@@ -36,6 +36,8 @@
 
 namespace AdvisingApp\Workflow\Filament\Blocks;
 
+use AdvisingApp\Workflow\Models\WorkflowDetails;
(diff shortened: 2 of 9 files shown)


app-modules/workflow/database/factories/WorkflowRunStepFactory.php
```diff
@@ -58,6 +59,7 @@ public function definition(): array
             WorkflowProactiveConcernDetails::factory(),
             // WorkflowCareTeamDetails::factory(),
             WorkflowCaseDetails::factory(),
+            WorkflowTagsDetails::factory(),
```
payal-canyon, line 62: We added `WorkflowTagsDetails::factory()` to `$detailFactories` but never added `WorkflowTagsDetails::class` to `$possibleModels`. So if `WorkflowRunStep::factory()` is called without `withDetails()` and the random pick lands on tags, the resolver loops through `$possibleModels`, finds nothing, and throws `Could not determine model type`. That's a 1-in-5 random failure waiting to happen. Add the class to `$possibleModels` so the two lists stay in sync.

app-modules/workflow/src/Filament/Resources/Workflows/RelationManagers/WorkflowStepsRelationManager.php
```diff
@@ -253,6 +254,10 @@ private function createWorkflowDetails(WorkflowActionBlock $block, array $data):
                 'suggested_intervention' => $transformedData['suggested_intervention'],
                 'status_id' => $transformedData['status_id'],
             ]),
+            'workflow_tags_block' => WorkflowTagsDetails::create([
+                'tag_ids' => array_merge($transformedData['student_tag_ids'], $transformedData['prospect_tag_ids']),
```
payal-canyon: I think it'd be better if we store `Student` and `Prospect` tags in two separate columns instead of merging both into a single `tag_ids` column.

This is likely also why, while editing an existing Tags step, the tag fields currently appear empty - the form fields don't fully align with the data structure being stored.

So, we could either:

- store `Student` and `Prospect` tags in their own separate columns so the form fields map cleanly to the saved data, or
- keep the single `tag_ids` column and add logic to split the values back into the two fields when editing a step.

What do you think, @Orrison ? If you have a cleaner approach in mind, we can go with that as well.

payal-canyon (changes requested):

app-modules/workflow/src/Filament/Blocks/TagsBlock.php
```diff
@@ -36,18 +36,87 @@
...
     }
 
+    public function prepareForEdit(WorkflowDetails $details): void
+    {
+        assert($details instanceof WorkflowTagsDetails);
+        $tags = Tag::find($details['tag_ids']);
+        $details['student_tag_ids'] = $tags->where('type', TagType::Student)->pluck('id');
+        $details['prospect_tag_ids'] = $tags->where('type', TagType::Prospect)->pluck('id');
+    }
+
```
payal-canyon, line 126: This currently only handles loading, not saving. When editing an existing Tags step - say, adding more Student or Prospect tags - the existing data loads correctly on opening the Edit form, but the updated selections aren't being saved. So we'll need to handle the save side as well, by merging the two fields back into `tag_ids` before the update.

payal-canyon (dismissed):
We'll also need to update the `TagsWorkflowActionJob` - at the moment it syncs the full merged `tag_ids` onto the educatable without checking the `type`, so a `Prospect` can end up with `Student` tags (and vice versa) saved in the `taggables` table as invisible orphan rows. It would be good to filter the tags down to only the ones matching the educatable's type - `Student` or `Prospect` - before syncing, so each record only ever receives its own kind of tags. It would also be worth adding a test case to cover this: ideally one where a step carries both Student and Prospect tags, asserting that a Prospect submission receives only the Prospect tags while a Student submission receives only the Student ones, so this cross-type scenario stays protected going forward.

payal-canyon:
As I’m OOD now (my office hours are over), @rebecca-canyon , once the suggestions are resolved, could you please request a review from @Orrison ? Thank you!

payal-canyon (approved):

Orrison (approved):
Looks great!

PR #2554, opened 2026-05-20
[ADVAPP-2659]: Enhance verification of phone numbers for text messaging

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2659

### Technical Description

> Validate new phone numbers with a one-time Telnyx lookup. Adds a one-time Telnyx Number Lookup for every new or changed Student and Prospect phone number; the result (line type + carrier) is stored and reused, so each number is only ever looked up once. Lookups are triggered by the phone-number observers, the bulk import, and an Olympus-key-protected endpoint called after a SIS sync. 

> Also fixes two pre-existing bugs surfaced while testing the bulk import: the `Sync Records` column-mapper fieldsets were permanently invisible after the Filament upgrade (so `columnMap` was never submitted and the import errored), and the phone-number importer rejected blank extensions because its ext column lacked a `nullable` rule.

> No

### Cleanup Tasks

> Yes. `.cleanup-tasks/2026_05_22_phone_number_lookup_feature.md`

---


app-modules/prospect/src/Observers/ProspectPhoneNumberObserver.php
```diff
@@ -47,4 +50,27 @@ public function creating(ProspectPhoneNumber $prospectPhoneNumber): void
...
+        if (! app(PhoneNumberLookupService::class)->isConfigured()) {
+            return;
+        }
+
+        // Reuse an existing lookup result rather than paying for another.
+        if (PhoneNumberLookup::query()->where('number', $prospectPhoneNumber->number)->exists()) {
+            return;
+        }
+
+        LookupPhoneNumber::dispatch($prospectPhoneNumber->number)->afterCommit();
```
Orrison: I would prefer we use the `dispatch()` helper instead of the `dispatch` static method on the job class, since that way you can new up the class and it has better completion and static analysis than the static dispatch method, which does NOT know the arguments needed.

app-modules/student-data-model/database/migrations/2026_05_20_170052_create_phone_number_lookups_table.php
```diff
@@ -0,0 +1,71 @@
...
+            Schema::create('phone_number_lookups', function (Blueprint $table) {
+                $table->uuid('id')->primary();
+                $table->string('number')->unique();
+                $table->string('status');
+                $table->string('carrier_name')->nullable();
+                $table->string('carrier_type')->nullable();
+                $table->jsonb('raw_response')->nullable();
+                $table->timestamps();
+
+                $table->index('number');
```
Orrison: Doesn't the `unique()` create a unique index?

app-modules/student-data-model/routes/api.php
```diff
@@ -61,6 +62,9 @@
     ->group(function () {
         Route::post('/update-sis-settings', UpdateStudentInformationSystemSettingsController::class)
             ->name('update-sis-settings');
+
+        Route::post('/trigger-phone-number-lookups', TriggerPhoneNumberLookupsController::class)
+            ->name('trigger-phone-number-lookups');
```
Orrison: I think that this route/name, the controller/name, comments / return of the controller, etc. should just be normalized around "SIS sync completed" and dispatching the event.

That fact that THEN the system has a listener for that event to do a phone number lookup scan is irrelevant to this general process of just getting notified of the completion.

app-modules/student-data-model/src/Http/Controllers/SisSyncCompletedController.php
Orrison, line 1: As I said in my other comment, I think this should all be normalized around "SisSyncCompleted" in naming, comments, etc.

How the system reacts to the event is not related to the event being triggered.

app-modules/student-data-model/src/Models/PhoneNumberLookup.php
```diff
@@ -0,0 +1,88 @@
...
+        return $this->belongsTo(StudentPhoneNumber::class, 'number', 'number');
+    }
+
+    /**
+     * @return BelongsTo<ProspectPhoneNumber, $this>
+     */
+    public function prospectPhoneNumber(): BelongsTo
+    {
+        return $this->belongsTo(ProspectPhoneNumber::class, 'number', 'number');
+    }
```
Orrison, line 84: Technically these actually should be `hasMany`, right?

I don't think `number` is unique for both `StudentPhoneNumber` and `ProspectPhoneNumber` so technically multiple Student/Prospect's / Phone numbers can have the same number.

app-modules/student-data-model/src/Observers/StudentPhoneNumberObserver.php
```diff
@@ -47,4 +50,26 @@ public function creating(StudentPhoneNumber $studentPhoneNumber): void
...
+            return;
+        }
+
+        // Reuse an existing lookup result rather than paying for another.
+        if (PhoneNumberLookup::query()->where('number', $studentPhoneNumber->number)->exists()) {
+            return;
+        }
+
+        LookupPhoneNumber::dispatch($studentPhoneNumber->number)->afterCommit();
+    }
```
Orrison, line 57: Because this and the prospect version are so similar I wonder if we should consider making something shared here, like a shared trait (like we do with user track saving) that applies this. So that way the logic remains centralized and shared between student / prospect phone numbers?

app-modules/integration-twilio/src/Services/TelnyxPhoneNumberLookupService.php
```diff
@@ -0,0 +1,212 @@
...
+
+    protected function performLookup(string $normalizedNumber): PhoneNumberLookup
+    {
+        $apiKey = app(TwilioSettings::class)->telnyx_api_key;
+
+        if (blank($apiKey)) {
+            throw new RuntimeException('Cannot perform a Telnyx number lookup: the Telnyx API key is not configured.');
+        }
+
+        Telnyx::setApiKey($apiKey);
```
Orrison, line 114: Because this is global, it could be good to call `Telnyx::setApiKey(null)` just before any return here to make sure it doesn't accidentally carry over to other Tenants.

app-modules/student-data-model/src/Providers/StudentDataModelServiceProvider.php
```diff
@@ -76,11 +82,28 @@ public function boot(): void
...
+
+        // Telnyx applies rate limits per account, and each tenant uses its own
+        // Telnyx account/API key, so the limit must be keyed per tenant. The
+        // conservative 2 requests/second default can be raised once real
+        // throughput is observed. Consumed by the LookupPhoneNumber job.
+        RateLimiter::for('telnyx-number-lookup', function (): Limit {
+            $tenantKey = Tenant::current()?->getKey() ?? 'landlord';
+
+            return Limit::perSecond(2)->by("telnyx-number-lookup:{$tenantKey}");
+        });
```
Orrison, line 108: I would change this to `10` per second.

Then in the job that reaches out to Telnyx, I would react to the rate limit headers: https://developers.telnyx.com/development/api-fundamentals/reliability/rate-limiting#rate-limit-headers

And if you get a 429 then release the job based on `x-ratelimit-reset`

This will mean that you will need to make sure the job that is rate limited in that way does `retryUntil` instead of being based on attempts.
And the retryUntil can be something large at first, like 6 hours.

app-modules/student-data-model/src/Providers/StudentDataModelServiceProvider.php
```diff
@@ -76,11 +82,28 @@ public function boot(): void
...
         Event::listen(
             SesBounceEvent::class,
             SaveBouncedEmailAddress::class
         );
+
+        // After a SIS sync completes, scan for any phone numbers that have not
+        // yet been looked up and queue the lookups.
+        Event::listen(SisSyncCompleted::class, function (): void {
+            QueuePhoneNumberLookups::dispatch();
+        });
```
Orrison: I would make this an actual listener class. But DON'T make it a queued listener, just dispatch the job from it.

Queued listeners are best avoided as they are limited in what they can do in customizing the underlying job. So it's always best imo to just have a listener that dispatches a job, much like what you are doing here.

app-modules/student-data-model/src/Jobs/LookupPhoneNumber.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        // Skip silently when no lookup provider is configured for this tenant;
+        // the number simply stays unchecked and can be picked up later.
+        if (! $phoneNumberLookupService->isConfigured()) {
+            return;
+        }
+
+        try {
+            $phoneNumberLookupService->lookup($this->phoneNumber);
+        } catch (InvalidArgumentException $exception) {
```
Orrison: I would make this a custom named `InvalidArgumentException` that could wrap the thrown `InvalidArgumentException` as the `$previous` that way we know for certain it was the phone number validation that through this and not something else

app-modules/student-data-model/src/Jobs/LookupPhoneNumber.php
```diff
@@ -0,0 +1,125 @@
...
+            );
+        }
+    }
+
+    /**
+     * @return array<object>
+     */
+    public function middleware(): array
+    {
+        return [new RateLimitedWithRedis('telnyx-number-lookup')];
```
Orrison: Based on my other comment in the service provider about this rate limiting, when you get a rate limit back from the API, it could be a good idea to store the TTL on when it is clear in the cache, then have middleware that checks if that is currently valid and then release that job based on the TTL. That way we don't even attempt to hit the API if we know we are just going to be getting a rate limited response back.

Orrison (changes requested):

app-modules/student-data-model/database/factories/PhoneNumberLookupFactory.php
```diff
@@ -0,0 +1,182 @@
...
+ */
+class PhoneNumberLookupFactory extends Factory
+{
+    /**
+     * @return array<string, mixed>
+     */
+    public function definition(): array
+    {
+        $carrierName = $this->faker->company();
+        $carrierType = $this->faker->randomElement(['mobile', 'fixed line', 'voip', 'toll free']);
```
Orrison: These should be inline with the arrary/return and if another field needs their value then they need to be defined before that field and that field can retrieve their value from `$attributes` within a closure.

You never want to have things like this outside of the array in a factory as that messes with people in the future that want to override some fields.

app-modules/student-data-model/database/factories/PhoneNumberLookupFactory.php
```diff
@@ -0,0 +1,182 @@
...
+            ],
+        ]);
+    }
+
+    /**
+     * @return Factory<PhoneNumberLookup>
+     */
+    protected function withCarrierType(string $carrierType): Factory
+    {
+        $carrierName = $this->faker->company();
```
Orrison: Same thing as the above in the definition here, If we can make sure we don't do this in any factory definition or state

app-modules/student-data-model/src/Observers/Concerns/QueuesPhoneNumberLookupOnSave.php
```diff
@@ -0,0 +1,77 @@
...
+ * Observer concern that queues a Telnyx lookup whenever a phone-number record
+ * is first introduced or its `number` changes — provided no lookup result
+ * exists for that number yet and the lookup service is configured.
+ *
+ * Used by both Student and Prospect phone-number observers; the model is
+ * expected to expose a `number` attribute.
+ */
+trait QueuesPhoneNumberLookupOnSave
+{
+    public function saved(Model $phoneNumber): void
```
Orrison: I think it may be best to type hint this as a union of the two models used here. That should prevent you from having to do the `getAttribute` call, type what it is, etc.

app-modules/student-data-model/src/Jobs/LookupPhoneNumber.php
Orrison, line 1: So, per what I was talking about with Joe during Sprint Planning.

Product had an expectation that this ticket would also result in making sure that invalid / non-textable numbers DON'T allow users to select them for sending, nor would the SMSChannel, etc. attempt to send a text to them.

So we need to include that in the changes here.

I believe, though please confirm, that the system is setup right now to react in this way to both phone number models `can_receive_sms` column.
So should we update that after we get the lookup? Or something else?

Additionally both models have a `type` column that I think users can set if they have the ability to do so, and that we also get synced in from the SIS sync.
How should the lookups play with that now since we actually get the "real" type back from Telynyx / the carrier.

Curious your thoughts on how we should handle all this, also we will need to make sure we handle making numbers non-textable in this increment.

Orrison (changes requested):

app-modules/student-data-model/src/Jobs/LookupPhoneNumber.php
payal-canyon: @Orrison , Regarding `can_receive_sms`: I agree that it currently gates sending, but today it's part of a three-way `AND` condition:
- `can_receive_sms`
- No `SmsOptOutPhoneNumber` record
- No `BouncedPhoneNumber` record

The bounce and opt-out checks do not modify `can_receive_sms`; they exist as separate records associated with the phone number. Following that same pattern, it seems cleaner to introduce a fourth condition based on the lookup status rather than mutating `can_receive_sms` itself.

My understanding is that `can_receive_sms` represents user consent/intent, while the lookup result represents carrier capability. Updating the column based on lookup results would overwrite the user's setting and would require additional handling if Telnyx later corrected the classification. Keeping these concerns separate appears more consistent with the existing bounce/opt-out implementation.

From what I can see, this gate would naturally flow through the existing enforcement points:
`canReceiveSms()` on `Student`/`Prospect`
`SmsChannel::send()`
The engagement recipient picker
The Healthy/Unhealthy primary phone scopes

For textability, I understand that only `Invalid` and `ValidLandline` should be considered non-textable. `ValidMobile`, `ValidVoip`, and `ValidTollFree` would remain textable, while `Unknown`, `LookupFailed`, and cases where no lookup record exists yet would not block SMS sending.

Regarding the `type` column, it also seems preferable not to overwrite the saved value, especially since a future SIS sync could simply replace it again. Instead, displaying the lookup's `carrier_type` and textability verdict as a badge and tooltip alongside the existing type and `can_receive_sms` fields would allow the lookup result to serve as the display-layer source of truth while preserving the saved values as user-editable context.

For manually created phone numbers, my understanding is that the existing flow already handles this. The observer dispatches a `LookupPhoneNumber` job whenever a phone number is saved, regardless of whether it originates from SIS sync, CSV import, the Filament create form, or the API. Since the lookup runs asynchronously, there will be a brief period where the number has not yet been scanned. During that time, it would follow the same behavior as `Unknown` or "no lookup record yet" and remain textable by default. No additional create-time handling should be necessary.

One limitation worth noting is that there is currently no manual override. If Telnyx incorrectly classifies a textable number as Invalid, the only recovery path would be deleting the `phone_number_lookups` record and allowing the number to be scanned again. A manual override or "force textable" option could be considered in a future enhancement if needed.

Please let me know if this approach looks good, or if you would recommend any changes. I'm happy to adjust it if you have something else in mind.

Orrison: @payal-canyon yes this is all making a lot of sense.

I think I agree with keeping both the `can_receive_sms` and `type` that is in the existing data model separate from the Telnyx lookup. Especially because a nightly SIS sync would just overwrite any changes anyways.

For `type` I think I agree that the best possible way to handle this is visually display like a badge or something of the Telnyx scan type. This keeps them separate and also has an added benefit of being able to show the institution that they have some data in their system that may not be accurate per carrier records.

I also agree with keeping text message -ability handling within the `canReceiveSms` check. BUT I am not totally sure on Products intention here and I think it's important enough to ask for clarity.
Please construct a message, Product focused, that we can add as a comment to the ticket for this blocking question.
You can DM it to me first for approval then post a comment in the ticket tagging Product and adding a flag. (Then once you have the comment DM it to me as well so I can ping folks to get an answer quicker.)

What we need answered is whether or not that want us to take into account the data models `can_receive_sms` field, which comes in from manual edits if enabled or via the SIS sync, AND the Telnyx check.
OR
If they want us to consider the Telnyx check overriding of the `can_receive_sms` information input into our data model?
So an example scenario would be, a number is only textable if BOTH are true, OR if a valid textable number scan from Telnyx technically overrides a data model `can_receive_sms` being true/false

app-modules/student-data-model/src/Models/Concerns/IsTextable.php
```diff
@@ -0,0 +1,98 @@
...
+                ->exists();
+        }
+
+        return (bool) $this->getAttribute('can_receive_sms');
+    }
+
+    /**
+     * @param Builder<static> $query
+     */
+    public function scopeTextable(Builder $query): void
```
Orrison: Though I think it is nice that this is wrapped together with the per model bool method above. I would like us to not use local magic method scopes. I need to add a PHPStan rule to ban them and clear out the ones that exist.

Can we please convert this to a tappable scope?

app-modules/student-data-model/src/Models/Scopes/HealthyEducatablePrimaryPhoneNumber.php
```diff
@@ -50,11 +50,10 @@ class HealthyEducatablePrimaryPhoneNumber
...
     {
-        return $query->where(function (Builder $query) {
-            $query->whereHas('primaryPhoneNumber', function (Builder $query) {
-                $query->where('can_receive_sms', true)
-                    ->whereDoesntHave('smsOptOut');
-            });
-        });
+        return $query->whereHas(
+            'primaryPhoneNumber',
+            // @phpstan-ignore method.notFound (scope is on StudentPhoneNumber|ProspectPhoneNumber, not the generic Builder)
```
Orrison: Having to do this ignore is why I prefer tappable scopes.

app-modules/student-data-model/src/Models/Scopes/UnhealthyEducatablePrimaryPhoneNumber.php
```diff
@@ -50,12 +50,12 @@ class UnhealthyEducatablePrimaryPhoneNumber
...
-                ->orWhereHas('primaryPhoneNumber', function (Builder $query) {
-                    $query->where('can_receive_sms', false)
-                        ->orWhereHas('smsOptOut');
-                });
-        });
+        // Either no primary phone OR a primary phone that is not textable
+        // (opted out, bounced, or lacking a textable Telnyx lookup).
+        return $query->whereDoesntHave(
+            'primaryPhoneNumber',
+            // @phpstan-ignore method.notFound (scope is on StudentPhoneNumber|ProspectPhoneNumber, not the generic Builder)
```
Orrison: Same as the above.

app-modules/student-data-model/src/Models/PhoneNumberLookup.php
```diff
@@ -0,0 +1,88 @@
...
+use Illuminate\Database\Eloquent\Relations\HasMany;
+use OwenIt\Auditing\Contracts\Auditable;
+
+/**
+ * @mixin IdeHelperPhoneNumberLookup
+ */
+class PhoneNumberLookup extends BaseModel implements Auditable
+{
+    use HasUuids;
+    use AuditableTrait;
```
Orrison: This doesn't need to be auditable since this is not a record that user creates

app-modules/student-data-model/src/Observers/Concerns/QueuesPhoneNumberLookupOnSave.php
```diff
@@ -0,0 +1,73 @@
...
+use AdvisingApp\StudentDataModel\Models\StudentPhoneNumber;
+
+/**
+ * Observer concern that queues a Telnyx lookup whenever a phone-number record
+ * is first introduced or its `number` changes — provided no lookup result
+ * exists for that number yet and the lookup service is configured.
+ */
+trait QueuesPhoneNumberLookupOnSave
+{
+    public function saved(StudentPhoneNumber|ProspectPhoneNumber $phoneNumber): void
```
Orrison: Defining this trait method as the same name as the observer life-cycle event could be problematic in the future.

If we needed to do something different in one observer over the over, we now would overwrite this by defining a `saved` method on either observer.

Perhaps we should name this method something different and simply call it from each observer saved method?

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPhoneNumbers/CreateStudentPhoneNumberControllerTest.php
```diff
@@ -158,5 +154,51 @@
     '`type` is max 255 characters' => [['type' => str_repeat('a', 256)], 'type', 'The type may not be greater than 255 characters.'],
     '`order` is integer' => [['order' => 'not-an-integer'], 'order', 'The order must be an integer.'],
     '`ext` is integer' => [['ext' => 'not-an-integer'], 'ext', 'The ext must be an integer.'],
-    '`can_receive_sms` is boolean' => [['can_receive_sms' => 'not-boolean'], 'can_receive_sms', 'The can receive sms field must be true or false.'],
 ]);
+
+// TODO: When the PhoneNumberLookupFeature is removed, delete this test
+// case and add a `can_receive_sms is boolean` dataset entry back to the
+// `validates` test above.
```
Orrison: "and add a `can_receive_sms is boolean` dataset entry back to the `validates` test above."

I don't think that's correct, right? `can_receive_sms` will be gone so there is no need for a test for it?

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/StudentPhoneNumbers/UpdateStudentPhoneNumberControllerTest.php
```diff
@@ -149,5 +150,57 @@
     '`type` is max 255 characters' => [['type' => str_repeat('a', 256)], 'type', 'The type may not be greater than 255 characters.'],
     '`order` is integer' => [['order' => 'not-an-integer'], 'order', 'The order must be an integer.'],
     '`ext` is integer' => [['ext' => 'not-an-integer'], 'ext', 'The ext must be an integer.'],
-    '`can_receive_sms` is boolean' => [['can_receive_sms' => 'not-boolean'], 'can_receive_sms', 'The can receive sms field must be true or false.'],
 ]);
+
+// TODO: When the PhoneNumberLookupFeature is removed, delete this test
+// case and add a `can_receive_sms is boolean` dataset entry back to the
+// `validates` test above.
```
Orrison: Same as above

app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/CreateStudentControllerTest.php
```diff
@@ -365,9 +366,60 @@
     '`phone_numbers.*.type` is max 255 characters' => [['phone_numbers' => [['number' => '123456789', 'type' => str_repeat('a', 256)]]], 'phone_numbers.0.type', 'The phone_numbers.0.type may not be greater than 255 characters.'],
     '`phone_numbers.*.order` is a valid integer' => [['phone_numbers' => [['number' => '123456789', 'order' => 'not-integer']]], 'phone_numbers.0.order', 'The phone_numbers.0.order must be an integer.'],
     '`phone_numbers.*.ext` is a valid integer' => [['phone_numbers' => [['number' => '123456789', 'ext' => 'not-integer']]], 'phone_numbers.0.ext', 'The phone_numbers.0.ext must be an integer.'],
-    '`phone_numbers.*.can_receive_sms` is boolean' => [['phone_numbers' => [['number' => '123456789', 'can_receive_sms' => 'not-boolean']]], 'phone_numbers.0.can_receive_sms', 'The phone_numbers.0.can_receive_sms field must be true or false.'],
 ]);
 
+// TODO: When the PhoneNumberLookupFeature is removed, delete this test
+// case and add a `phone_numbers.*.can_receive_sms is boolean` dataset
+// entry back to the `validates` test above.
```
Orrison: Same as above

app-modules/student-data-model/src/Jobs/LookupPhoneNumber.php
```diff
@@ -0,0 +1,132 @@
...
+use AdvisingApp\StudentDataModel\Jobs\Middleware\SkipWhilePhoneNumberLookupIsRateLimited;
+use AdvisingApp\StudentDataModel\Models\PhoneNumberLookup;
+use DateTimeInterface;
+use Illuminate\Contracts\Queue\ShouldBeUnique;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Foundation\Queue\Queueable;
+use Illuminate\Queue\Middleware\RateLimitedWithRedis;
+use Throwable;
+
+class LookupPhoneNumber implements ShouldBeUnique, ShouldQueue
```
Orrison, line 52: It could be a good idea to also add a max exceptions of `3` to this job as well.

That way it fails if we get 3 uncaught exceptions without trying to retry for 6 hours.

app-modules/student-data-model/src/Models/Scopes/Textable.php
Orrison, line 1: This migration is dated AFTER the other two.

Meaning the tmp migration will dispatch with the feature flag being turned off and very likely not run because of that, right?

Perhaps we should just consolidate this one and the tmp dispatch all into `app-modules/student-data-model/database/migrations/2026_05_20_170052_create_phone_number_lookups_table.php` to be safe?

Orrison (changes requested):
Looks great!

Just a few minor issues and this should be good to go.

Orrison (approved):

PR #2555, opened 2026-05-20
[ADVAPP-2662]: Enhance the semester order table with new columns and optimized actions

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2662

### Technical Description

> Enhance the semester order table with new columns and optimized actions

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/student-data-model/src/Filament/Resources/EnrollmentSemesters/EnrollmentSemesterResource.php
+++ b/app-modules/student-data-model/src/Filament/Resources/EnrollmentSemesters/EnrollmentSemesterResource.php
@@ -98,7 +98,7 @@ class EnrollmentSemesterResource extends Resource
                 $query
                     ->addSelect('enrollment_semesters.*')
                     ->addSelect(DB::raw(
-                        "EXISTS(SELECT 1 FROM enrollments WHERE LOWER(enrollments.semester_name) = LOWER(enrollment_semesters.name) AND enrollments.deleted_at IS NULL) AS is_mapped"
+                        'EXISTS(SELECT 1 FROM enrollments WHERE LOWER(enrollments.semester_name) = LOWER(enrollment_semesters.name) AND enrollments.deleted_at IS NULL) AS is_mapped'
                     ));
             })
             ->columns([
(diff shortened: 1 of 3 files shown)


app-modules/student-data-model/src/Filament/Resources/EnrollmentSemesters/EnrollmentSemesterResource.php
```diff
@@ -93,13 +94,24 @@ public static function infolist(Schema $schema): Schema
     public static function table(Table $table): Table
     {
         return $table
+            ->modifyQueryUsing(function (Builder $query) {
+                $query
+                    ->addSelect('enrollment_semesters.*')
+                    ->addSelect(DB::raw(
+                        'EXISTS(SELECT 1 FROM enrollments WHERE LOWER(enrollments.semester_name) = LOWER(enrollment_semesters.name) AND enrollments.deleted_at IS NULL) AS is_mapped'
+                    ));
+            })
```
Orrison: Rather than using addSelects here, I think it may be better to use laravels `withExists`, this is a little rough so you will need to test if this works, but something like this:

```php
->modifyQueryUsing(fn (Builder $query) => $query->withExists([
    'enrollments as is_mapped' => fn ($query) => $query->whereRaw(
        'LOWER(enrollments.semester_name) = LOWER(enrollment_semesters.name)'
    )
]))
```

I think that should also account for the deleted_at but you will need to test.

Let's also add tests confirming the `is_mapped` column of this table returns correctly in various scenarios, which will be useful in confirming the `deleted_at` part above.

Orrison (changes requested):

app-modules/student-data-model/tests/Tenant/Filament/Resources/EnrollmentSemesters/Pages/ManageEnrollmentSemestersTest.php
```diff
@@ -144,3 +145,43 @@
...
 });
+
+test('the view action is not present on the table', function () {
+    asSuperAdmin();
+
+    $semester = EnrollmentSemester::factory()->create();
+
+    livewire(ManageEnrollmentSemesters::class)
+        ->assertActionDoesNotExist(TestAction::make('view')->table($semester));
+});
```
Orrison: I don't think we need to test this, this can be removed

Orrison (changes requested):

Orrison (approved):

PR #2557, opened 2026-05-20
[ADVAPP-2658]: Add channel details in the unified inbox sent page

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2658

### Technical Description

- Added `getRecipientRoute()` and `getRecipientRouteHealthStatus() `methods that resolve the recipient address from the latest message and check bounce/opt-out status using exists() queries
-  Switched to `ViewColumn` for the channel column and eager-loaded
- Created `channel-detail.blade.php` rendering the channel icon, recipient route text, and a color-coded health status icon with tooltip (healthy/bounced/opted-out)
- Test cases added

No 

### Cleanup Tasks

No

---


Orrison (dismissed):
Can we change the styling / colors of the status icons to match those that are currently set in the holistic profiles please?

Once you have done so make sure to update your uploaded screenshot in Jira

Orrison:
@payal-canyon unfortunately some things came up. Can you please take over final review of this item for me?

app-modules/engagement/src/Models/Engagement.php
```diff
@@ -331,6 +362,32 @@ public function setUpRichContent(): void
             ->mergeTags($this->getMergeData());
     }
 
+    protected function resolveEmailHealthStatus(string $address): EmailHealthStatus
```
payal-canyon: Reuse the existing `getHealthStatus()` methods instead of duplicating that logic.
`StudentEmailAddress`, `ProspectEmailAddress, StudentPhoneNumber`, and `ProspectPhoneNumber` already have `getHealthStatus()` - and Sent Items shows both students and prospects, so all four apply. Look up the matching email/phone record on the recipient by address/number and call its existing method. That kills the duplication, picks up `PhoneHealthStatus::NoSmsCapability` for free, and keeps a single source of truth if bounce/opt-out logic ever changes.

Also, eager-load `emailAddresses` / `phoneNumbers` on the Sent Items query alongside the existing message relations to avoid an N+1 when resolving the recipient's matching record per row.

Once that's done, the tests will need to attach a `StudentEmailAddress` / `StudentPhoneNumber` to the recipient with the matching address/number - currently they only seed the BouncedEmailAddress / opt-out rows, so the lookup would return null and the health icon wouldn't render.

app-modules/engagement/resources/views/filament/columns/channel-detail.blade.php
```diff
@@ -0,0 +1,62 @@
...
+
+    $channel = $record->channel;
+    $recipientRoute = $record->getRecipientRoute();
+    $healthStatus = $record->getRecipientRouteHealthStatus();
+@endphp
+
+<div class="flex items-center gap-2">
+    <x-filament::icon
+        class="h-5 w-5 shrink-0 text-gray-500 dark:text-gray-400"
+        icon="{{ $channel->getIcon() }}"
```
payal-canyon, line 43: Match the holistic profile styling - drop the hardcoded text/icon colors.
The channel icon hardcodes `text-gray-500 dark:text-gray-400` and the recipient route span hardcodes `text-sm text-gray-950 dark:text-white`. The holistic profile's `email-address-detail` / `phone-number-detail` components don't set any of these - they just use `h-5 w-5 shrink-0` on the icon and no text classes on the address, so everything inherits from the surrounding cell. Remove those classes here so the column matches the rest of the app visually.

app-modules/engagement/tests/Tenant/Filament/Pages/SentItemsTest.php
```diff
@@ -0,0 +1,360 @@
...
+        ->deliverNow()
+        ->forStudent()
+        ->create(['dispatched_at' => now()]);
+
+    livewire(SentItems::class)
+        ->assertSuccessful()
+        ->assertTableColumnExists('channel')
+        ->assertCanSeeTableRecords([$engagement])
+        ->assertSeeHtml('student@university.edu')
+        ->assertSeeHtml(EmailHealthStatus::Healthy->getColorClasses())
```
payal-canyon: This value is just `text-success-600 dark:text-success-400` - a generic class string that can appear on any success-colored element anywhere on the page (status badges, other icons, etc.). The test can pass even if the channel column renders wrong.

Add `->state(fn (Engagement $r) => $r->getRecipientRoute())` to the `ViewColumn` in `SentItems.php`, then assert with `assertTableColumnStateSet('channel', 'student@university.edu', record: $engagement)` - that's tied to the specific record's column, not "anywhere on the page".

payal-canyon (changes requested):

app-modules/engagement/src/Models/Engagement.php
```diff
@@ -337,4 +373,22 @@ protected static function booted(): void
             $builder->tap(new LicensedToEducatable('recipient'));
         });
     }
+
+    private function resolveEmailHealthStatus(Educatable $recipient, string $route): EmailHealthStatus
+    {
+        /** @var StudentEmailAddress|ProspectEmailAddress $emailAddress */
+        $emailAddress = $recipient->emailAddresses()->where('address', $route)->first()
```
payal-canyon: Can we use `->emailAddresses->firstWhere('address', $route)` (and `->phoneNumbers->firstWhere('number', $route))` instead? The current `->emailAddresses()->where(...)->first()` re-queries per row. Combined with eager-loading `emailAddresses` / `phoneNumbers` via morphWith on the Sent Items query, it skips the N+1 entirely.

app-modules/engagement/tests/Tenant/Filament/Pages/SentItemsTest.php
```diff
@@ -0,0 +1,469 @@
...
+        ->deliverNow()
+        ->create([
+            'dispatched_at' => now(),
+            'recipient_id' => $student->sisid,
+            'recipient_type' => (new Student())->getMorphClass(),
+        ]);
+
+    livewire(SentItems::class)
+        ->assertSuccessful()
+        ->loadTable()
```
payal-canyon: I noticed that `loadTable()` is being used in several places throughout this test case file. Is there a specific reason for that? It doesn't seem to be necessary, so I think we can remove those calls.

app-modules/engagement/resources/views/filament/columns/channel-detail.blade.php
```diff
@@ -0,0 +1,62 @@
...
+
+<div class="flex items-center gap-2">
+    <x-filament::icon
+        class="h-5 w-5 shrink-0"
+        icon="{{ $channel->getIcon() }}"
+        x-tooltip.raw="{{ $channel->getLabel() }}"
+    />
+
+    @if (filled($recipientRoute))
+        <span class="text-sm">
```
payal-canyon, line 48: Let's remove only `class="text-sm"` and leave the rest as is.

app-modules/engagement/tests/Tenant/Filament/Pages/SentItemsTest.php
```diff
@@ -0,0 +1,469 @@
...
+        ->deliverNow()
+        ->create(['recipient_route' => null]);
+
+    expect($engagement->getRecipientRoute())->toBeNull();
+});
+
+it('returns healthy email health status when address is not bounced or opted out', function () {
+    asSuperAdmin();
+
+    $student = Student::factory()->create();
```
payal-canyon, line 278: All of the new tests currently use Student, StudentEmailAddress, and StudentPhoneNumber. Since the functionality supports both Students and Prospects, and Sent Items renders both, it would be good to add a test case covering the Prospect side as well to provide more complete coverage.

payal-canyon (changes requested):

app-modules/engagement/src/Filament/Pages/SentItems.php
```diff
@@ -99,7 +100,9 @@ public function table(Table $table): Table
     {
         return $table
             ->query(
-                Engagement::query()->whereHas('recipient')
+                Engagement::query()
+                    ->whereHas('recipient')
+                    ->with(['latestEmailMessage', 'latestSmsMessage', 'recipient'])
```
payal-canyon: let's change `recipient` to a `morphWith` that also loads `emailAddresses` and `phoneNumbers` for both `Student` and `Prospect`.
something like this:
```javascript
'recipient' => fn (MorphTo $morphTo) => $morphTo->morphWith([
    Student::class => ['emailAddresses', 'phoneNumbers'],
    Prospect::class => ['emailAddresses', 'phoneNumbers'],
]),
```

app-modules/engagement/src/Models/Engagement.php
```diff
@@ -337,4 +373,22 @@ protected static function booted(): void
             $builder->tap(new LicensedToEducatable('recipient'));
         });
     }
+
+    private function resolveEmailHealthStatus(Educatable $recipient, string $route): EmailHealthStatus
+    {
+        /** @var StudentEmailAddress|ProspectEmailAddress $emailAddress */
+        $emailAddress = $recipient->emailAddresses()->firstWhere('address', $route)
```
payal-canyon: let's drop the `()` from `->emailAddresses()` and `->phoneNumbers()` so it reads the loaded collection.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2563, opened 2026-05-21
[ADVAPP-2665]: Enhance forms and admissions by introducing notifications for forms and admissions

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2665

### Technical Description

> Enhance forms and admissions by introducing notifications for forms and admissions

> No

### Cleanup Tasks

> Yes

---


app-modules/application/tests/Tenant/Jobs/SendApplicationNotificationJobTest.php
```diff
@@ -0,0 +1,330 @@
...
+});
+
+test('a user appearing in all three recipient lists only receives one notification', function () {
+    Notification::fake();
+
+    $sharedUser = User::factory()->licensed(LicenseType::cases())->create();
+    $careTeamRole = CareTeamRole::factory()->create();
+    $student = Student::factory()->create();
+    $student->careTeam()->attach($sharedUser, ['care_team_role_id' => $careTeamRole->id]);
+    // $sharedUser->subscriptions()->create([
```
danharrin: Is there a reason we have commented code here?

danharrin: The form test also names itself "all three recipient lists" but the shared user is only in notification users + care team (no subscription created). No code is commented out here, it's just missing.

danharrin (changes requested):
I have tested this on applications and forms, and it appears that I only get email notifications for applications and not forms, can you confirm if emails are currently working for forms on your side please?

app-modules/application/tests/Tenant/Jobs/SendApplicationNotificationJobTest.php
jayushi-canyon: So subscription relationship was removed because when you relate with careTeam(), `CareTeamObserver` automatically triggers `TriggeredAutoSubscription` which the creates subscription for us. So no need to create once again manually.

danharrin (approved):

PR #2565, opened 2026-05-21
[Chore]: Add zizmor scanning of GitHub Actions

Orrison:
### Ticket(s) or GitHub Issue

- N/A

### Technical Description

Adds a `zizmor` scan of our GitHub Actions with reporting of the SARIF file for our code scanning setup.

Fixes `zizmor` reported issues such as properly tagging, locking images to particular SHAs, etc.

Collapses `ci.yml` into `pull_requests.yml` to avoid having to use insecure `workflow_dispatch` calls.

@danharrin once you reach approval on this I'll need to update the branch protection rules to target the new names. Let me know if you have any questions, concerns, or opinions on the job names as it would be easiest to change it all now.

I chose to run `zizmor` as the `auditor` persona which is one of the highest as my thought is that we should be either addressing or ignoring reports.
We can also decide on which level of reports block merge.

No

### Cleanup Tasks

N/A

---

#### Before contributing and submitting this PR, make sure you have Read, agree, and are compliant with the [contributing guidelines](https://github.com/canyongbs/advisingapp/blob/main/README.md#contributing).


.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,34 @@
...
+        permissions:
+            contents: read
+            security-events: write # required to upload SARIF to code scanning
+        steps:
+            - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+              with:
+                  persist-credentials: false
+            - uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d # v0.5.6
+              with:
+                  advanced-security: true
```
danharrin, line 35: I actually disabled this setting on my version I think, it was pushing issues into the security tab instead of failing the workflow iirc, and the security tab doesn't have much visibility IMO since the errors get lost amongst dependabot errors.

danharrin (changes requested):

.github/workflows/zizmor.yml
Orrison: @danharrin this is actually the better way to work in an org like ours.

This will push them to "code checks" which you can see in the CI scans below (it's green right now since none currently exceed the default thresholds) If the code checks were at or exceed the level threshhold it would be red, would report the problems in a block in the PR (like the sonar cloud one you see above) and block the PR from merging.

Though I do have to finish setting that up once this merges.

The SARIF publishing keeps a much better audit trail of the flagged events than just a PR blocker does.

danharrin (dismissed):

danharrin (approved):

PR #2570, opened 2026-05-22
[ADVAPP-2660]: Enhance semester ordering by introducing the ability for semesters to be automatically added as they become available

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2660

### Technical Description

Adds new settings and scheduled command to import semesters as requested.

No

### Cleanup Tasks

`2026_05_22_enrollment_semester_auto_import_settings_feature`

---


Orrison (changes requested):
@danharrin I didn't get a chance to look at this all yet. But the primary reason I had this in my queue as it was a little reliant on @payal-canyon's PR https://github.com/canyongbs/advisingapp/pull/2554 with the `SisSyncCompleted` event being the thing that should trigger this re-check, instead of the scheduler.

How would you like to handle that? I'll most likely be able to finish remove of https://github.com/canyongbs/advisingapp/pull/2554 tomorrow.

danharrin:
@Orrison I implemented it like this because it mentions in the description "job frequency: once a day" and wasn't linked to any other tickets that would suggest otherwise. It would be helpful if you could review and merge this PR first and then implement the changes in #2554 since this is a much smaller PR that works independently even if the other PR does not make it.

app-modules/student-data-model/database/migrations/2026_05_22_142947_add_enrollment_semester_auto_import_settings.php
```diff
@@ -0,0 +1,61 @@
...
+use App\Features\EnrollmentSemesterAutoImportSettingsFeature;
+use Illuminate\Support\Facades\DB;
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            $this->migrator->add('student_information_system.is_enrollment_semester_auto_import_enabled', false);
+            $this->migrator->add('student_information_system.enrollment_semester_auto_import_default_order', 'first');
```
Orrison: Each of these need the try/catch wrapping like in other settings migrations

app-modules/student-data-model/src/Jobs/AutoImportEnrollmentSemesters.php
```diff
@@ -0,0 +1,109 @@
...
+                $semester->updateQuietly(['order' => $order--]);
+            }
+
+            foreach ($newSemesterNames as $name) {
+                EnrollmentSemester::create([
+                    'name' => $name,
+                    'order' => $order--,
+                ]);
+            }
+        });
```
Orrison, line 114: I think there is perhaps a more efficient way to do this with only two queries and without having to load all into memory. This isn't complete, but something like the following:

```php
    $newCount = $newSemesterNames->count();

    // Shift all existing semesters' order up by the number of new semesters in one query
    EnrollmentSemester::query()
        ->lockForUpdate()
        ->increment('order', $newCount);

    // Bulk insert new semesters with order values 1 through $newCount
    $newSemesters = $newSemesterNames->values()->map(fn ($name, $index) => [
        'name' => $name,
        'order' => $index + 1,
        'created_at' => now(),
        'updated_at' => now(),
    ])->all();

    EnrollmentSemester::insert($newSemesters);
```

Orrison (changes requested):

app-modules/student-data-model/src/Console/Commands/AutoImportEnrollmentSemesters.php
```diff
@@ -0,0 +1,107 @@
...
+use AdvisingApp\StudentDataModel\Enums\EnrollmentSemesterAutoImportDefaultOrder;
+use AdvisingApp\StudentDataModel\Models\Enrollment;
+use AdvisingApp\StudentDataModel\Models\EnrollmentSemester;
+use AdvisingApp\StudentDataModel\Settings\StudentInformationSystemSettings;
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+use Spatie\Multitenancy\Commands\Concerns\TenantAware;
+
+class AutoImportEnrollmentSemesters extends Command
```
Orrison: @danharrin sorry I didn't think of this when I first reviewed it. But we can't have the operation here happen inline in a command.

It needs to be a job dispatched by the scheduler.

If this runs inline via a command like this then every Tenant is going to wait on the others before it to run this command. All actions where possible dispatched by the scheduler should just be queued jobs so that the scheduler is non-blocking.

Orrison (changes requested):

Orrison (dismissed):

Orrison (approved):

PR #2572, opened 2026-05-24
[ADVAPP-2681]: Improve the internal handling of navigation labels

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2681

### Technical Description

- chore: refactor navigation groups to use enum values

No

### Cleanup Tasks

No

---


app-modules/report/src/Filament/Pages/ArtificialIntelligence.php
```diff
@@ -43,14 +43,15 @@
...
+use App\Enums\NavigationGroup;
 use App\Filament\Clusters\ReportLibrary;
 use UnitEnum;
 
 class ArtificialIntelligence extends AiReport
 {
     protected static ?string $cluster = ReportLibrary::class;
 
-    protected static string | UnitEnum | null $navigationGroup = 'Enterprise AI';
+    protected static string | UnitEnum | null $navigationGroup = NavigationGroup::EnterpriseAi;
```
payal-canyon: The main (top-level) navigation groups are correctly converted to the `NavigationGroup` enum (`HasLabel` + `HasIcon`). However, a few pages/resources that live within a cluster are also using this enum for their cluster sub-groups (third-level navigation).

Since the enum implements `HasIcon`, the group icon is now rendered in the cluster sub-menu. For example, the `Enterprise AI` sub-group within the Report Library cluster displays the sparkles icon, which is inconsistent with sibling sub-groups such as Prospects, Students, Users, and Project Management that do not display icons.

Image

To avoid this regression, please create a dedicated label-only enum for each cluster that has sub-groups (HasLabel only, without HasIcon) and update the sub-group references to use those enums instead.

```javascript
enum ReportLibraryNavigationGroup implements HasLabel // no HasIcon
{
    case EnterpriseAi;
    case Prospects;
    case Students;
    case Users;
    ......
}
```

Please apply the same pattern throughout the project. Any cluster that contains its own sub-groups should have a dedicated label-only enum (for example, Communication -> Settings / Templates / Email Templates, AI -> Configuration / Customer Advisor, Authentication -> Local Authentication / Single Sign-On (SSO), etc.), ensuring that all sub-group references are enum-based while remaining icon-free.

Also, could you please review the other two project PRs for the same issue and apply the fix there as well if needed?

app-modules/research/src/Filament/Pages/ManageResearchRequests.php
```diff
@@ -66,7 +67,7 @@ public static function getNavigationItems(): array
     {
         return [
             NavigationItem::make('Research Advisor')
-                ->group('Enterprise AI')
+                ->group(NavigationGroup::EnterpriseAi->getLabel())
```
payal-canyon: `->getLabel()` returns the string 'Enterprise AI', while the other items in this group use the enum directly. Since Filament treats the string and the enum as different groups, one group ends up hiding the items from the other. So, let's remove `->getLabel()`.

app-modules/report/src/Abstract/AiReport.php
```diff
@@ -38,14 +38,21 @@
...
 {
     use HasFiltersForm;
 
     protected string $view = 'report::filament.pages.report';
 
+    public static function getNavigationIcon(): string | BackedEnum | Htmlable | null
+    {
+        return null;
+    }
+
```
payal-canyon: If it isn't needed, could you please remove the method?

app/Filament/Pages/ProductHealth.php
```diff
@@ -36,14 +36,30 @@
...
 use Illuminate\Contracts\Support\Htmlable;
-use ShuvroRoy\FilamentSpatieLaravelHealth\Pages\HealthCheckResults;
+use Illuminate\Support\Facades\Artisan;
+use Spatie\Health\Commands\RunHealthChecksCommand;
 use Spatie\Health\Enums\Status;
 use Spatie\Health\ResultStores\ResultStore;
+use UnitEnum;
 
-class ProductHealth extends HealthCheckResults
+class ProductHealth extends Page
```
payal-canyon: As Dan suggested, the same approach should be applied here as well.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2582, opened 2026-05-29
[ADVAPP-2686]: Improve unified inbox by defaulting view to care team for prospects and students

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2686

### Technical Description

> Improve unified inbox by defaulting view to care team for prospects and students

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
+++ b/app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
@@ -85,6 +85,14 @@ it('can properly filter engagement response type', function () {
         ->assertCanNotSeeTableRecords($smsEngagementResponses);
 });
 
+it('When the component loads, the Care Team filter is already active.', function () {
+    asSuperAdmin();
+
+    livewire(Inbox::class)
+        ->assertTableFilterExists('care_team')
+        ->assertSet('tableFilters.care_team.isActive', true);
+});
+
 it('defaults to the care team filter', function () {
     $user = User::factory()->licensed(LicenseType::cases())->create();
     asSuperAdmin($user);
--- a/app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
+++ b/app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
@@ -34,11 +34,13 @@
 </COPYRIGHT>
 */
 
+use AdvisingApp\Authorization\Enums\LicenseType;
 use AdvisingApp\Engagement\Enums\EngagementResponseType;
 use AdvisingApp\Engagement\Filament\Pages\Inbox;
 use AdvisingApp\Engagement\Models\EngagementResponse;
 use AdvisingApp\Prospect\Models\Prospect;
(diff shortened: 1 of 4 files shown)


app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
```diff
@@ -80,3 +84,26 @@
         ->assertCanSeeTableRecords($emailEngagementResponses)
         ->assertCanNotSeeTableRecords($smsEngagementResponses);
 });
+
+it('defaults to the care team filter', function () {
```
Orrison: Is there a Filament helper / way to check if a specific filter is applied via that filters name/identifier?

If so, I think this test should JUST check that just loading that component has the care team filter applied.

There should be a separate test for the actual functionality of the care team filter. If there isn't one already.

app-modules/engagement/tests/Tenant/Filament/Pages/SentItemsTest.php
```diff
@@ -34,38 +34,36 @@
...
-/**
- * @mixin IdeHelperFormNotificationUser
- */
-class FormNotificationUser extends Pivot
-{
-    use HasUuids;
+use function Pest\Livewire\livewire;
+use function Tests\asSuperAdmin;
+
+it('defaults to the care team filter', function () {
```
Orrison: Same note as the test in the other file.

Orrison (changes requested):

app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
```diff
@@ -80,3 +84,34 @@
         ->assertCanSeeTableRecords($emailEngagementResponses)
         ->assertCanNotSeeTableRecords($smsEngagementResponses);
 });
+
+it('When the component loads, the Care Team filter is already active.', function () {
```
Orrison: The `it` helper for tests we use actually appends "it" to the test name. So we should either, make the message make sense with that. Or use `test`

```suggestion
it('loads the Care Team filter as active when the component first renders', function () {
```

app-modules/engagement/tests/Tenant/Filament/Pages/InboxTest.php
Orrison: Since there is now a test ensuring that the care team filter is applied when it first loads. The name and content of this test should reflect that it is now just a test that the care team filter works properly.

So let's change the name to match that. As well as specifically set the care team filter to true (even though it is the default now, that may not be the case in the future so this should properly cover it)

app-modules/engagement/tests/Tenant/Filament/Pages/SentItemsTest.php
Orrison, line 1: Same comments as the other test file for this one

Orrison (changes requested):

Orrison (approved):

PR #2594, opened 2026-06-04
[ADVAPP-2677]: Add failure retry to Customer Advisors

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2677

### Technical Description

> add retry functionality for QnA advisor messages and enhance error handling

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/ai/routes/widgets.php
+++ b/app-modules/ai/routes/widgets.php
@@ -42,6 +42,7 @@ use AdvisingApp\Ai\Http\Controllers\QnaAdvisors\QnaAdvisorResourceController;
 use AdvisingApp\Ai\Http\Controllers\QnaAdvisors\QnaAdvisorResourcesController;
 use AdvisingApp\Ai\Http\Controllers\QnaAdvisors\RegisterProspectController;
 use AdvisingApp\Ai\Http\Controllers\QnaAdvisors\RequestAuthenticationController;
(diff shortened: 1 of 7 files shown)


danharrin (changes requested):
This isn't quite what we are after I think. I don't think that the institutional advisor has retries like this on the backend, I think they are actually driven by the frontend. Did you have a look at the institutional advisor feature? What retry features does it have on the frontend and backend for failures, rate limits, etc?

payal-canyon:
> This isn't quite what we are after I think. I don't think that the institutional advisor has retries like this on the backend, I think they are actually driven by the frontend. Did you have a look at the institutional advisor feature? What retry features does it have on the frontend and backend for failures, rate limits, etc?

You're right - this is frontend-driven, not a backend retry.
I reviewed the `Institutional Advisor` implementation and found that the backend only provides the rate-limit reset time (`rate_limit_resets_after_seconds` on `AdvisorMessageFinished`). The actual handling happens in `chat.js`, which automatically waits and retries on rate limits and provides a manual `Try Again` option for other errors.

Based on that, I reverted the previous implementation and replicated the existing mechanism instead - the QnA backend now surfaces the rate-limit time the same way, with a dedicated retry endpoint, and the widget auto-retries on rate limits and shows `Try Again` on errors.

PR #2601, opened 2026-06-08
[ADVAPP-2678]: If the Customer Advisor settings are set to NOT allow selection of Model that selection does NOT override previously selected Model for individual Advisors

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2678

### Technical Description

> If the Customer Advisor settings are set to NOT allow selection of Model that selection does NOT override previously selected Model for individual Advisors

> No

### Cleanup Tasks

> No

---


danharrin (changes requested):
Please resolve the merge conflicts

app-modules/ai/src/Filament/Resources/CustomerAdvisors/Pages/EditCustomerAdvisor.php
```diff
@@ -153,6 +153,17 @@ public function form(Schema $schema): Schema
             ]);
     }
 
+    protected function mutateFormDataBeforeFill(array $data): array
```
danharrin, line 156: Is there a specific reason for this method? I do not really think it matters which model is stored in the DB, if `allow_selection_of_model` then we do not read it anyway.

app-modules/ai/src/Http/Controllers/CustomerAdvisors/SendAdvisorMessageController.php
```diff
@@ -107,6 +114,7 @@ public function __invoke(Request $request, GetCustomerAdvisorInstructions $getCu
             $advisor,
             $thread,
             $data['content'],
+            model: $effectiveModel,
```
danharrin: Instead of adding a new param to the job and passing it in, please do the same `$effectiveModel = ` work inside the job too

danharrin (changes requested):
#2609 is currently merging in and adds a new controller and job for customer advisors. When that is merged, please merge `main` into your PR branch and add the model selection fixes there too

app-modules/ai/src/Filament/Resources/CustomerAdvisors/Pages/EditCustomerAdvisor.php
amit-canyon: In tech planning they added: `it should show the force model regardless of what the QnA Advisor has in it’s column` 
That's why I added this method.

danharrin (approved):

PR #2615, opened 2026-06-11
[ADVAPP-2680]: Optimize testing by removing legacy approaches to data generation

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2680

### Technical Description

- Refactor factory definitions to use direct factory calls for related models

No

### Cleanup Tasks

No

---


app-modules/division/database/factories/DivisionFactory.php
```diff
@@ -66,12 +65,4 @@ public function default(): Factory
             ];
         });
     }
-
-    public function configure(): DivisionFactory|Factory
```
danharrin, line 70: Hmm, do we really want to remove this? Or just create a new user instead of picking a random one like the ticket says?

app-modules/meeting-center/database/factories/EventRegistrationFormFactory.php
```diff
@@ -87,20 +87,20 @@ public function configure(): EventRegistrationForm|Factory
...
-                            'status' => $eventRegistrationFormSubmission->attendee_status,
-                            'event_id' => $eventRegistrationForm->event->getKey(),
-                        ]))
-                        ->save()
-                );
+                    'event_attendee_id' => $attendee->getKey(),
+                ]);
+
+                $submission->author()->associate($attendee)->save();
+            });
```
danharrin: Can you please explain the benefits of this refactor? It looks to do the same as before to me, and there wasn't any random existing records being used anyway, right? Did I miss something?

app-modules/team/database/factories/TeamFactory.php
```diff
@@ -52,11 +51,4 @@ public function definition(): array
             'description' => $this->faker->sentence(),
         ];
     }
-
-    public function configure(): TeamFactory|Factory
```
danharrin, line 56: As above I don't think we want to remove this, just create a new division, in the normal way, half the time, right?

app/Observers/UserObserver.php
```diff
@@ -41,13 +41,18 @@
...
 {
     public function saving(User $user): void
     {
         if ($user->isDirty('password')) {
-            $numPreviousPasswords = app(LocalPasswordSettings::class)->getNumPreviousPasswords();
+            try {
+                $numPreviousPasswords = app(LocalPasswordSettings::class)->getNumPreviousPasswords();
+            } catch (Throwable) {
+                $numPreviousPasswords = LocalPasswordSettings::DEFAULT_NUM_PREVIOUS_PASSWORDS;
+            }
```
danharrin: Please explain this change and how it relates to the ticket, it looks like it could be dangerous to discard all `Throwable` especially in this part of the app which is related to password changes.

danharrin (changes requested):

app-modules/meeting-center/database/factories/EventRegistrationFormFactory.php
monali4-canyon: I added this refactor because `UtilizationMetricsApisTest` was failing. The test "checks the API returns Events Created" was hitting infinite recursion. The first fix restructured `EventRegistrationFormFactory` to manually pass `event_attendee_id`, but that only protected that one caller.

So I made change in `EventRegistrationFormSubmissionFactory`. This reuses the same Event from the form instead of creating a new one, breaking the cycle at the source. It's a single-line fix that works for all callers

app-modules/team/database/factories/TeamFactory.php
monali4-canyon: Changed `afterMaking` → `afterCreating` to prevent persisting side-effect records when a Team is only make()'d. Still associates a Division 50% of the time (via `$this->faker->boolean()`), null otherwise — preserving original behavior.

app-modules/division/database/factories/DivisionFactory.php
monali4-canyon: Kept configure() with `afterCreating`. always creates a new user (never picks random existing one) for `createdBy` and `lastUpdatedBy`

app/Observers/UserObserver.php
monali4-canyon: The try-catch was originally added as a workaround during the factory refactor: when `User::factory()->create()` was called in tests, causing getNumPreviousPasswords() to throw. Wrapping it in `try-catch(Throwable)` silently fell back to the default value.

The code has been restored to its original state

app-modules/division/database/factories/DivisionFactory.php
```diff
@@ -69,9 +69,10 @@ public function default(): Factory
 
     public function configure(): DivisionFactory|Factory
     {
-        return $this->afterMaking(function (Division $division) {
-            $division->createdBy()->associate($this->faker->randomElement([User::inRandomOrder()->first(), null]));
-            $division->lastUpdatedBy()->associate($this->faker->randomElement([User::inRandomOrder()->first(), null]));
+        return $this->afterCreating(function (Division $division) {
+            $division->createdBy()->associate(User::factory()->create());
```
danharrin: I think we want this to not overwrite an exiting user that is attached to these relationships if the caller of the factory passes one - so only associates if there is no existing `createdBy` or `lastUpdatedBy` please.

app-modules/meeting-center/database/factories/EventRegistrationFormSubmissionFactory.php
```diff
@@ -57,8 +57,10 @@ public function definition(): array
         return [
             'attendee_status' => $this->faker->randomElement(EventAttendeeStatus::class),
             'submitted_at' => $this->faker->dateTime(),
-            'form_id' => EventRegistrationForm::inRandomOrder()->first() ?? EventRegistrationForm::factory()->create(),
-            'event_attendee_id' => EventAttendee::inRandomOrder()->first() ?? EventAttendee::factory()->create(),
+            'form_id' => EventRegistrationForm::factory(),
+            'event_attendee_id' => fn (array $attributes) => EventAttendee::factory()->state([
+                'event_id' => EventRegistrationForm::find($attributes['form_id'])?->event_id,
+            ])->create()->getKey(),
```
danharrin: Do we definitely want the `create()` here? Will the factory not do that for us if we return the factory instance, like on a normal prop?

app-modules/team/database/factories/TeamFactory.php
```diff
@@ -55,8 +55,11 @@ public function definition(): array
 
     public function configure(): TeamFactory|Factory
     {
-        return $this->afterMaking(function (Team $team) {
-            $team->division()->associate($this->faker->randomElement([Division::inRandomOrder()->first(), null]));
+        return $this->afterCreating(function (Team $team) {
+            if ($this->faker->boolean()) {
+                $team->division()->associate(tap(Division::factory()->make())->save());
```
danharrin, line 60: Let's only [assume a division if one is not already associated.

danharrin (changes requested):

danharrin (changes requested):
I'm sorry to keep going back about this, but `createdBy()->exists()` might actually silently hide bugs. The `exists()` check there will check if it exists in the DB, but we don't care about that. It might not be saved yet, and we are silently overwriting it if that is the case. A simple `$division->createdBy` (etc) check will suffice and pass even if the model doesn't exist yet.

Also, is there a reason we should do this in `afterCreating()` instead of `afterMaking()`? `afterMaking()` seems more correct to be honest

monali4-canyon:
If I keep `afterMaking()` then 2 test cases are failing in `app-modules/authorization/tests/Tenant/Feature/Filament/Resources/Users/ListUsersTest.php` 
- can filter users by multiple teams
- it filters users based on team

In DivisionFactory if we use afterMaking, every time Division::factory()->make() is called from TeamFactory, it triggers:
User::factory()->create();  // createdBy
User::factory()->create();  // lastUpdatedBy

This creates 2 extra unexpected Users in the database for each Team that randomly gets a Division assigned. The `ListUsers` tests then fail

danharrin (approved):

PR #2619, opened 2026-06-15
[ADVAPP-2722]: Update the appointment availability is considered

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2722

### Technical Description

> Personal booking will take into account personal booking availability + actual calendar availability + Out of Office.
> Group booking should take into account group booking availability + actual calendar availability + Out of office(Personal booking availability will not be considered).
> Created migration to remove working_hours column from users table.
> Modified test cases to ensure they correctly handle scenarios based on office hours.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_06_15_working_hous_feature.md

---


danharrin (changes requested):
I think the feature flag is not relevant here. Feature flags are only relevant to prevent deployment errors before migrations are run, but if the codebase is already not using the deleted columns, I think there is no point in the feature flag? Am I wrong?

danharrin (approved):

PR #2622, opened 2026-06-16
[ADVAPP-2701]: Introduce form and admissions versions

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2701

### Technical Description

Introduces versioning for forms and admissions so that they can be edited after they have submissions.

No

### Cleanup Tasks

`2026_06_15_form_versioning_feature `

---


app-modules/form/src/Filament/Resources/Forms/Pages/EditForm.php
```diff
@@ -54,12 +59,43 @@ class EditForm extends EditRecord
...
+        if (! FormVersioningFeature::active()) {
+            return parent::handleRecordUpdate($record, $data);
+        }
+
+        /** @var Form $record */
+        $newVersion = app(CreateFormVersion::class)->execute($record, $data);
+
+        $this->record = $newVersion;
+
+        app(SaveSubmissibleFieldsFromContent::class)->execute($newVersion, $this->versioningFormData);
```
payal-canyon: `CreateFormVersion` commits its own transaction before `SaveSubmissibleFieldsFromContent` runs. As a result, if saving the fields fails, the previous version has already been archived and a broken, field-less version could become live in the widget.
Could we wrap both operations in a single `DB::transaction()` so they either succeed or fail together? That way, if an error occurs while saving the fields, the entire operation is rolled back cleanly and the live form is never left in a partially updated state.
The same concern appears to apply in `EditApplication` as well, so could we make the same change there? WDYT?

app-modules/application/database/migrations/2026_06_15_143001_add_versioning_to_applications_table.php
```diff
@@ -0,0 +1,85 @@
...
+use Illuminate\Support\Facades\DB;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('applications', function (Blueprint $table) {
+                $table->uuid('root_id')->nullable()->after('id');
```
payal-canyon: Small thing: `->after('id')` is a no-op on PostgreSQL, so we can remove it.

app-modules/form/database/migrations/2026_06_15_143000_add_versioning_to_forms_table.php
```diff
@@ -0,0 +1,80 @@
...
+use Illuminate\Support\Facades\DB;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('forms', function (Blueprint $table) {
+                $table->uuid('root_id')->nullable()->after('id');
```
payal-canyon: Same here - `->after('id')` is a no-op on PostgreSQL, so we can remove it.

app-modules/application/src/Actions/CreateApplicationVersion.php
```diff
@@ -0,0 +1,81 @@
...
+            $oldVersion->archive();
+
+            $newVersion = new Application();
+            $newVersion->fill($newData);
+            $newVersion->root_id = $oldVersion->root_id;
+            $newVersion->notify_to_care_team = $oldVersion->notify_to_care_team;
+            $newVersion->notify_to_subscribers = $oldVersion->notify_to_subscribers;
+            $newVersion->notify_via_app = $oldVersion->notify_via_app;
+            $newVersion->notify_via_email = $oldVersion->notify_via_email;
+            $newVersion->save();
```
payal-canyon, line 58: Applications can embed images in the rich content editor. It copies the content JSON, but it doesn't carry over the associated media. As a result, the new version ends up with no media records while its content still references the image UUIDs from the previous version. I confirmed this on a test application: the newly active version had `media = 0`, while the archived version retained the media records.

The tricky part is that this isn't immediately obvious after editing. The saved content includes pre-signed S3 URLs, which remain valid for roughly `30–35 minutes`. Because of that, the images continue to appear correctly during a quick verification, but once those URLs expire, the images break since the new version has no media available to re-resolve them.

I think we should copy the content collection media to the new version and remap the image IDs within the content so they point to the newly copied media records.

Forms don't appear to be affected since they don't support embedded attachments.

payal-canyon (changes requested):

payal-canyon (changes requested):
One edge case I think we misses: wizard (multi-step) applications. Step content uses the same editor with the attach-files button enabled, and those images are stored as media on the `ApplicationStep` model rather than on the `Application`. Since `copyMedia` only copies the application-level content media (and early-returns when that's empty, which it always is for wizards), a step image isn't carried over - the new version's step ends up with media = 0 while its content still references the old step's image UUID. I confirmed it on a multi-step test app: after editing, the active version's step had no media, while the archived version kept it.

Could we extend `copyMedia` to also loop over the steps - copy each step's content-collection media onto the matching new step and remap that step's content UUIDs?

payal-canyon (approved):

PR #2625, opened 2026-06-17
[ADVAPP-2716]: Users without the ability to update settings can change the default care team roles

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2716

### Technical Description

- disable default toggle for care team roles based on user permissions
- Edit pages(`EditStudentCareTeamRole` and `EditProspectCareTeamRole`): Added missing `beforeSave()` to clear existing defaults before save, preventing unique constraint violation. Fixed Student edit hint querying wrong type (Prospect → Student).
- Create pages(`CreateStudentCareTeamRole` and `CreateProspectCareTeamRole`): Added ->where('type', ...) filter to beforeCreate() so clearing defaults is scoped to the correct type.

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/care-team/src/Filament/Resources/StudentCareTeamRoles/Pages/ListStudentCareTeamRoles.php
+++ b/app-modules/care-team/src/Filament/Resources/StudentCareTeamRoles/Pages/ListStudentCareTeamRoles.php
@@ -64,6 +64,7 @@ class ListStudentCareTeamRoles extends ListRecords
                     ->searchable(),
                 ToggleColumn::make('is_default')
                     ->label('Default')
+                    ->disabled(! auth()->user()->can('settings.*.update'))
                     ->beforeStateUpdated(
                         fn (CareTeamRole $record) => CareTeamRole::where('id', '!=', $record->getKey())
                             ->where('type', CareTeamRoleType::Student)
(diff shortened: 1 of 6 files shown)


app-modules/care-team/src/Filament/Resources/StudentCareTeamRoles/Pages/ListStudentCareTeamRoles.php
```diff
@@ -64,6 +64,7 @@ public function table(Table $table): Table
                     ->searchable(),
                 ToggleColumn::make('is_default')
                     ->label('Default')
+                    ->disabled(! auth()->user()->can('settings.*.update'))
```
payal-canyon, line 67: Looks good overall, but I found a few existing bugs while testing:
The Student/Prospect Edit pages (`EditStudentCareTeamRole` and `EditProspectCareTeamRole`)are missing the `beforeSave()` hook that clears the existing default (so saving a role with `Default` on throws a `care_team_roles_type_is_default_unique` violation - the Create/List pages already handle this), and the Student edit page's `you are replacing it` hint mistakenly queries `CareTeamRoleType::Prospect` instead of `Student`, so it shows the wrong type's current default. 

Could you please address these issues in this PR as well?
Also, once they're fixed, could you please update the PR description to include these bug fixes so the scope of the changes is clearly documented?
Thanks!

payal-canyon (approved):

PR #2630, opened 2026-06-18
Fix lint issues

jayushi-canyon:
### Ticket(s) or GitHub Issue

- Lint fixes

### Technical Description

> Fix lint issues. 
> Solve around 15 lint issues
> Since lint now has 1000 lint errors limit so exact lint errors amount were not known.

> No

### Cleanup Tasks

>N/A.

---

### DIFF
--- a/app-modules/in-app-communication/database/migrations/2025_03_20_125908_data_remove_realtime_chat_api_permission.php
+++ b/app-modules/in-app-communication/database/migrations/2025_03_20_125908_data_remove_realtime_chat_api_permission.php
@@ -42,7 +42,7 @@ return new class () extends Migration {
     use CanModifyPermissions;
 
     /**
-     * @var array<string> $permissionsToDelete
+     * @var array<string, string> $permissionsToDelete
      */
     private array $permissionsToDelete = [
         'realtime_chat.*.view' => 'Realtime Chat',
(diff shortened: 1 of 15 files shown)


app-modules/in-app-communication/database/migrations/2025_03_20_125908_data_remove_realtime_chat_api_permission.php
```diff
@@ -41,10 +41,16 @@
 return new class () extends Migration {
     use CanModifyPermissions;
 
+    /**
+     * @var array<string> $permissionsToDelete
```
payal-canyon: This is a string-keyed associative array, so the precise type is array<string, string>. 
`/** @var array<string, string> $permissionsToDelete */`

app-modules/interaction/src/Filament/Actions/BulkCreateInteractionAction.php
```diff
@@ -169,7 +169,7 @@ public static function make(): BulkAction
                     });
 
                     DB::commit();
-                } catch (Exception $e) {
+                } catch (Exception $exce) {
```
payal-canyon: For better readability, could we rename `$exce` to `$exception`? It would also keep the naming consistent with other parts of the codebase.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2659, opened 2026-06-29
[ADVAPP-2750]: Notify and restrict access for expired subscriptions

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2750

### Technical Description

> Add subscription management features and expiration banner. Introduced `TenantSubscriptionStatus` enum to manage tenant subscription states. Updated `SyncTenantController` to handle subscription status and expiration banner text. Created `SubscriptionExpiredBanner` component to display expiration notifications. Added `TenantExpirationSettings` for managing expiration banner text. Implemented `SubscriptionAwareDomainTenantFinder` to restrict access for expired tenants. Updated database migrations to include subscription status and expiration settings.

> No

### Cleanup Tasks

> Yes. `.cleanup-tasks/2026_07_08_subscription_expiration_feature.md`

---

### DIFF
--- a/app/Http/Controllers/Tenants/SyncTenantController.php
+++ b/app/Http/Controllers/Tenants/SyncTenantController.php
@@ -42,6 +42,7 @@ use App\DataTransferObjects\LicenseManagement\LicenseData;
 use App\DataTransferObjects\LicenseManagement\LicenseLimitsData;
 use App\DataTransferObjects\LicenseManagement\LicenseSubscriptionData;
 use App\Enums\SubscriptionStatus;
+use App\Features\SubscriptionExpirationFeature;
 use App\Http\Requests\Tenants\SyncTenantRequest;
 use App\Jobs\UpdateTenantLicenseData;
 use App\Models\Tenant;
@@ -61,15 +62,17 @@ class SyncTenantController
 
         dispatch_sync(new UpdateTenantLicenseData($tenant, $licenseData));
 
-        if (filled($subscriptionStatus = $request->validated('subscriptionStatus'))) {
-            $tenant->subscription_status = SubscriptionStatus::from($subscriptionStatus);
-            $tenant->save();
-        }
+        if (SubscriptionExpirationFeature::active()) {
+            if (filled($subscriptionStatus = $request->validated('subscriptionStatus'))) {
+                $tenant->subscription_status = SubscriptionStatus::from($subscriptionStatus);
+                $tenant->save();
+            }
 
-        if (filled($bannerText = $request->validated('expirationBannerText'))) {
-            $settings = app(TenantExpirationSettings::class);
-            $settings->period_2_banner_text = $bannerText;
-            $settings->save();
+            if (filled($bannerText = $request->validated('expirationBannerText'))) {
+                $settings = app(TenantExpirationSettings::class);
+                $settings->period_2_banner_text = $bannerText;
+                $settings->save();
+            }
         }
 
         $tenant->execute(function () use ($request) {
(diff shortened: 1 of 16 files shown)


Orrison (changes requested):
@payal-canyon I think it would be much more efficient for me to have the Olympus related ticket settled and approved first before I dug too much into this one. As there may be changes you need to make from my feedback from that one anyway.

So once we get that one settled approved / merged, please re-request my review here.

app/Http/Controllers/Tenants/SyncTenantController.php
```diff
@@ -59,6 +61,17 @@ public function __invoke(SyncTenantRequest $request, Tenant $tenant): JsonRespon
...
+        if (filled($subscriptionStatus = $request->validated('subscriptionStatus'))) {
+            $tenant->subscription_status = SubscriptionStatus::from($subscriptionStatus);
+            $tenant->save();
+        }
+
+        if (filled($bannerText = $request->validated('expirationBannerText'))) {
+            $settings = app(TenantExpirationSettings::class);
+            $settings->period_2_banner_text = $bannerText;
+            $settings->save();
+        }
```
Orrison: I think we need a feature flag protecting this to make sure the migrations have run.

It needs extend `LandlordAbstractFeatureFlag` since these columns are based on the landlord database.

app/Http/Requests/Tenants/SyncTenantRequest.php
```diff
@@ -86,6 +88,8 @@ public function rules(): array
             'subscription.partnerName' => ['required', 'string'],
             'subscription.startDate' => ['required', 'string'],
             'subscription.endDate' => ['required', 'string'],
+            'subscriptionStatus' => ['required', Rule::enum(SubscriptionStatus::class)],
```
Orrison: I think this needs to NOT be required at first, to prevent an issue during the time period when these are deploying and Olympus is deploying.

Then making it required can be a part of the cleanup task along with the feature flag.

Additionally, do we think we should have this data synced in during create as well? We may need an extra small change on the Olympus side to support that.

Orrison (changes requested):

app/Http/Requests/Tenants/SyncTenantRequest.php
payal-canyon: Hey @Orrison !
I made `subscriptionStatus` nullable and added flipping it back to required as part of the feature flag cleanup task.

Regarding syncing on create, Olympus's create job doesn't currently send this value, and since the `subscription_status` column defaults to `active`, a new tenant is already in the correct state from the start (the first sync reconciles it afterward). Because of that, I wasn't sure the additional Olympus change was worth it unless there's a scenario where a tenant can actually be created in a non-`active` state.

If you still feel we should include the change, just let me know and I'm happy to add it.

app/Http/Controllers/Tenants/SyncTenantController.php
```diff
@@ -59,6 +62,19 @@ public function __invoke(SyncTenantRequest $request, Tenant $tenant): JsonRespon
...
+            if (filled($bannerText = $request->validated('expirationBannerText'))) {
+                $settings = app(TenantExpirationSettings::class);
+                $settings->period_2_banner_text = $bannerText;
+                $settings->save();
+            }
+        }
+
         $tenant->execute(function () use ($request) {
             app(SyncTenantSmartPrompts::class)->execute($request);
         });
```
Orrison: Seeing this all together now since we are updating so many different things I think we should wrap this all in a try/catch and DB transaction.

I do think we will need two different transactions though, one from the Tenant and one for the Landlord. I think that should be fine, we will just need to call the `DB::` comments twice with the different connections.

Orrison (changes requested):

Orrison (approved):

PR #2662, opened 2026-06-30
[ADVAPP-2757]: Refactor and improve access to reporting

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2757

### Technical Description

> Refactor and improve access to reporting

> No

### Cleanup Tasks

> Yes, Feaature Flag: `ReportingFeature`

---


app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -0,0 +1,294 @@
...
+                    ->searchable()
+                    ->options(
+                        fn (): array => $usersQuery
+                            ->limit(50)
+                            ->pluck('name', 'id')
+                            ->all(),
+                    )
+                    ->getSearchResultsUsing(
+                        fn (string $search): array => $usersQuery
+                            ->where('name', 'ilike', "%{$search}%")
```
Orrison: We should use the lower strategy we use through out the app here

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -0,0 +1,294 @@
...
+                    ->searchable()
+                    ->options(
+                        fn (): array => $teamsQuery
+                            ->limit(50)
+                            ->pluck('name', 'id')
+                            ->all(),
+                    )
+                    ->getSearchResultsUsing(
+                        fn (string $search): array => $teamsQuery
+                            ->where('name', 'ilike', "%{$search}%")
```
Orrison: We should use the lower strategy we use through out the app here

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -0,0 +1,294 @@
...
+            ->action(function (array $record, array $data): void {
+                $this->syncUserAccess($record['report_key'], $data['users'] ?? []);
+                $this->syncTeamAccess($record['report_key'], $data['teams'] ?? []);
+
+                $this->resetTable();
+
+                Notification::make()
+                    ->success()
+                    ->title('Report access updated')
+                    ->send();
```
Orrison: I would wrap this in a try/catch and DB transaction, rolling back and reporting an error if one comes up and making a user Notification as appropriate.

app-modules/report/src/Enums/ReportAccessKey.php
```diff
@@ -0,0 +1,216 @@
...
+            self::StudentTaskManagement => 'Students',
+            self::ProspectActionCenter,
+            self::ProspectReport,
+            self::ProspectCaseReport,
+            self::ProspectInteractionReport,
+            self::ProspectMessagesDetailReport,
+            self::ProspectMessagesOverviewReport,
+            self::ProspectTaskManagement => 'Prospects',
+            self::UserLoginActivity => 'Users',
+        };
```
Orrison, line 181: In this match statement, and the ones below, can we put a space (line break) between groupings. If the formatter won't let us do that, it's okay. But That would make it easier to read here.

app-modules/report/src/Models/ReportTeamAccess.php
```diff
@@ -0,0 +1,57 @@
...
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Report\Models;
+
+use AdvisingApp\Team\Models\Team;
+use App\Models\BaseModel;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class ReportTeamAccess extends BaseModel
```
Orrison: Let's make this model auditable

app-modules/report/src/Models/ReportUserAccess.php
```diff
@@ -0,0 +1,57 @@
...
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Report\Models;
+
+use App\Models\BaseModel;
+use App\Models\User;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+class ReportUserAccess extends BaseModel
```
Orrison: Let's make this model auditable

app-modules/report/src/Support/ReportAccess.php
```diff
@@ -0,0 +1,107 @@
...
+*/
+
+namespace AdvisingApp\Report\Support;
+
+use AdvisingApp\Report\Enums\ReportAccessKey;
+use AdvisingApp\Report\Models\ReportTeamAccess;
+use AdvisingApp\Report\Models\ReportUserAccess;
+use App\Models\User;
+
+class ReportAccess
```
Orrison: IMO I think these should just be methods on the `ReportAccessKey` ENUM itself.

`userCanAccessPage` can be replaced by just doing `ReportAccessKey::fromPageClass(...)::userCanAccess(...)`

app-modules/report/src/Support/ReportAccess.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        if (blank($user->team_id)) {
+            return false;
+        }
+
+        return ReportTeamAccess::query()
+            ->where('report_key', $key->value)
+            ->where('team_id', $user->team_id)
+            ->exists();
```
Orrison: Is there a way to do this all in one query with joins?

app-modules/report/src/Support/ReportAccess.php
```diff
@@ -0,0 +1,107 @@
...
+        $teamUserIds = $teamIds->isEmpty()
+            ? collect()
+            : User::query()
+                ->whereIn('team_id', $teamIds)
+                ->pluck('id');
+
+        return $directUserIds
+            ->merge($teamUserIds)
+            ->unique()
+            ->count();
```
Orrison: Is it possible to do this all just in one query?

Orrison (changes requested):

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -0,0 +1,317 @@
...
+    public function manageAction(): Action
+    {
+        $usersQuery = User::query()
+            ->tap(new WithoutAnyAdmin())
+            ->orderBy('name');
+
+        $teamsQuery = Team::query()
+            ->orderBy('name');
+
+        return Action::make('manage')
```
Orrison, line 168: I think we should add some authorization to this action to make sure that they have the reporting update permission in order to use it.

There could be people that only have the ability to view this, and NOT be able to edit.

So I think we should check the update permission.

app-modules/authorization/tests/Tenant/Filament/Pages/ReportingTest.php
Orrison, line 1: When we add authorization to the update action we should test that here as well.

app-modules/authorization/tests/Tenant/Filament/Pages/ReportingTest.php
```diff
@@ -0,0 +1,212 @@
...
+    $user = User::factory()->create();
+
+    $user->givePermissionTo('reporting.view-any');
+
+    actingAs($user);
+
+    get(Reporting::getUrl())->assertSuccessful();
+});
+
+it('only lists reports that are available to the tenant', function () {
```
Orrison: Honestly, I think it would be good here to add a full dataset to this test to ensure that every single ENUM case / Report correctly ensures that the correct licenses, addons, and settings are enabled for the Tenant for it to appear in the list here.

Copilot would be good at generating all of that.

Just spitballing, but the dataset could consist of a function that does everything to enable the settings needed for the feature, and then the ENUM case.

Then the test would set everything to be 0 licenses, turn off all addons, ensure that the report is not listed in the table, then run the function, then confirm that the report/case IS now listed in the table.

I would also use the methods for asserting a that the filament table row exists directly, instead of `assertSee`, `assertSee` isn't great as it really ONLY validates it can see certain text on the page. Whereas there are much better methods to assert records on a table.

app-modules/authorization/tests/Tenant/Filament/Pages/ReportingTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+    $user = User::factory()->create();
+    $user->givePermissionTo('reporting.view-any');
+
+    actingAs($user);
+
+    livewire(Reporting::class)
+        ->searchTable(ReportAccessKey::ProjectReport->getName())
+        ->assertSee(ReportAccessKey::ProjectReport->getName())
+        ->assertDontSee(ReportAccessKey::UserLoginActivity->getName());
```
Orrison: I would try the methods for asserting filament table rows, not `assertSee` if possible. If you can check everywhere you are using `assertSee` and `assertDontSee` in this PR and see if you can replace it with methods that actually validate table row presence, etc.

Orrison (changes requested):

Orrison (approved):

PR #2668, opened 2026-07-01
[ADVAPP-2752]: Introduce the navigation group Scheduling

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2752

### Technical Description

> Renamed 'My Appointemnts' to 'My Schedule'
> Created new subnavigation named 'Scheduling' and move 'My Schedule', 'Group Appointments' and 'Events' under it.
> Moved 'Personal Booking Page' and 'Connected Accounts' from 'Profile settings' in new tab with gear icon in 'My schedule' page.
> Moved 'Configuration' tab to new gear tab in 'Group Appointments' Page. 

> No

### Cleanup Tasks

> N/A

---


app-modules/meeting-center/src/Filament/Pages/SharedCalendar.php
```diff
@@ -61,22 +61,21 @@
...
     use InteractsWithForms;
     use InteractsWithTable;
 
-    protected static ?int $navigationSort = 10;
+    protected static ?int $navigationSort = 20;
 
     protected string $view = 'meeting-center::filament.pages.shared-calendar';
 
-    protected static ?string $cluster = GroupAppointments::class;
+    protected static string | UnitEnum | null $navigationGroup = NavigationGroup::Scheduling;
```
payal-canyon, line 76: It looks like the `PermissionDescription` file still says `...the My Appointments feature in the primary navigation group CRM...`. Since this text is displayed in the roles/permissions UI, it now contradicts the changes in this ticket.

Could we update it to reference `My Schedule` / `Scheduling` instead?

app-modules/meeting-center/resources/views/filament/pages/list-calendar-events.blade.php
```diff
@@ -112,5 +125,10 @@
         <div>
             @livewire(CalendarEventWidget::class)
         </div>
+    @elseif ($viewType === 'settings')
```
payal-canyon, line 128: I think it would be worth adding a render test to verify that the relocation is actually surfaced. For example:
```php
livewire(ListCalendarEvents::class)
    ->set('viewType', 'settings')
    ->assertSeeLivewire(ManagePersonalBookingPage::class)
    ->assertSeeLivewire(ConnectedAccounts::class);
```
(and the equivalent for `SharedCalendar` -> `ListBookingGroups`).
That way, if a future Blade change accidentally removes or relocates these components, the test will catch it.

payal-canyon (changes requested):
@Orrison , I'd like your opinion on this one.

I think it's probably out of scope, but I still wanted to check with you.

In `app/Filament/Pages/ManageLicenseSettings.php`, the `scheduleAndAppointments` addon toggle is labeled `My Appointments`. Since this is the addon name rather than the navigation item, I'm not sure whether it should be updated as part of this ticket.

Do you think we should rename `My Appointments` here as well, or leave it as-is?

Orrison:
> @Orrison , I'd like your opinion on this one.
> 
> I think it's probably out of scope, but I still wanted to check with you.
> 
> In `app/Filament/Pages/ManageLicenseSettings.php`, the `scheduleAndAppointments` addon toggle is labeled `My Appointments`. Since this is the addon name rather than the navigation item, I'm not sure whether it should be updated as part of this ticket.
> 
> Do you think we should rename `My Appointments` here as well, or leave it as-is?

This is a great callout, I think at this moment I would prefer we leave it out of scope for this ticket and take care of it in another one as changing it probably warrants changing it in Olympus as well. I'll bring it up with Product / make tickets to handle it at a later date.

payal-canyon (approved):

PR #2673, opened 2026-07-02
[ADVAPP-2765]: Implement the resource hub portal article categorization framework

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2765

### Technical Description

> Implements the resource hub category page by consuming shared portal UI components from 'canyongbs/common'.

> No

### Cleanup Tasks

> N/A

---


danharrin (changes requested):
I noticed that the article titles are not aligned to the centre of the list rows, looks like there is too much space below, this isn't a problem in Aiding App since there are tags below which fill the space, but we should fix this problem, you probably need to change the code in common to fix the class names.

I think this is a great opportunity to improve the user experience of the portal across both apps at the same time, especially regarding data loading states.

When I click on a category from the homepage, most of the page content disappears while the category data is being loaded from the API. I think we instead want to implement a top progress bar in the portal like [nprogress](https://ricostacruz.com/nprogress) that acts as a loading state.

When clicking on the tabs between "All Articles", "Featured" and "Most Viewed", there should be no loading state as we should load all of the articles in from the API when the category page loads initially. You will probably need to modify the API in order to fetch this data with one call, you should reach out to Amit or Jayushi for help with this if they are available, if not then someone else on the team.

In addition, ensure that when switching between pagination pages of articles on the category page, the page content does not disappear please. No need for a full page loading bar there, but the pagination page button that was clicked should become disabled with a little less opacity while the new page loads, unless you have a better idea.

nency-canyon:
> I noticed that the article titles are not aligned to the centre of the list rows, looks like there is too much space below, this isn't a problem in Aiding App since there are tags below which fill the space, but we should fix this problem, you probably need to change the code in common to fix the class names.
> 
> I think this is a great opportunity to improve the user experience of the portal across both apps at the same time, especially regarding data loading states.
> 
> When I click on a category from the homepage, most of the page content disappears while the category data is being loaded from the API. I think we instead want to implement a top progress bar in the portal like [nprogress](https://ricostacruz.com/nprogress) that acts as a loading state.
> 
> When clicking on the tabs between "All Articles", "Featured" and "Most Viewed", there should be no loading state as we should load all of the articles in from the API when the category page loads initially. You will probably need to modify the API in order to fetch this data with one call, you should reach out to Amit or Jayushi for help with this if they are available, if not then someone else on the team.
> 
> In addition, ensure that when switching between pagination pages of articles on the category page, the page content does not disappear please. No need for a full page loading bar there, but the pagination page button that was clicked should become disabled with a little less opacity while the new page loads, unless you have a better idea.

- Fixed article title alignment.
- Added NProgress bar on category navigation so that old content stays visible while new category loads.
- Added `loadingPage` prop to `Pagination`, button gets less opacity and all buttons are disabled until the page resolves.
- Regarding the Featured tab, it exists in `Aiding App` only. The featured column doesn't exist in the `Advising App` database, so it was essentially dead UI. Removed it from the category page.

app-modules/application/database/migrations/2026_06_15_143001_add_versioning_to_applications_table.php
danharrin, line 1: Please look at the diff of the PR and remove any unrelated changes

danharrin (changes requested):

app-modules/application/database/migrations/2026_06_15_143001_add_versioning_to_applications_table.php
nency-canyon: I have reverted those changes. Please check.

danharrin (changes requested):
When I navigate to the category page, all the home page content still seems to disappear, is there a way we can avoid this so that page navigation feels smoother?

When directly accessing the category page, we don't need a loading bar there either since we already have a spinner. This is the same for the homepage loading which is immediate after the spinner completes.

I think the approach we have for data fetching currently doesn't seem to be fit for purpose, and it would be great if we could clean this up.

nency-canyon:
> When I navigate to the category page, all the home page content still seems to disappear, is there a way we can avoid this so that page navigation feels smoother?
> 
> When directly accessing the category page, we don't need a loading bar there either since we already have a spinner. This is the same for the homepage loading which is immediate after the spinner completes.
> 
> I think the approach we have for data fetching currently doesn't seem to be fit for purpose, and it would be great if we could clean this up.

- Replaced watch(route) with onMounted in App.vue so the portal config is fetched once on app boot, eliminating the blank flash during navigation.
- Removed NProgress from ViewCategory.vue since the global AppLoading spinner already covers the loading state.
- Optimized category loading by fetching all_articles and most_viewed_articles in a single initial API call for instant tab switching. Pagination now fetches only the requested page while keeping existing content visible, with the active page button disabled during loading.

danharrin:
That isn't quite what I was asking for when I was talking about refactoring our approach to data fetching.

I did some research into the best approaches myself and refactored the portal to use them. The portal now uses **Vue Router data loaders** (backed by Pinia Colada) instead of calling `axios`/`consumer()` inside each page. The way it fits together is: you define a loader for the page's data in `Pages/loaders.js` (using `defineColadaLoader` + the shared `apiGet` helper from `Services/api.js`), attach it to the route in portal.js under `meta.loaders`, and then the page component just reads the result with `const { data } = useYourData()` and stays purely presentational (no manual loading spinners or fetch calls).

For any new page in a future PR, please follow this exact pattern: add the route + its loader, and let the loader do the fetching: the loading progress bar, "keep the old page visible until data is ready", caching, and scroll-to-top are all handled globally, so you don't need to build those per page. The easiest way in is to copy how Home.vue and ViewCategory.vue (plus their entries in `loaders.js` and portal.js) are wired up and adapt them. If anything's unclear, ping me before adding a new patterns please, since we're trying to keep all data fetching in loaders from now on.

danharrin (approved):

nency-canyon:
> That isn't quite what I was asking for when I was talking about refactoring our approach to data fetching.
> 
> I did some research into the best approaches myself and refactored the portal to use them. The portal now uses **Vue Router data loaders** (backed by Pinia Colada) instead of calling `axios`/`consumer()` inside each page. The way it fits together is: you define a loader for the page's data in `Pages/loaders.js` (using `defineColadaLoader` + the shared `apiGet` helper from `Services/api.js`), attach it to the route in portal.js under `meta.loaders`, and then the page component just reads the result with `const { data } = useYourData()` and stays purely presentational (no manual loading spinners or fetch calls).
> 
> For any new page in a future PR, please follow this exact pattern: add the route + its loader, and let the loader do the fetching: the loading progress bar, "keep the old page visible until data is ready", caching, and scroll-to-top are all handled globally, so you don't need to build those per page. The easiest way in is to copy how Home.vue and ViewCategory.vue (plus their entries in `loaders.js` and portal.js) are wired up and adapt them. If anything's unclear, ping me before adding a new patterns please, since we're trying to keep all data fetching in loaders from now on.

Thanks for the detailed explanation! That makes a lot more sense now.

I understand the pattern you're aiming for, and I'll follow this approach for future pages instead of fetching data directly in the components. Thanks for taking the time to explain it and researching on this!

PR #2675, opened 2026-07-06
[ADVAPP-2766]: Implement the resource hub portal article detail page

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2766

### Technical Description

> Refactors the resource hub portal ViewArticle and ViewCategory pages to use shared UI components from `canyongbs/common`

> No

### Cleanup Tasks

> N/A

---


portals/resource-hub/src/Pages/Home.vue
danharrin, line 1: Please explain the changes to this file

danharrin (changes requested):

portals/resource-hub/src/Pages/Home.vue
nency-canyon: Since you're already on the root page, there's nowhere to navigate back to.The breadcrumb shows "Home" as a non-clickable current crumb (not a link, since you're already on the home page).

That said, this is a UI decision. If you'd prefer to keep it for visual consistency with other pages, I can revert that change.

danharrin: Please revert it, otherwise removing the breadcrumbs causes layout shift when switching pages, and the user might not be aware that they are on a top-level page.

And the description of the hero was removed? Please put it back?

danharrin (changes requested):

portals/resource-hub/src/Pages/Home.vue
nency-canyon: Addressed your feedback, Please review and let me know if you find any issue. Thanks!

danharrin (approved):

PR #2682, opened 2026-07-10
[ADVAPP-2764]: Implement resource hub portal search results page

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2764

### Technical Description

> Implements the resource hub portal search page by consuming shared portal UI components from 'canyongbs/common'.

> No

### Cleanup Tasks

> N/A

---


danharrin (changes requested):
1) Please ensure that the search bar in the navbar works: when you search and then click "enter", it should redirect you to the homepage with that search query selected. The loading bar should be visible when the switch to the homepage is happening, and then the search results loading state should be visible while the results are being fetched.

2) Please hide the search bar when the current router page is the homepage / view category page, so that there is not a duplicate search bar on those pages

3) Please add the hero search bar to the view category page like we have in Aiding App, and ensure it works

4) I noticed some weird loading behaviour when searching: it shows the search results page temporarily, then switches to the loading state, then switches back to the search results page. When searching, it should never show the search results page before they have been loaded.

nency-canyon:
> 1. Please ensure that the search bar in the navbar works: when you search and then click "enter", it should redirect you to the homepage with that search query selected. The loading bar should be visible when the switch to the homepage is happening, and then the search results loading state should be visible while the results are being fetched.
> 2. Please hide the search bar when the current router page is the homepage / view category page, so that there is not a duplicate search bar on those pages
> 3. Please add the hero search bar to the view category page like we have in Aiding App, and ensure it works
> 4. I noticed some weird loading behaviour when searching: it shows the search results page temporarily, then switches to the loading state, then switches back to the search results page. When searching, it should never show the search results page before they have been loaded.

Addressed these changes and updated PR, Please check and let me know if you find any further issues. Thanks!

danharrin (changes requested):
I pulled the changes and tested it again. The search input in the navbar still does nothing when you search and press enter (view article page). And it is still visible on the homepage / view category pages so there are two search inputs on the page at once.

Additionally, I noticed that when searching and then removing the search from the search bar again, the loading bar appears at the top of the page even though this shouldn't require any data loading, it should just return the user back to the main page content that they were on before they started searching.

nency-canyon:
> I pulled the changes and tested it again. The search input in the navbar still does nothing when you search and press enter (view article page). And it is still visible on the homepage / view category pages so there are two search inputs on the page at once.
> 
> Additionally, I noticed that when searching and then removing the search from the search bar again, the loading bar appears at the top of the page even though this shouldn't require any data loading, it should just return the user back to the main page content that they were on before they started searching.

- Addressed these feedbacks and Raised this fix in the shared common package. 
- Issue was `Header.vue` depended on `useRoute()/useRouter(),` but when loaded from the vendored common package, the router context wasn't available, so both returned undefined.
- Updated `Header.vue` to be router-agnostic by replacing the router dependency with a `hideSearch` prop and a `search` event. The host app now handles the routing.

danharrin (changes requested):
When clicking a category or article from category page search results, the old search should not be pre-populated into the search input in the navbar or the hero please, it should clear when you click a result.

nency-canyon:
> When clicking a category or article from category page search results, the old search should not be pre-populated into the search input in the navbar or the hero please, it should clear when you click a result.

This was happening because navigating between categories reused the same page instance instead of remounting it, so the search state persisted. Keyed the route view by path so it now remounts on category/article changes, clearing the search input in both the navbar and hero automatically. Verified pagination/tab filters still don't reset. Please re-test and confirm.
Thanks

danharrin (approved):

PR #2684, opened 2026-07-13
[ADVAPP-2816]: Introduce support for bulk manage for report assignments

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2816

### Technical Description

> Introduce support for bulk manage for report assignments

> NO

### Cleanup Tasks

> N/A

---


.cleanup-tasks/2026_07_13_bulk_manage_report_assignments_feature.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Bulk Manage Report Assignments Feature
+created: 2026-07-13
+---
+
+## Feature Flags
+
+- App\Features\BulkManageReportAssignmentsFeature
```
Orrison: I don't think that this feature flag, and therefore this entire cleanup task, are needed at all.

I would recommend reading `docs/explanations/feature-flags.md` and `docs/how-tos/manage-feature-flags.md` from Advising App as a refresher on the purpose of feature flags in our systems.

If there is nothing being changed in the system or database that would result in problems in the code you changed till something has happened (like a DB migration) then there is no need for this.

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -150,6 +153,14 @@ public function table(Table $table): Table
                 SelectFilter::make('category')
                     ->options($this->getCategoryOptions()),
             ])
+            ->toolbarActions([
+                // TODO: Cleanup Task - After BulkManageReportAssignmentsFeature is removed, keep the BulkActionGroup and its contents
+                BulkActionGroup::make([
+                    ManageReportAssignmentsBulkAction::make()
+                        ->authorize(fn (): bool => auth()->user()->can('reporting.*.update'))
+                        ->after(fn () => $this->resetTable()),
+                ])->visible(fn (): bool => BulkManageReportAssignmentsFeature::active()),
```
Orrison: As mentioned in my other comment, there really is no need for this feature flag, this `visible` check or the TODO.

The scenario where it would make sense to add this would be if in your branch you were adding a new DB column that this code relied on and an error would be thrown if it was ran before that new column existed.

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -150,6 +153,14 @@ public function table(Table $table): Table
                 SelectFilter::make('category')
                     ->options($this->getCategoryOptions()),
             ])
+            ->toolbarActions([
+                // TODO: Cleanup Task - After BulkManageReportAssignmentsFeature is removed, keep the BulkActionGroup and its contents
+                BulkActionGroup::make([
+                    ManageReportAssignmentsBulkAction::make()
+                        ->authorize(fn (): bool => auth()->user()->can('reporting.*.update'))
```
Orrison: I think we can move this `authorize` into `ManageReportAssignmentsBulkAction` itself, wdyt?

app-modules/authorization/src/Filament/Pages/Reporting.php
```diff
@@ -150,6 +153,14 @@ public function table(Table $table): Table
...
+                        ->after(fn () => $this->resetTable()),
```
Orrison: This `after` as well could probably be pulled into the Action class.

Orrison (changes requested):

.cleanup-tasks/2026_07_13_bulk_manage_report_assignments_feature.md
mittal-canyon: Understood!

app-modules/authorization/src/Filament/Pages/Reporting.php
mittal-canyon: Yes, that will be a cleaner approach; then we just need to call action.

Orrison (dismissed):
Great work, @mittal-canyon, the code is well written and the tests have good coverage and are easy to understand.

Orrison (approved):

PR #2686, opened 2026-07-13
[ADVAPP-2814]: Introduce event registration form versioning

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2814

### Technical Description

> Introduce versioning so that events can be edited even after few attendees have registered for the event.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_07_13_event_versioning_feature.md

---

### DIFF
--- a/.cleanup-tasks/2026_07_13_event_versioning_feature.md
+++ b/.cleanup-tasks/2026_07_13_event_versioning_feature.md
@@ -10,3 +10,5 @@ created: 2026-07-13
 ## Temporary Migrations
 
 ## Additional Cleanup
+
+- Search for `TODO: Cleanup Task - EventVersioningFeature` in the codebase for inline cleanup instructions (migration backfill comments)
--- a/.cleanup-tasks/2026_07_13_event_versioning_feature.md
+++ b/.cleanup-tasks/2026_07_13_event_versioning_feature.md
@@ -17,6 +17,7 @@ created: 2026-07-13
         // TODO: Cleanup Task - EventVersioningFeature - This backfill can be removed once all environments have run this migration
         DB::update('UPDATE event_registration_forms SET root_id = id WHERE root_id IS NULL');
         ```
+    2. Remove the feature flag activate and deactivate from `up()` and `down()` method respectively.
 
 - In `app-modules/meeting-center/src/Models/Event.php`:
     1. Replace the guarded `eventRegistrationForm()` relationship with an unconditional version — keep `->whereNull('archived_at')` but remove the `if (EventVersioningFeature::active())` wrapper:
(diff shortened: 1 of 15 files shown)


.cleanup-tasks/2026_07_13_event_versioning_feature.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Feature Flags
+
+- App\Features\EventVersioningFeature
+
+## Temporary Migrations
+
+## Additional Cleanup
+
+- Search for `TODO: Cleanup Task - EventVersioningFeature` in the codebase for inline cleanup instructions (migration backfill comments)
```
payal-canyon: Could we make this cleanup file a bit more descriptive, similar to our other cleanup files? It would be helpful to include details such as the relevant files and everything that needs to be removed.

You can use `.cleanup-tasks/2026_07_08_subscription_expiration_feature.md` as a reference for the expected structure and level of detail.

app-modules/meeting-center/src/Filament/Resources/Events/Pages/EditEventRegistration.php
```diff
@@ -67,14 +69,19 @@ public function form(Schema $schema): Schema
...
                     $relationship = $component->getRelationship();
                     $record = $component->getCachedExistingRecord();
                     $data = $component->getChildComponentContainer()->getState(shouldCallHooksBefore: false);
 
-                    if ($record) {
-                        $record->fill($data);
-                        $record->save();
+                    if ($record instanceof EventRegistrationForm) {
+                        if (EventVersioningFeature::active()) {
+                            $newVersion = app(CreateEventRegistrationFormVersion::class)->execute($record, $data);
```
payal-canyon: I tested the wizard (multi-step) form, and there's an issue with the versioning flow. When editing and saving a wizard form, the new active version ends up with **0 steps**. The archived version correctly retains its existing steps, but they're never recreated on the new version. It looks like the `Repeater::relationship()` save path isn't carrying the steps across, whereas the non-wizard path works because it rebuilds the fields via the `RichEditor`.

The form module already has a pattern for handling this. In `EditForm::handleRecordUpdate()`, it versions the record and then calls `SaveSubmissibleFieldsFromContent`, which explicitly recreates both the steps and fields on the new version. Since `EventRegistrationForm` is also a `Submissible`, I think we should follow the same approach here (i.e. `handleRecordUpdate()` + reusing or mirroring `SaveSubmissibleFieldsFromContent`) instead of relying on the nested `Fieldset` + `Repeater::relationship()` save flow.

Also, could we add a test that covers the wizard/multi-step path? The current test suite only covers the non-wizard flow, which is why this regression wasn't caught.

app-modules/meeting-center/database/migrations/2026_07_13_115355_add_versioning_to_event_registration_forms_table.php
```diff
@@ -0,0 +1,77 @@
...
+            });
+
+            // TODO: Cleanup Task - EventVersioningFeature - This backfill can be removed once all environments have run this migration
+            DB::update('UPDATE event_registration_forms SET root_id = id WHERE root_id IS NULL');
+
+            Schema::table('event_registration_forms', function (Blueprint $table) {
+                $table->uuid('root_id')->nullable(false)->change();
+                $table->foreign('root_id')->references('id')->on('event_registration_forms');
+                $table->index('root_id');
+            });
```
payal-canyon, line 60: Both `latestVersion()` and the `eventRegistrationForm` relationship assume there's exactly one non-archived form per event, but that isn't currently enforced at the database level. In a concurrent double-save scenario, it's possible to end up with two active rows, and `->whereNull('archived_at')->first()` would then return an arbitrary record.

The form module already guards against this with a partial unique index (see  `add_versioning_to_forms_table.phpapp-modules/form/database/migrations/2026_06_15_143000_add_versioning_to_forms_table.php`), so I think we should follow the same pattern here.

I'd recommend keying the index on `event_id`, as it matches the existing `hasOne` relationship semantics and remains safe with `DuplicateEvent`, since the duplicated form is associated with a new `event_id`.

payal-canyon (changes requested):
`->replicate()` copies both `root_id` and `archived_at`, so the duplicated form ends up in the original form's version tree. As a result, `versions()` and `latestVersion()` can return records across both events.

Could we set both values to `null` before saving the replicated form? That will allow the observer to assign a fresh `root_id` and create a separate version tree for the duplicated form.

app-modules/meeting-center/src/Filament/Resources/Events/Pages/EditEventRegistration.php
```diff
@@ -55,26 +57,63 @@
...
-                ->saveRelationshipsBeforeChildrenUsing(static function (Component | CanEntangleWithSingularRelationships $component): void {
+                ->saveRelationshipsBeforeChildrenUsing(function (Component | CanEntangleWithSingularRelationships $component): void {
                     $relationship = $component->getRelationship();
                     $record = $component->getCachedExistingRecord();
                     $data = $component->getChildComponentContainer()->getState(shouldCallHooksBefore: false);
 
-                    if ($record) {
-                        $record->fill($data);
-                        $record->save();
+                    if ($record instanceof EventRegistrationForm) {
```
payal-canyon, line 83: I tested this on an existing multi-step (wizard) event registration form. If you add a new step and save, the new step isn't persisted, and the form remains at its previous number of steps after saving.

Could you please try to reproduce this issue? We'll need to fix it before moving forward with this PR.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2689, opened 2026-07-14
[ADVAPP-2803]: Introduce the ability to configure the default view for admissions systems

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2803

### Technical Description

> Introduce the ability to configure the default view for admissions systems

> NO

### Cleanup Tasks

> Yes. Feature Flag : ApplicationSubmissionStateDefaultViewFeature

---


.cleanup-tasks/2026_07_14_application_submission_state_default_view_feature.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Feature Flags
+
+- App\Features\ApplicationSubmissionStateDefaultViewFeature
+
+## Temporary Migrations
+
+- app-modules/application/database/migrations/2026_07_14_000002_activate_application_submission_state_default_view_feature.php
+
+## Additional Cleanup
```
payal-canyon, line 14: Could we make this cleanup file a bit more descriptive, similar to our other cleanup files? It would be helpful to include details such as the relevant files and everything that needs to be removed.

You can use `.cleanup-tasks/2026_07_08_subscription_expiration_feature.md` as a reference for the expected structure and level of detail.

app-modules/application/database/migrations/2026_07_14_000002_activate_application_submission_state_default_view_feature.php
```diff
@@ -0,0 +1,50 @@
...
+</COPYRIGHT>
+*/
+
+use App\Features\ApplicationSubmissionStateDefaultViewFeature;
+use Illuminate\Database\Migrations\Migration;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        ApplicationSubmissionStateDefaultViewFeature::activate();
```
payal-canyon: This file doesn't seem necessary, since the feature flag is already being activated in the migration above. Could we remove this file to avoid duplicating the same behavior?

app-modules/application/src/Filament/Resources/Applications/Pages/ManageApplicationSubmissions.php
```diff
@@ -75,14 +76,23 @@ class ManageApplicationSubmissions extends ManageRelatedRecords
...
+            $firstState = ApplicationSubmissionState::query()
+                ->withoutArchivedAndUnused()
+                ->oldest('id')
+                ->first();
+
+            return $firstState ? $firstState->id : 'all';
+        }
+
+        $defaultState = ApplicationSubmissionState::query()
+            ->where('is_default', true)
```
payal-canyon, line 93: `getDefaultActiveTab()` currently selects the default state without verifying that it's actually available as a tab. `getTabs()` excludes states that are archived and have no submissions, but this query doesn't. As a result, if a state is configured as the default and is later archived while still unused, `getDefaultActiveTab()` returns an ID that doesn't exist in the tabs. This leaves the Submissions page with no tab selected and the table unfiltered.

Could we apply `->withoutArchivedAndUnused()` to this query as well? That keeps it consistent with the visibility rules used by `getTabs()`, ensuring the default is only selected when it has a corresponding tab. Otherwise, it will correctly fall back to `'all'`.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2690, opened 2026-07-15
[ADVAPP-2756]: Deprecate Custom Reports

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2756

### Technical Description

Remove Custom Reports feature and table

No

### Cleanup Tasks

N/A

---


app-modules/report/src/Policies/ReportPolicy.php
```diff
@@ -1,162 +0,0 @@
...
-
-class ReportPolicy implements PerformsChecksBeforeAuthorization
-{
-    use PerformsLicenseChecks;
-
-    public function before(Authenticatable $authenticatable): ?Response
-    {
-        if (
-            (! $authenticatable->hasAnyLicense([LicenseType::RetentionCrm, LicenseType::RecruitmentCrm])) &&
-            $authenticatable->cannot('viewAny', User::class)
```
payal-canyon, line 55: Could you also add a migration to remove the orphaned Custom Reports permissions? I checked, and `report.create`, `report.view-any`, `report.*.view`, `report.*.update`, `report.*.delete`, `report.*.restore`, and `report.*.force-delete` are no longer referenced anywhere in the project now that the model, resource, and policy have been removed. They're currently just dangling on existing roles, so we should clean them up.

Could we delete these permissions for both the `web` and `api` guards? The `Report` permission group should remain, though, as `report.access_assistant` and `report.access_assistant_settings` still belong to it.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2691, opened 2026-07-15
[ADVAPP-2811]: Deprecate projects

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2811

### Technical Description

Remove Projects feature, related models, and tables. Remove Pipelines feature, related models, and tables.

No

### Cleanup Tasks

N/A

---


payal-canyon (changes requested):
Could you please resolve the merge conflicts when you get a chance? Thanks!

payal-canyon (changes requested):
I'm so sorry! It looks like there are merge conflicts again.

Could you please resolve them when you get a chance, and then re-request my review? Thanks!

database/migrations/2026_07_17_150155_drop_pipelines_and_related_tables.php
```diff
@@ -0,0 +1,123 @@
...
+                });
+
+            Schema::create('pipelines', function (Blueprint $table) {
+                $table->uuid('id')->primary();
+                $table->string('name');
+                $table->text('description');
+                $table->foreignUuid('user_id')->constrained('users')->cascadeOnDelete();
+                $table->foreignUuid('segment_id')->constrained('segments')->cascadeOnDelete();
+                $table->string('default_stage');
+                $table->unsignedBigInteger('order');
```
payal-canyon: It adds an unsignedBigInteger('order') column (which is NOT NULL) and makes default_stage NOT NULL, but neither matches the current schema for pipelines. It looks like this was copied from pipeline_stages, where the order column actually exists.

Could we remove these changes from the rollback? Otherwise, rolling back would recreate a schema that doesn't match the original table.

payal-canyon (changes requested):
1. We should also drop the `prospect_pipeline.is_enabled` setting created in `database/settings/2024_10_23_154606_prospect_pipeline_settings.php`, since its `ProspectPipelineSettings` class is being deleted.

2. Since the `ProjectReport` case is being removed from `ReportAccessKey`, we should also delete the `report_key = 'project-report'` rows from `report_user_accesses` and `report_team_accesses` in the drop migration, otherwise they stay orphaned in every tenant with no matching enum case to ever surface or clean them up.

3. We should also delete the audits rows for the removed morph aliases ( `project`, `project_file`, `project_milestone`, `project_milestone_status`, `pipeline`, `pipeline_stage`, `educatable_pipeline_stage`) in these drop migrations - once the models are gone their aliases leave the morph map, and `ViewAudit` throws Class "project_milestone_status" not found when `getModified()` tries to resolve the `auditable` relation. Every tenant has 10 `project_milestone_status` audit rows from `ProjectMilestoneStatusSeeder`, so this breaks for all tenants, not just ones that used Projects.

4. Could we also remove the corresponding PHPStan baseline entries? Each of these has its own `$ignoreErrors[]` block in `phpstan-baseline.php`, so they should be cleaned up as part of this change:
* `app-modules/pipeline/src/Models/Pipeline.php`
* `app-modules/project/src/Models/Project.php`
* `app-modules/project/src/Models/ProjectFile.php`
* `app-modules/project/src/Models/ProjectMilestone.php`
* `app-modules/project/src/Models/ProjectMilestoneStatus.php`
* `app-modules/report/src/Filament/Widgets/ProjectManagementTable.php`
* `app/Livewire/ProspectPipelineKanban.php`

payal-canyon (approved):

PR #2698, opened 2026-07-16
[ADVAPP-2813]: Update forms, events, and admissions behavior when submissions have occurred

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2813

### Technical Description

> Update forms, events, and admissions behavior when submissions have occurred to have archive functionality instead of delete

> No

### Cleanup Tasks

> .cleanup-tasks/2026_07_16_event_archiving_feature.md

---


danharrin (changes requested):
The common package now has new utilities for this, as discussed privately

app-modules/application/src/Filament/Resources/Applications/Pages/ListApplications.php
```diff
@@ -112,8 +112,7 @@ public function table(Table $table): Table
             ])
             ->toolbarActions([
                 BulkActionGroup::make([
-                    DeleteBulkAction::make()
-                        ->authorizeIndividualRecords('delete'),
+                    ArchiveBulkAction::make()->authorizeIndividualRecords('delete'),
```
danharrin: For any `delete` policy methods that are not any different to the `deleteAny`, no need to authorise individual records. This is slower and doesn't provide safety since no auth conditions per-record change.

app-modules/application/src/Models/Application.php
```diff
@@ -187,4 +190,27 @@ public function latestVersion(): ?Application
...
+            'versions',
+            fn (Builder $query) => $query
+                ->withoutGlobalScopes()
+                ->whereHas('submissions'),
+        );
+    }
+
+    public function isUsed(): bool
+    {
+        return (bool) ($this->hasSubmissions ??= ApplicationSubmission::query()
```
danharrin: As per the common docs, name this consistently with the Laravel `withExists()` convention so that eager loading with `withExists()` allows this check to be fast. so `submissions_exists` I think. And you don't need to define the property on the class as it is a model attribute like any other.

app-modules/meeting-center/src/Filament/Resources/Events/Pages/ViewEvent.php
```diff
@@ -91,7 +90,6 @@ protected function getHeaderActions(): array
                 ->url(fn (Event $record): string => route('event-registration.show', ['event' => $record]))
                 ->icon('heroicon-m-arrow-top-right-on-square')
                 ->openUrlInNewTab(),
-            DeleteAction::make(),
```
danharrin, line 94: For each of these removed `DeleteAction`, should we put an `ArchiveAction` instead?

danharrin (changes requested):

danharrin (dismissed):
As discussed lets remove archived records from the lists

app-modules/application/src/Models/Application.php
```diff
@@ -187,4 +188,37 @@ public function latestVersion(): ?Application
...
+     */
+    public function used(Builder $query): void
+    {
+        $query->whereHas(
+            'versions',
+            fn (Builder $query) => $query
+                ->withoutGlobalScopes()
+                ->whereHas('submissions'),
+        );
+    }
```
Orrison, line 203: Hmm, here and for the other Form derivatives, technically don't we need to look at whether or not THIS entity has submissions? Not just it's versions?

It's possible that this could be the only instance, right? The root? If it was never edited?

@danharrin

Orrison (changes requested):

app-modules/application/src/Models/Application.php
danharrin: I think it's already covered, versions relation is defined as `hasMany(Application::class, 'root_id', 'root_id')`, so it keys on `root_id` and includes the root record itself, meaning a never-edited root's own submissions are already counted (same family logic as `isUsed()`).

Orrison (approved):

PR #2706, opened 2026-07-17
[ADVAPP-2812]: Deprecate cases

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2812

### Technical Description

Remove Case feature, related models, and tables.

No

### Cleanup Tasks

N/A

---


Orrison (changes requested):
This is looking great but I believe we forgot to mention in the ticket so it was missed here that we need to remove the Case related widgets as well. `widgets/case-form` and `widgets/case-feedback-form`, and making sure we remove them from the widget build scripts and stuff like that.

`resources/views/filament/infolists/components/submission-entry.blade.php` is looking like it is still referencing `caseFormSubmission`, we should look how / where that is used and if we need to fix or remove it.

I think we still have some permissions (and their Groups) left over:
- case_update
- case_assignment

`CaseAssignment` and `CaseUpdate` are also still mentioned in `app-modules/authorization/src/Enums/PermissionDescription.php`

`resources/lang/en/cruds.php` still seems to mention some case related stuff.

There may still be some `interactions` (and potentially subscriptions/timeline/notifications) where a case is a morph type

Orrison (approved):

PR #2716, opened 2026-07-22
[ADVAPP-2825]: Interaction related Selects on Campaigns are duplicating options

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2825

### Technical Description

- Add interaction options filtering for campaign population model in InteractionBlock

No

### Cleanup Tasks

No

---


Orrison (dismissed):
I think we should also test that this is working with in the create flow.

It also made me realize that we probably should add an after state updated on the population group field to tell it to wipe these fields if changed. That way we wouldn't have anything weird happen if they went back in steps and changed the population group after already making selections and they switched it to a group for a different model type.

payal-canyon (approved):

PR #2730, opened 2026-07-28
[ADVAPP-2829]: Some users have reported that the advanced filtering in reports is not working as expected

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2829

### Technical Description

- Add reactive alert stats update on population filter change

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/alert/src/Filament/Widgets/AlertStats.php
+++ b/app-modules/alert/src/Filament/Widgets/AlertStats.php
@@ -43,7 +43,6 @@ use Filament\Widgets\StatsOverviewWidget;
 use Filament\Widgets\StatsOverviewWidget\Stat;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Number;
-use Livewire\Attributes\Computed;
 
 /**
  * @property-read Stat[] $stats
@@ -55,27 +54,7 @@ class AlertStats extends StatsOverviewWidget
     /**
      * @return Stat[]
      */
-    #[Computed]
-    public function stats(): array
-    {
-        return $this->calculateStats();
-    }
-
-    /**
-     * @return Stat[]
-     */
     public function getStats(): array
-    {
-        return $this->calculateStats();
-    }
-
-    /**
-     * Calculate alert statistics for the current filter state.
-     * This method is testable and can be called directly in tests.
-     *
-     * @return Stat[]
-     */
-    protected function calculateStats(): array
     {
         $groupId = $this->getSelectedGroup();
 
--- a/app-modules/alert/src/Filament/Widgets/AlertStats.php
+++ b/app-modules/alert/src/Filament/Widgets/AlertStats.php
@@ -43,12 +43,39 @@ use Filament\Widgets\StatsOverviewWidget;
 use Filament\Widgets\StatsOverviewWidget\Stat;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Number;
+use Livewire\Attributes\Computed;
 
+/**
+ * @property-read Stat[] $stats
+ */
 class AlertStats extends StatsOverviewWidget
 {
     use InteractsWithPageFilters;
 
+    /**
+     * @return Stat[]
+     */
+    #[Computed]
+    public function stats(): array
+    {
+        return $this->calculateStats();
+    }
+
+    /**
+     * @return Stat[]
+     */
     public function getStats(): array
+    {
+        return $this->calculateStats();
+    }
+
+    /**
+     * Calculate alert statistics for the current filter state.
+     * This method is testable and can be called directly in tests.
+     *
+     * @return Stat[]
+     */
+    protected function calculateStats(): array
     {
         $groupId = $this->getSelectedGroup();
 
--- a/app-modules/alert/src/Filament/Widgets/AlertStats.php
+++ b/app-modules/alert/src/Filament/Widgets/AlertStats.php
@@ -48,7 +48,7 @@ class AlertStats extends StatsOverviewWidget
 {
     use InteractsWithPageFilters;
 
-  protected ?string $pollingInterval = null;
+    protected ?string $pollingInterval = null;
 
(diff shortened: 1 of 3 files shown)


app-modules/alert/src/Filament/Widgets/AlertStats.php
```diff
@@ -43,12 +43,39 @@
...
 class AlertStats extends StatsOverviewWidget
 {
     use InteractsWithPageFilters;
 
+    /**
+     * @return Stat[]
+     */
+    #[Computed]
+    public function stats(): array
+    {
```
payal-canyon: This method is never called - Filament renders stats via `getCachedStats()` -> `getStats()`, and nothing reads `$this->stats`. Please remove it.

app-modules/alert/src/Filament/Widgets/AlertStats.php
```diff
@@ -43,12 +43,39 @@
...
+        return $this->calculateStats();
+    }
+
+    /**
+     * Calculate alert statistics for the current filter state.
+     * This method is testable and can be called directly in tests.
+     *
+     * @return Stat[]
+     */
+    protected function calculateStats(): array
```
payal-canyon: The docblock says this is callable directly in tests, but it's `protected` and no test calls it. Please fold the body back into getStats().

app-modules/alert/tests/Tenant/Filament/Widgets/AlertStatsTest.php
```diff
@@ -442,3 +442,87 @@
         ->and($stats[0]->getLabel())->toBe($alertConfig->preset->getInsightsPaneTitle())
         ->and($stats[0]->getValue())->toBe((string) $inGroupCount);
 });
+
+it('updates alert stats reactively when population filter changes', function () {
```
payal-canyon: The name says `reactively`, but the test just assigns `pageFilters` on a bare new `AlertStats()` and calls `getStats()` - no Livewire, no page, no reactivity. Please rename it to match what it asserts, or drive it through the page with `livewire(RetentionCrmDashboard::class)->callAction(...)`.

app-modules/alert/tests/Tenant/Filament/Widgets/AlertStatsTest.php
```diff
@@ -442,3 +442,87 @@
...
+    Student::factory()
+        ->count($secondGroupCount)
+        ->state(['last' => 'Jones', 'firstgen' => true])
+        ->create();
+
+    app(GenerateStudentAlertsView::class)->execute();
+
+    $widget = new AlertStats();
+
+    $widget->pageFilters = ['population' => ['type' => 'saved', 'groupId' => $firstGroup->getKey()]];
```
payal-canyon: This largely duplicates `it('filters alerts by population group')` above. The only new coverage is the `population => ['type' => 'saved', ...]` shape, so could we fold that into the existing test instead of adding a third?

payal-canyon (changes requested):

payal-canyon (changes requested):
Also, could you please test a few different scenarios? For example, after applying a filter and then clearing it, the count should return to the original value. At the moment, that doesn't seem to be happening.

app-modules/alert/src/Filament/Widgets/AlertStats.php
```diff
@@ -44,10 +44,16 @@
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Number;
 
+/**
+ * @property-read Stat[] $stats
+ */
 class AlertStats extends StatsOverviewWidget
 {
     use InteractsWithPageFilters;
```
payal-canyon, line 49: Please add `protected ?string $pollingInterval = null;` - the inherited 5-second poll keeps a request in flight, so the widget misses the updated pageFilters and shows a stale saved group.

payal-canyon (changes requested):
In `app-modules/prospect/src/Filament/Widgets/ProspectStats.php`:
Please add `protected ?string $pollingInterval = null;` here too - the Prospects Action Center has the same Advanced Filtering section and the same stale-cards bug for the same reason.

app-modules/alert/src/Filament/Widgets/AlertStats.php
```diff
@@ -44,10 +44,18 @@
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Number;
 
+/**
+ * @property-read Stat[] $stats
+ */
```
payal-canyon: This docblock appears to be left over from the removed computed method. Since there's no corresponding property or method on the class anymore, could you please remove it?

app-modules/alert/src/Filament/Widgets/AlertStats.php
```diff
@@ -44,10 +44,18 @@
...
 class AlertStats extends StatsOverviewWidget
 {
     use InteractsWithPageFilters;
 
+    protected ?string $pollingInterval = null;
+
+    /**
+     * @return Stat[]
+     */
```
payal-canyon: Could you please remove this one as well?

app-modules/alert/tests/Tenant/Filament/Widgets/AlertStatsTest.php
```diff
@@ -434,11 +434,38 @@
 
     app(GenerateStudentAlertsView::class)->execute();
 
+    $totalCount = $inGroupCount + $outGroupCount;
+
+    // Test with no filter - should show all students with alerts
     $widget = new AlertStats();
+    $widget->pageFilters = [];
+    $unfileredStats = $widget->getStats();
```
payal-canyon: There's a typo here: `unfileredStats` should be `unfilteredStats`. Could you please fix it?

payal-canyon (changes requested):

payal-canyon (approved):

PR #2741, opened 2026-08-03
[ADVAPP-2847]: Introduce support for export and import for customer and employee chatbot categories

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2847

### Technical Description

>  Implemented CustomerAdvisorCategoryExporter and EmployeeAdvisorCategoryExporter for exporting categories.
> Created CustomerAdvisorCategoryImporter and EmployeeAdvisorCategoryImporter for importing categories.
> Enhanced ManageEmployeeAdvisorCategories and ManageCategories pages to include import and export actions.
> Added tests for import/export functionality and validation rules for unique category names.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_08_05_rename_customer_and_employee_categories_duplicate_name.md

---


app-modules/ai/database/migrations/2026_08_03_171930_convert_advisor_category_names_to_citext.php
```diff
@@ -0,0 +1,147 @@
...
+*/
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
```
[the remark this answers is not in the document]
jayushi-canyon: There is already a migration that enables the citext extension

payal-canyon (changes requested):
It looks like the lint checks have failed. Could you please resolve them and then re-request my review? Thanks!

app-modules/ai/src/Filament/Imports/EmployeeAdvisorCategoryImporter.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $category = new EmployeeAdvisorCategory();
+        $category->employeeAdvisor()->associate($this->getEmployeeAdvisorId());
+
+        return $category;
+    }
+
+    public static function getCompletedNotificationBody(Import $import): string
+    {
+        $body = 'Your employee chatbot categories import has completed and ' . number_format($import->successful_rows) . ' ' . str('row')->plural($import->successful_rows) . ' imported.';
```
payal-canyon: The completion notification currently hardcodes the wording (`"Your employee chatbot categories import has completed..."`), while the corresponding exporters use `static::EXPORT_NAME`. Since `IMPORT_NAME` now exists on these classes, the two could drift if the label is ever changed.

Could you please use `static::IMPORT_NAME` in `getCompletedNotificationBody()` as well? That would keep it consistent with the exporter convention and ensure the label stays in sync with the Import/Export.

app-modules/ai/src/Filament/Imports/CustomerAdvisorCategoryImporter.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        $category = new CustomerAdvisorCategory();
+        $category->customerAdvisor()->associate($this->getCustomerAdvisorId());
+
+        return $category;
+    }
+
+    public static function getCompletedNotificationBody(Import $import): string
+    {
+        $body = 'Your customer chatbot categories import has completed and ' . number_format($import->successful_rows) . ' ' . str('row')->plural($import->successful_rows) . ' imported.';
```
payal-canyon: The completion notification currently hardcodes the wording (`"Your customer chatbot categories import has completed..."`), while the corresponding exporters use `static::EXPORT_NAME`. Since `IMPORT_NAME` now exists on these classes, the two could drift if the label is ever changed.

Could you please use `static::IMPORT_NAME` in `getCompletedNotificationBody()` as well? That would keep it consistent with the exporter convention and ensure the label stays in sync with the Import/Export.

app-modules/ai/tests/Tenant/Filament/Resources/CustomerAdvisors/Pages/ManageCategoriesTest.php
```diff
@@ -110,6 +112,25 @@
     );
 });
 
+test('shows import and export actions for customer advisor categories', function () {
```
payal-canyon: The new tests verify that the Import and Export actions are visible for a user with the required permissions, but they don't cover the negative case.

Could you please add tests to verify that a user without `customer_advisor.create` / `assistant_custom.create` cannot see the Import action, and that a user without the appropriate view permission cannot see the Export action?

It would also be great to add a `CustomerCategoryUniqueTest`, similar to the existing `EmployeeCategoryUniqueTest`, so both tables have database-level coverage for the case-insensitive uniqueness constraint.

app-modules/ai/database/migrations/2026_08_05_175003_tmp_rename_customer_and_employee_categories_duplicate_name.php
```diff
@@ -0,0 +1,138 @@
...
+
+    public function down(): void {}
+
+    private function renameCaseInsensitiveDuplicates(string $table, string $ownerColumn): void
+    {
+        $records = DB::table($table)
+            ->select(['id', $ownerColumn, 'name'])
+            ->orderBy($ownerColumn)
+            ->orderByRaw('LOWER(name)')
+            ->orderBy('id')
```
payal-canyon: This orders by `(owner, LOWER(name), id)` and keeps the first row it encounters. Including soft-deleted rows is correct, since our unique index is a plain unique([advisor_id, name]), but using id as the tie-breaker doesn't prefer live records.

Could you please add `deleted_at IS NULL DESC` ahead of `id` in the ordering so live rows retain the original name and only soft-deleted duplicates receive the suffix?

Also, we already have a shared `FixesDuplicateNames` trait that could be reused here instead of maintaining another custom implementation. If you do go that route, I think a few overrides would still be needed since our schema differs from Aiding App's:

`usesSoftDeletes` should be `false`, since our unique index isn't partial.
`orderDuplicateRecords()` should use the live-rows-first ordering above.
`buildDeduplicatedValue()` should continue truncating the base name before appending the suffix to avoid exceeding the `varchar(255)` limit.
I'd leave `down()` as-is rather than calling `revertDuplicates()`, since the ticket doesn't require duplicate renames to be reversed.

Could you please check whether this approach makes sense, and if so, implement and test it? I'm assuming this should work, but I'd appreciate it if you could cross-check it as well.

payal-canyon (changes requested):

app-modules/ai/database/migrations/2026_08_05_175003_convert_customer_advisor_categories_name_to_citext.php
```diff
@@ -0,0 +1,56 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            DB::statement('ALTER TABLE customer_advisor_categories ALTER COLUMN name TYPE citext');
+            DB::statement('ALTER TABLE employee_advisor_categories ALTER COLUMN name TYPE citext');
+        });
```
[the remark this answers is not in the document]
jayushi-canyon: citext extension migration already exists

payal-canyon (approved):

PR #2744, opened 2026-08-04
[ADVAPP-2867]: Enhance the campaign list view by providing details about the population of students represented

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2867

### Technical Description

- Enhance the campaign list view by providing details about the population of students represented

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
+++ b/app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
@@ -73,7 +73,6 @@ class ListCampaigns extends ListRecords
                         static $populationCountsByGroupId = [];
 
                         $groupId = (string) $group->getKey();
-                        
                         $populationCount = $populationCountsByGroupId[$groupId] ??= ($group->type === \AdvisingApp\Group\Enums\GroupType::Static)
                              ? $group->model->query()->whereIn(
                                  $group->model->instance()->getQualifiedKeyName(),
--- a/app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
+++ b/app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
@@ -38,6 +38,7 @@ namespace AdvisingApp\Campaign\Filament\Resources\Campaigns\Pages;
 
 use AdvisingApp\Campaign\Filament\Resources\Campaigns\CampaignResource;
 use AdvisingApp\Campaign\Models\Campaign;
+use AdvisingApp\Group\Actions\TranslateGroupFilters;
 use App\Filament\Tables\Columns\IdColumn;
 use App\Models\User;
 use Filament\Actions\CreateAction;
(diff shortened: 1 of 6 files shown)


app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
```diff
@@ -60,9 +61,18 @@ public function table(Table $table): Table
...
+                    ->description(function (Campaign $record): ?string {
+                        $group = $record->group;
+
+                        if ($group === null) {
+                            return null;
+                        }
+
+                        $population = Str::of($group->model->getPluralLabel())->ucfirst();
+
+                        return "{$group->name} (" . number_format($group->subjects_count) . " {$population})";
```
payal-canyon: `subjects_count` only counts `segment_subjects` rows, which exist exclusively for static groups. Dynamic groups store their population as `QueryBuilder` filters and don't have any `subjects` rows, so they always render as `(0 Students)`. For example, a dynamic group with a real population of 3 students would display as `Everyone (0 Students)`, and since `GroupType::default()` is `Dynamic`, this is the common case.

It also disagrees with the **Population Count** column on the Groups list page, which resolves the count via `TranslateGroupFilters`. Could you please resolve the count through `TranslateGroupFilters` here as well, caching it on the record the same way `ListGroups` does? That would also avoid the `number_format(null)` deprecation risk when `subjects_count` isn't loaded.

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/ListCampaigns.php
```diff
@@ -86,7 +96,8 @@ public function table(Table $table): Table
             // @phpstan-ignore argument.templateType
             ->modifyQueryUsing(function (Builder $query) {
                 /** @var Builder<Campaign> $query */
-                $query->withoutArchived();
+                $query->withoutArchived()
+                    ->with(['group' => fn ($query) => $query->withCount('subjects')]);
```
payal-canyon: Once the count comes from `TranslateGroupFilters`, `withCount('subjects')` is dead weight - could you please drop it and keep just `->with('group')`?

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/ListCampaignsTest.php
```diff
@@ -194,3 +198,28 @@
...
+
+it('shows the group name and population as the name column description', function () {
+    asSuperAdmin();
+
+    $students = Student::factory()->count(3)->create();
+
+    $group = Group::factory()->create([
+        'name' => 'Everyone',
+        'model' => GroupModel::Student,
+        'type' => GroupType::Static,
```
payal-canyon: This test only covers a static group, which happens to be the one case the current implementation gets right - could you please add coverage for a dynamic group as well, since that's the default type? It would also be good to exercise both `GroupModel::Student` and `GroupModel::Prospect` so the pluralised label is verified for each.

payal-canyon (changes requested):

payal-canyon (changes requested):
Hey @monali4-canyon ,
Could you please review the Copilot suggestions? If you agree with them, could you please give them a thumbs up and resolve them? If you don't agree, could you please leave a short explanation as to why? Once that's done, could you please re-request my review? Thanks!

payal-canyon (approved):

PR #2767, opened 2026-08-14
[ADVAPP-2889]: Refactor the way automated chat naming works

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2889

### Technical Description

> When a chat is started with no context initilally the chat title will look like this 'New Chat 8/13/26 @ 7:10 AM'.
> After three prompts and responses name will be updated based on the prompts context by AI.
> If user has already edited the chat name then AI won't update name again.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_08_14_ai_thread_auto_naming_feature.md

---

### DIFF
--- a/app-modules/ai/src/Actions/CreateThread.php
+++ b/app-modules/ai/src/Actions/CreateThread.php
@@ -42,6 +42,7 @@ use AdvisingApp\Ai\Models\AiThread;
 use AdvisingApp\Ai\Settings\AiSettings;
 use App\Features\AiThreadAutoNamingFeature;
 use App\Models\Tenant;
+use App\Settings\DisplaySettings;
 
 class CreateThread
 {
@@ -80,7 +81,7 @@ class CreateThread
 
     protected function getDefaultThreadName(): string
     {
-        return 'New Chat ' . now()->format('n/j/y @ g:i A');
+        return 'New Chat ' . now()->setTimezone(app(DisplaySettings::class)->getTimezone())->format('n/j/y @ g:i A');
     }
 
     protected function getDefaultAiAssistant(AiAssistantApplication $application): AiAssistant
--- a/app-modules/ai/src/Actions/CreateThread.php
+++ b/app-modules/ai/src/Actions/CreateThread.php
@@ -40,6 +40,7 @@ use AdvisingApp\Ai\Enums\AiAssistantApplication;
 use AdvisingApp\Ai\Models\AiAssistant;
 use AdvisingApp\Ai\Models\AiThread;
 use AdvisingApp\Ai\Settings\AiSettings;
+use App\Features\AiThreadAutoNamingFeature;
 use App\Models\Tenant;
 
 class CreateThread
@@ -48,11 +49,17 @@ class CreateThread
     {
         $assistant ??= $this->getDefaultAiAssistant($application);
 
-        $existingThread = auth()->user()->aiThreads()
-            ->whereNull('name')
+        $existingThreadQuery = auth()->user()->aiThreads()
             ->whereBelongsTo($assistant, 'assistant')
-            ->whereDoesntHave('messages')
-            ->first();
+            ->whereDoesntHave('messages');
+
+        if (AiThreadAutoNamingFeature::active()) {
+            $existingThreadQuery->whereNull('named_by_user_at');
+        } else {
+            $existingThreadQuery->whereNull('name');
+        }
+
+        $existingThread = $existingThreadQuery->first();
 
         if ($existingThread) {
             return $existingThread;
@@ -61,9 +68,19 @@ class CreateThread
         $thread = new AiThread();
         $thread->assistant()->associate($assistant);
         $thread->user()->associate(auth()->user());
+
+        if (AiThreadAutoNamingFeature::active()) {
+            $thread->name = $this->getDefaultThreadName();
+        }
+
         $thread->save();
 
         return $thread;
+    }
+
+    protected function getDefaultThreadName(): string
+    {
+        return 'New Chat ' . now()->format('n/j/y @ g:i A');
     }
 
     protected function getDefaultAiAssistant(AiAssistantApplication $application): AiAssistant
--- a/app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
+++ b/app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
@@ -36,6 +36,7 @@
 
 namespace AdvisingApp\Ai\Jobs\Advisors;
 
+use AdvisingApp\Ai\Events\Advisors\AdvisorThreadRenamed;
 use AdvisingApp\Ai\Models\AiMessage;
 use AdvisingApp\Ai\Models\AiThread;
 use AdvisingApp\Report\Enums\TrackedEventType;
@@ -101,6 +102,8 @@ class GenerateAiThreadName implements ShouldQueue
         $this->thread->name = trim($name);
         $this->thread->saved_at = now();
         $this->thread->save();
+
+        event(new AdvisorThreadRenamed($this->thread));
 
         dispatch(new RecordTrackedEvent(
             type: TrackedEventType::AiThreadSaved,
--- a/app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
+++ b/app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Advising App® is licensed under the Elastic License 2.0. For more details,
+    see https://github.com/canyongbs/advisingapp/blob/main/LICENSE.
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Advising App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    https://www.canyongbs.com or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AdvisingApp\Ai\Jobs\Advisors;
+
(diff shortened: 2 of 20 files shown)


app-modules/integration-open-ai/src/Services/BaseOpenAiService.php
```diff
@@ -887,25 +888,27 @@ protected function sendNewMessage(AiMessage $message, array $files, bool $hasIma
...
+            if (! AiThreadAutoNamingFeature::active()) {
+                try {
+                    if (is_null($message->thread->name)) {
+                        $prompt = $message->context . "\nThe following is the start of a chat between you and a user:\n" . $message->content;
 
-                    $message->thread->name = $this->complete($prompt, 'Generate a title for this chat, in 5 words or less. Do not respond with any greetings or salutations, and do not include any additional information or context. Just respond with the title:');
+                        $message->thread->name = $this->complete($prompt, 'Generate a title for this chat, in 5 words or less. Do not respond with any greetings or salutations, and do not include any additional information or context. Just respond with the title:');
 
-                    $message->thread->saved_at = now();
+                        $message->thread->saved_at = now();
```
payal-canyon: Could we set `saved_at` when the first user message is persisted, rather than only in `GenerateAiThreadName`?

With the new flow, `saved_at` stays `null` for threads with fewer than three messages, so `DeleteUnsavedAiThreads` will delete those threads after three days. This could result in short conversations being deleted unexpectedly.

Something like guarding `saved_at` with `blank()` when the first message is saved should preserve the existing behaviour.

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
```diff
@@ -0,0 +1,113 @@
...
+
+            return;
+        }
+
+        if (filled($this->thread->fresh()?->named_by_user_at)) {
+            return;
+        }
+
+        $this->thread->name = trim($name);
+        $this->thread->saved_at = now();
```
payal-canyon: same as above.

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
```diff
@@ -0,0 +1,113 @@
...
+        $this->thread->save();
+
+        event(new AdvisorThreadRenamed($this->thread));
+
+        dispatch(new RecordTrackedEvent(
+            type: TrackedEventType::AiThreadSaved,
+            occurredAt: now(),
```
payal-canyon: Could we also move the `TrackedEventType::AiThreadSaved` dispatch to wherever `saved_at` ends up being set as part of the fix above - i.e. on the first exchange rather than after three?

Saved Conversations in `AiStats` reads the cumulative `tracked_event_counts` row without a date window, so the number won’t visibly drop, but threads that never reach three exchanges will no longer be counted.

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
payal-canyon: This is now the only place where `saved_at` is set for new threads. Since `DeleteUnsavedAiThreads` deletes threads where `saved_at IS NULL` after 3 days, threads with only one or two exchanges could now be soft-deleted.

Could we set `saved_at` when the first user message is persisted in `SendAdvisorMessage`, guarded by `blank($this->thread->saved_at)`, and leave `GenerateAiThreadName` responsible only for the name?

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        $this->thread->name = trim($name);
+        $this->thread->saved_at = now();
+        $this->thread->save();
+
+        event(new AdvisorThreadRenamed($this->thread));
+
+        dispatch(new RecordTrackedEvent(
+            type: TrackedEventType::AiThreadSaved,
```
payal-canyon: `Saved Conversations` uses a cumulative counter, so the displayed number won’t immediately drop. However, since the event now fires only after three exchanges, most short threads will no longer be counted.

Could we move the `TrackedEventType::AiThreadSaved` dispatch to wherever `saved_at` is set in the fix above, so it continues to be recorded on the first exchange?

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
```diff
@@ -0,0 +1,113 @@
...
+use Throwable;
+
+class GenerateAiThreadName implements ShouldQueue
+{
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    public int $tries = 3;
```
payal-canyon: Could we set `$tries = 1` and wrap the `event()` call in a `try/catch` that reports the broadcast failure?

The AI failure is already caught, so the retry is mainly useful when `AdvisorThreadRenamed` fails after the name has been saved. In that case, the job retries and makes another billed AI call and records the event again.

Also, since `retry_after` is 90s and the job has no `$timeout`, a slow AI call could potentially be re-reserved and run concurrently. Setting `$tries = 1` would avoid both cases.

app-modules/ai/src/Jobs/Advisors/SendAdvisorMessage.php
```diff
@@ -249,6 +250,14 @@ public function handle(): void
                 'updated_at' => $message->created_at,
             ]);
         });
+
+        if (
+            AiThreadAutoNamingFeature::active() &&
+            blank($this->thread->named_by_user_at) &&
+            ($this->thread->messages()->whereNotNull('user_id')->count() === 3)
```
payal-canyon, line 269: Could we add the same naming dispatch to `RetryAdvisorMessage` after `$response->save()` / `$this->thread->touch()`?

If the third send fails after the user message is saved, the count reaches 3 but `SendAdvisorMessage` never reaches the naming check. `RetryAdvisorMessage` then completes the exchange without triggering it, and the next message takes the count to 4, so `=== 3` never matches and the thread is never named.

app-modules/ai/src/Jobs/Advisors/GenerateAiThreadName.php
```diff
@@ -0,0 +1,113 @@
...
+    public function handle(): void
+    {
+        if (! AiThreadAutoNamingFeature::active()) {
+            return;
+        }
+
+        // The thread may have been renamed by the user between when this job was
+        // dispatched and when it runs, in which case it must not be overwritten.
+        if (filled($this->thread->fresh()?->named_by_user_at)) {
+            return;
```
payal-canyon, line 75: Could we guard against `fresh()` returning `null` here?

If the user deletes the thread between the third exchange and the job running, `fresh()` returns `null` because of the `SoftDeletingScope`. `filled(null)` is `false`, so the job would still make the AI call and attempt to update the soft-deleted thread.

Something like this should cover both cases:
```php
$thread = $this->thread->fresh();

if (! $thread || $thread->trashed() || filled($thread->named_by_user_at)) {
    return;
}
```

app-modules/ai/src/Actions/CreateThread.php
```diff
@@ -61,11 +68,21 @@ public function __invoke(AiAssistantApplication $application, ?AiAssistant $assi
...
+        }
+
         $thread->save();
 
         return $thread;
     }
 
+    protected function getDefaultThreadName(): string
+    {
+        return 'New Chat ' . now()->format('n/j/y @ g:i A');
```
payal-canyon: Could we use the authenticated user's timezone here instead of the app's UTC timezone?

`CreateThread` runs with an authenticated user, and `DisplaySettings::getTimezone()` already resolves the user's timezone. Currently, `now()->format(...)` uses UTC, so users in other timezones could see the thread timestamp several hours off.

Something like:

```php
now()->setTimezone(app(DisplaySettings::class)->getTimezone())->format('n/j/y @ g:i A')
```

The test expectation could be converted to the same timezone as well.

app-modules/ai/src/Actions/CreateThread.php
```diff
@@ -48,11 +49,17 @@ public function __invoke(AiAssistantApplication $application, ?AiAssistant $assi
...
-        $existingThread = auth()->user()->aiThreads()
-            ->whereNull('name')
+        $existingThreadQuery = auth()->user()->aiThreads()
             ->whereBelongsTo($assistant, 'assistant')
-            ->whereDoesntHave('messages')
-            ->first();
+            ->whereDoesntHave('messages');
+
+        if (AiThreadAutoNamingFeature::active()) {
+            $existingThreadQuery->whereNull('named_by_user_at');
```
payal-canyon, line 58: Could we assert this in the existing `it('can rename a thread')` test in `AIAssistantPageTest`?

Currently, no test verifies that `named_by_user_at` is actually written. Since this is what prevents `GenerateAiThreadName` from overwriting a user's chosen name, a simple assertion like this would cover it:

`expect($thread->refresh()->named_by_user_at)->not->toBeNull();`

app-modules/ai/src/Jobs/Advisors/SendAdvisorMessage.php
```diff
@@ -249,6 +250,14 @@ public function handle(): void
...
+        ) {
+            GenerateAiThreadName::dispatch($this->thread);
```
payal-canyon: Could we align this with the existing style in this file and use `dispatch(new ...)` instead?

app-modules/ai/tests/Tenant/Jobs/Advisors/GenerateAiThreadNameTest.php
```diff
@@ -0,0 +1,191 @@
...
+
+        public function complete(string $prompt, string $content, bool $shouldTrack = true): string
+        {
+            return ($this->complete)();
+        }
+    });
+}
+
+it('reports the exception and leaves the thread unnamed when the AI service fails', function () {
+    asSuperAdmin();
```
payal-canyon, line 207: Could we add an assertion that the exception is actually reported? Since the test is named `reports the exception and leaves the thread unnamed`, `Exceptions::fake()` with `Exceptions::assertReported(Exception::class)` would make sure we're testing the reported behaviour as well.

app-modules/ai/src/Actions/CreateThread.php
```diff
@@ -48,11 +49,17 @@ public function __invoke(AiAssistantApplication $application, ?AiAssistant $assi
...
+        } else {
+            $existingThreadQuery->whereNull('name');
```
payal-canyon, line 60: Could we also exclude `named_by_user_at` in `CloneAiThread.php` when cloning the thread?

Currently, `replicate(except: [...])` excludes `saved_at` but not `named_by_user_at`, so the cloned thread inherits the original thread's “named by user” state.

payal-canyon (changes requested):

payal-canyon (changes requested):
Hey @jayushi-canyon,
Could you please also review the Copilot suggestions? If you agree with a suggestion, please fix it; otherwise, please reply with the reason. Thanks!

app-modules/ai/src/Jobs/Advisors/SendAdvisorMessage.php
```diff
@@ -123,6 +126,16 @@ public function handle(): void
 
         $message->save();
 
+        if (blank($this->thread->saved_at)) {
```
payal-canyon: Could we add `AiThreadAutoNamingFeature::active()` to this condition? While the flag is inactive, `CreateThread` doesn't set a default `name`, so the legacy block in `BaseOpenAiService::sendNewMessage()` still fires and dispatches `AiThreadSaved` as well.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2768, opened 2026-08-14
[ADVAPP-2805]: Decouple notification branding from Division

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2805

### Technical Description

Replace NotificationSetting model with tenant-wide Notification Settings

No

### Cleanup Tasks

.cleanup-tasks/2026_08_14_notification_settings_feature.md

---


app/Settings/NotificationSettings.php
```diff
@@ -34,43 +34,30 @@
...
-    protected static string $resource = NotificationSettingResource::class;
+    public ?string $name = null;
+
+    public ?string $from_name = null;
+
+    public ?string $description = null;
+
+    public null $logo = null;
+
+    public ?Color $primary_color = null;
```
payal-canyon, line 52: In `resources/views/vendor/mail/html/layout.blade.php`, could we update the following line to use the enum’s backed value?
```php
$color = \Filament\Support\Colors\Color::convertToRgb(\Filament\Support\Colors\Color::all()[$settings?->primary_color ?? 'blue'][600]);
```
`primary_color` is now a `Color` enum, so using it directly as an array key will cause a `TypeError` when a colour is set. Could we change it to:

```php
$color = Color::convertToRgb(Color::all()[$settings?->primary_color?->value ?? 'blue'][600]);
```

Also, could we import `Filament\Support\Colors\Color` at the top of the `@php` block instead of using the inline FQCN, as done in `header.blade.php`?

resources/views/vendor/mail/html/header.blade.php
```diff
@@ -34,17 +34,20 @@
 @props(['url', 'settings' => null])
 @php
     use App\Models\SettingsProperty;
+    use App\Settings\NotificationSettings;
     use AdvisingApp\Theme\Settings\ThemeSettings;
 
     $themeSettings = app(ThemeSettings::class);
     $logo = $themeSettings->getSettingsPropertyModel('theme.is_logo_active')->getFirstMedia('logo');
+
+    $notificationLogo = $settings ? NotificationSettings::getSettingsPropertyModel('notifications.logo') : null;
```
payal-canyon: could we gate this branding lookup with `NotificationSettingsFeature` instead of `$settings`?

`getSettingsPropertyModel()` has a non-nullable return type but can return `null` when the property doesn’t exist, causing a `TypeError`. During the zero-downtime deployment window, the migration may not have run yet, while `$settings` is always non-null, so this call can fatal.

Could we use the feature flag here instead?
```javascript
$notificationLogo = NotificationSettingsFeature::active()
    ? NotificationSettings::getSettingsPropertyModel('notifications.logo')
    : null;
```
Alternatively, having `MailMessage::settings()` when the flag is `inactive` would protect both the `logo` and `colour lookups` with a single guard.

app/Filament/Pages/ManageNotificationSettings.php
```diff
@@ -34,20 +34,39 @@
...
+    protected static string | null $navigationLabel = 'Notifications';
+
+    protected static ?int $navigationSort = 120;
+
+    public static function canAccess(): bool
+    {
+        $user = auth()->user();
+        assert($user instanceof User);
+
+        return NotificationSettingsFeature::active() && $user->can('settings.view-any') && parent::canAccess();
```
payal-canyon, line 71: Could we restore the `settings.*.update` write permission on `ManageNotificationSettings`? The old `EditNotificationSetting` flow enforced this through `NotificationSettingPolicy::update()`, but the new page only checks `settings.view-any`, allowing read-only users to modify tenant-wide email branding, including password reset/setup emails.

Could we add `->disabled(! auth()->user()->can('settings.*.update'))` to the schema and guard `save()`/`getFormActions()` when the permission is missing? This restores the authorization enforced by the previous page.

It would also be good to add a test confirming that a user with only `settings.view-any` can access the page but cannot persist changes.

database/migrations/2026_08_14_162031_create_notification_settings.php
```diff
@@ -34,36 +34,29 @@
...
-        ->assertTableBulkActionVisible(DeleteBulkAction::class);
-});
+use Illuminate\Support\Facades\DB;
+use Spatie\LaravelSettings\Migrations\SettingsMigration;
+
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            $this->migrator->add('notifications.name');
```
payal-canyon: Could we make this settings migration idempotent by wrapping each `add()` in a `try/catch` for `SettingAlreadyExists`?

Settings migrations run per tenant, so a later migration failure could leave some properties already created. On retry, this migration would then fail with `SettingAlreadyExists`.

tests/Tenant/Feature/Filament/Pages/ManageNotificationSettingsTest.php
```diff
@@ -0,0 +1,104 @@
...
+use App\Settings\NotificationSettings;
+use CanyonGBS\Common\Enums\Color;
+
+use function Pest\Laravel\actingAs;
+use function Pest\Laravel\get;
+use function Pest\Livewire\livewire;
+use function Tests\asSuperAdmin;
+
+beforeEach(function () {
+    NotificationSettingsFeature::activate();
```
payal-canyon: I don’t think we need to activate the feature flag in the tests if it’s already activated by a migration, since the migrations run before the tests start.

The main case where we’d need to change the flag in a test would be when we’re deactivating it to test the disabled path, or in the less common case of a true unit test where the activation migration isn’t included.

app/Filament/Pages/ManageNotificationSettings.php
```diff
@@ -62,21 +81,15 @@ public function form(Schema $schema): Schema
                     ->string()
                     ->maxLength(150)
                     ->autocomplete(false),
-                Textarea::make('description')
+                TextInput::make('description')
```
payal-canyon: Could we keep `description` as a `Textarea` here? The previous Create/Edit pages used `Textarea::make('description')`, and the underlying column is `longText`. The ticket also shows a multi-line field, so changing it to `TextInput` appears to be an unintended UX regression.

Could we restore `Textarea::make('description')`?

app/Filament/Pages/ManageNotificationSettings.php
payal-canyon, line 80: Could you please double-check whether `name` and `description` are still needed? On main, both were only used by the multi-package admin UI, and those call sites have been removed in this branch. I also traced the remaining references, and neither value appears to be read anywhere now, so keeping `name` as `->required()` seems to require a value that has no effect.

If you confirm they’re no longer needed, could we remove them from the settings class, form, settings migration, seed migration, and related tests?

database/migrations/2026_08_14_214258_tmp_seed_notification_settings.php
```diff
@@ -0,0 +1,86 @@
...
+                $settings->name = $notificationSetting->getAttribute('name');
+                $settings->from_name = $notificationSetting->getAttribute('from_name');
+                $settings->description = $notificationSetting->getAttribute('description');
+                $settings->primary_color = Color::tryFrom((string) $notificationSetting->getAttribute('primary_color'));
+
+                $settings->save();
+
+                $logo = $notificationSetting->getFirstMedia('logo');
+
+                $logo?->copy(NotificationSettings::getSettingsPropertyModel('notifications.logo'), 'logo');
```
payal-canyon: Hey @Orrison , I’m coming across this kind of scenario for the first time, so I’d really appreciate your opinion here.

Would it be better to move the media `copy()` outside the `DB::transaction`? `Media::copy()` performs S3 I/O (download + re-upload), so keeping it inside the transaction means the transaction stays open during those network calls and could roll back the migration if the S3 operation fails.

Also, would you recommend passing `'s3'` explicitly to `copy()` to match the `->disk('s3')` used by the form?

What do you think would be the better approach here?

tests/Tenant/Feature/Filament/Pages/ManageNotificationSettingsTest.php
```diff
@@ -0,0 +1,104 @@
...
+});
+
+it('can render the manage notification settings page', function () {
+    asSuperAdmin();
+
+    get(ManageNotificationSettings::getUrl())
+        ->assertOk();
+});
+
+it('requries proper permissions to access', function () {
```
payal-canyon: Could we fix the typo in the test description?

```php
it('requires proper permissions to access')
```

payal-canyon (changes requested):

app/Filament/Pages/ManageNotificationSettings.php
rebecca-canyon: I believe that removing these fields is more of a Product decision, regardless of their utility now. So I think it makes more sense to leave them for the time being, even if we are later told to remove them.

database/migrations/2026_08_14_214258_tmp_seed_notification_settings.php
rebecca-canyon: Rerequesting review for other changes while awaiting input from @Orrison on this

app/Filament/Pages/ManageNotificationSettings.php
```diff
@@ -34,25 +34,46 @@
...
+        assert($user instanceof User);
+
+        return NotificationSettingsFeature::active() && $user->can('settings.view-any') && parent::canAccess();
+    }
 
     public function form(Schema $schema): Schema
     {
         return $schema
             ->columns(1)
+            ->disabled(! auth()->user()->can('settings.*.update'))
```
payal-canyon, line 78: `settings.*.update` is the correct ability, and `->disabled()` handles the UI side, but it doesn’t enforce the permission server-side. Could we also add `save()` and `getFormActions()` guards, following the same pattern as `ManageDisplaySettings` and `ManageInstitutionDetailsSettings`?

One small note: please keep `settings.*.update` here rather than copying those checks directly, since those pages still reference the old `product_admin.*.update` permission.

The current test only verifies that the fields are disabled, so it wouldn’t catch a read-only user modifying values through the request. Could we also add a test that fills the form and calls `save()` as a `settings.view-any`-only user, then asserts the settings remain unchanged?

payal-canyon (commented):
Hey @Orrison ,
While reviewing this one again, I noticed something unrelated to this PR but worth flagging `ManageDisplaySettings`, `ManageInstitutionDetailsSettings`, and five other files still use the old `product_admin.*` permissions, which were renamed to `settings.*` in `2025_07_14_105203_data_rename_product_admin_to_settings_permissions` and no longer exist in the permissions table.

Those checks now only pass for super/partner admins via the `Gate::after` fallback. I think it would be worth creating a separate ticket to address this. WDYT?

database/migrations/2026_08_14_214258_tmp_seed_notification_settings.php
Orrison: @payal-canyon and @rebecca-canyon I think it is a good idea to have it within the DB transaction. If there is an IO failure we actually want WANT all of this to roll back because that means that it failed to do what it was supposed to.

Technically I think `s3` is the default so it is implied here, but IMO it never hurts to be explicit, so I would put it here.
Though are we 100% sure that this is using S3 right now? and not the public one? Is this logo being used in emails? If so are we generated temp URLS for it it? We will need to make a separate ticket to fix all that if so

payal-canyon (approved):

PR #2775, opened 2026-08-20
[ADVAPP-2810]: Remove Division permissions and delete the Division module

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2810

### Technical Description

Remove Division module, reference in PermissionDescription, and seeders from DatabaseSeeder and NewTenantSeeder

No

### Cleanup Tasks

N/A

---


database/migrations/2026_08_20_023543_drop_divisions_table.php
```diff
@@ -0,0 +1,107 @@
...
+        'api',
+    ];
+
+    public function up(): void
+    {
+        DB::transaction(function () {
+            collect($this->guards)
+                ->each(fn (string $guard) => $this->deletePermissions(array_keys($this->permissions), $guard));
+
+            Schema::dropIfExists('divisions');
```
payal-canyon, line 74: Could we also clean up the related `audits` records in this migration?

Since `Division` was an `audited` model, existing `audits` rows with `auditable_type = 'division'` will remain after the `divisions` table and morph mapping are removed. These records would still appear in the `audit` list but could no longer be filtered by their model type.

Could we add the following inside the existing `DB::transaction()`?
```php
DB::table('audits')
    ->where('auditable_type', 'division')
    ->delete();
```

payal-canyon (changes requested):

database/migrations/2026_08_20_023543_drop_divisions_table.php
```diff
@@ -0,0 +1,109 @@
...
+
+            DB::table('audits')->where('auditable_type', 'divisions')->delete();
```
payal-canyon: Could we change `divisions` to `division` here?

The `Division` audit morph alias is singular (`division`), so the current query matches no rows and leaves the orphaned audit records behind.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2776, opened 2026-08-20
[ADVAPP-2883]: Introduce support for mobile login capabilities with website switching and SSO notices

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2883

### Technical Description

- Intercepts the `AdvisingAppMobile` user-agent to conditionally render mobile-specific UI adjustments on the Filament login view.
- Injects a "Switch tenant" escape hatch link below the login form that fires the `advisingappmobile://switch-tenant` protocol, returning mobile users to the domain-entry screen.
- Hides the Google Single Sign-On button explicitly for mobile users (due to embedded WebView limitations) and replaces it with an inline unavailability notice.
- Adds Pest test coverage verifying that mobile-specific controls and notices are displayed exclusively for mobile app user-agents and omitted for standard browsers.

> No

### Cleanup Tasks

> N/A

---


app-modules/authorization/resources/views/login.blade.php
```diff
@@ -64,14 +64,20 @@
             {{ $this->form }}
 
             <x-filament::actions :actions="$this->getFormActions()" :full-width="$this->hasFullWidthFormActions()" />
-            @if (count($this->getSsoFormActions()) > 0)
+            @if (count($this->getSsoFormActions()) > 0 || $this->isGoogleSsoUnavailableInMobileApp())
                 <small class="text-gray-800 dark:text-gray-300">or log in with single sign-on</small>
             @endif
```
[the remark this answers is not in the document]
nency-canyon: Removed the `|| $this->isGoogleSsoUnavailableInMobileApp()` clause from the condition. The "or log in with single sign-on" hint now only renders when `count($this->getSsoFormActions()) > 0`. The "Google sign-in isn't available in the mobile app" message stays as its own independent block below, so mobile users with only Google SSO configured no longer see the SSO hint with nothing under it.

app-modules/authorization/tests/Tenant/Filament/Pages/Auth/LoginTest.php
```diff
@@ -0,0 +1,87 @@
...
+}
+
+it('does not show the switch tenant control or a google unavailable message for a normal browser', function () {
+    enableGoogleSso();
+
+    get(route('filament.admin.auth.login'), ['User-Agent' => 'Mozilla/5.0'])
+        ->assertOk()
+        ->assertDontSee('Switch tenant')
+        ->assertDontSee("Google sign-in isn't available in the mobile app")
+        ->assertSee('Google');
```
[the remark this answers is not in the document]
nency-canyon: Added `false` to `assertDontSee()` to disable escaping. The text is rendered as raw/literal HTML in the Blade view, not via `{{ }}`, so the apostrophe is never HTML-escaped in the response. Without `false`, the assertion compared against the escaped string and could pass even when the real text was present.

app-modules/authorization/resources/views/login.blade.php
```diff
@@ -88,6 +94,18 @@ class="cursor-pointer"
                 @endif
             </x-filament::link>
         @endif
+
+        @if ($this->isMobileApp())
+            <div class="mt-6 flex justify-center">
+                <x-filament::link
+                    :href="\AdvisingApp\Authorization\Filament\Pages\Auth\Login::SWITCH_TENANT_URL"
```
Orrison, line 101: Hmm, is the expectation that this would result in the stored data (stored url) being cleared in the mobile app? We would want it to result in that happening.

app-modules/authorization/resources/views/login.blade.php
```diff
@@ -72,6 +72,12 @@
                 :actions="$this->getSsoFormActions()"
                 :full-width="$this->hasFullWidthFormActions()"
             />
+
+            @if ($this->isGoogleSsoUnavailableInMobileApp())
+                <p class="text-sm text-gray-500 dark:text-gray-400">
+                    Google sign-in isn't available in the mobile app - please use another sign-in method.
+                </p>
+            @endif
```
Orrison, line 80: Instead of doing this, I wonder if it makes more sense to have the `getSsoFormActions` be modified to make the google sso action be disabled and then I wonder if an action can have `description()` added to it to add text under the button?

Orrison (changes requested):

app-modules/authorization/resources/views/login.blade.php
nency-canyon: Yes, exactly. The web side renders the link with the agreed deep link scheme.

When clicked inside the mobile app's WebView, the native mobile app intercepts `advisingappmobile://switch-tenant`, clears the saved tenant domain from local storage/preferences (`clearTenantDomain()`), and navigates the user back to the initial `DomainEntryScreen`.

app-modules/authorization/resources/views/login.blade.php
nency-canyon: I explored this approach! Filament `Action` components don't support a `description()` method. While they do support `tooltip()`, tooltips rely on mouse hover events which are not accessible on mobile touchscreens (iOS/Android WebViews do not trigger hover states on disabled elements). 

Showing a disabled button without hover gives mobile users no way to see why sign-in is unavailable and makes the button appear broken. 

To ensure the best mobile UX and stay aligned with the ticket's Acceptance Criteria:
- On mobile (`AdvisingAppMobile`), we hide the unsupported Google button and render the clear inline message: *"Google sign-in isn't available in the mobile app - please use another sign-in method."*
- Normal desktop/browser users remain completely unaffected (Google SSO renders and functions as usual).

app-modules/authorization/resources/views/login.blade.php
```diff
@@ -88,6 +94,18 @@ class="cursor-pointer"
                 @endif
             </x-filament::link>
         @endif
+
+        @if ($this->isMobileApp())
+            <div class="mt-6 flex justify-center">
+                <x-filament::link
```
Orrison: Looking at the screenshot, can we actually use a filament button instead? That will help the click target process on mobile and help users know it is clickable.

app-modules/authorization/resources/views/login.blade.php
```diff
@@ -72,6 +72,12 @@
                 :actions="$this->getSsoFormActions()"
                 :full-width="$this->hasFullWidthFormActions()"
             />
+
+            @if ($this->isGoogleSsoUnavailableInMobileApp())
+                <p class="text-sm text-gray-500 dark:text-gray-400">
+                    {!! \AdvisingApp\Authorization\Filament\Pages\Auth\Login::GOOGLE_SSO_UNAVAILABLE_IN_MOBILE_APP_MESSAGE !!}
```
Orrison: since `GOOGLE_SSO_UNAVAILABLE_IN_MOBILE_APP_MESSAGE` is just text I don't think we need `{!! !!}` which renders html. Did you try with just `{{ }}`?

Orrison (changes requested):

app-modules/authorization/resources/views/login.blade.php
nency-canyon: Addressed your feedback and added Filament button instead. Please verify.

app-modules/authorization/resources/views/login.blade.php
nency-canyon: Thanks for the feedback, switched to `{{ }}` safe escaping since the message is plain text.

nency-canyon:
Also Updated the screenshots.

Orrison (approved):

PR #2777, opened 2026-08-20
[ADVAPP-2901]: Some users have reported they are unable to send SMS messages in Demo mode

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2901

### Technical Description

> Solve the issue of users not being able to send SMS in demo mode.

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/notification/src/Notifications/Channels/SmsChannel.php
+++ b/app-modules/notification/src/Notifications/Channels/SmsChannel.php
@@ -197,10 +197,18 @@ class SmsChannel
             try {
                 if ($result->success) {
                     $smsMessage->quota_usage = $this->determineQuotaUsage($result, $smsMessage);
-                    $smsMessage->external_reference_id = match ($twilioSettings->provider) {
-                        SmsMessagingProvider::Twilio => $result->message->sid,
-                        SmsMessagingProvider::Telnyx => $result->message->id, // @phpstan-ignore property.notFound
-                    };
+
+                    if ($twilioSettings->is_demo_mode_enabled) {
+                        // In demo mode, the result is always a fake Twilio `MessageInstance`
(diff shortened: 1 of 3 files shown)


app-modules/notification/src/Notifications/Channels/SmsChannel.php
```diff
@@ -197,10 +197,15 @@ public function send(object $notifiable, Notification $notification): void
...
-                    };
+
+                    if ($twilioSettings->is_demo_mode_enabled) {
+                        $smsMessage->external_reference_id = $result->message->sid;
+                    } else {
+                        $smsMessage->external_reference_id = match ($twilioSettings->provider) {
+                            SmsMessagingProvider::Twilio => $result->message->sid,
+                            SmsMessagingProvider::Telnyx => $result->message->id, // @phpstan-ignore property.notFound
+                        };
+                    }
```
Orrison: Hmm, I think a better fix actually would be to change this `$result` in the `else` statement above to be changed to return a mocked expected `Telnyx` result if the `$twilioSettings->provider` is `SmsMessagingProvider::Telnyx`

That was it returns the proper body / content for this and everything else going forward

Orrison (changes requested):

app-modules/notification/tests/Tenant/Notifications/Channels/SmsChannelSmsMessageTest.php
Orrison, line 1: There should also be a process that sends a fake response back "from" the prospect/student.

Can you please add a test ensuring that is happening so we have end-to-end proof that is happening before we ship this.

Orrison (changes requested):

Orrison (approved):

PR #2782, opened 2026-08-24
[ADVAPP-2885]: Some users have reported that in campaigns, the population group drop‑down only shows their population groups

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2885

### Technical Description

> Some users have reported that in campaigns, the population group drop‑down only shows their population groups

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
+++ b/app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
@@ -36,6 +36,7 @@
 
 namespace AdvisingApp\Campaign\Filament\Forms\Components;
 
+use AdvisingApp\Campaign\Enums\GroupOwnership;
 use AdvisingApp\Campaign\Models\Campaign;
 use AdvisingApp\Group\Enums\GroupModel;
 use AdvisingApp\Group\Models\Group;
@@ -48,35 +49,36 @@ use Illuminate\Database\Eloquent\Builder;
 
 class PopulationGroupSelector
 {
-    public const OWNERSHIP_MINE = 'mine';
-
-    public const OWNERSHIP_DEPARTMENT = 'department';
-
-    public const OWNERSHIP_ALL = 'all';
-
     /**
      * @return array<int, Select | ToggleButtons>
      */
     public static function make(?Closure $afterSegmentIdUpdated = null): array
     {
+        $availablePopulationTypes = self::availablePopulationTypes();
+
         $segmentIdField = Select::make('segment_id')
             ->label('Population Group')
             ->options(function (Get $get) {
                 $populationType = $get('population_type');
-                $groupOwnership = $get('group_ownership');
+                $groupOwnership = $get->enum('group_ownership', GroupOwnership::class, isNullable: true);
 
                 if (blank($populationType) || blank($groupOwnership)) {
+                    return [];
+                }
+
+                // Group Ownership is client-side state, so "All" must be re-authorized here rather than trusted from the form.
+                if (($groupOwnership === GroupOwnership::All) && auth()->user()->cannot('group.view-any') && auth()->user()->cannot('group.*.view')) {
                     return [];
                 }
 
                 $query = Group::query()->where('model', $populationType);
 
                 match ($groupOwnership) {
-                    self::OWNERSHIP_MINE => $query->where('user_id', auth()->id()),
-                    self::OWNERSHIP_DEPARTMENT => $query->whereHas('user', function (Builder $query) {
+                    GroupOwnership::Mine => $query->where('user_id', auth()->id()),
+                    GroupOwnership::Department => $query->whereHas('user', function (Builder $query) {
                         $query->whereRelation('department.users', 'id', auth()->id());
                     }),
-                    default => null,
+                    GroupOwnership::All => null,
                 };
 
                 return $query->pluck('name', 'id');
@@ -92,26 +94,29 @@ class PopulationGroupSelector
         return [
             ToggleButtons::make('population_type')
                 ->label('Population Type')
-                ->options([
-                    GroupModel::Student->value => 'Students',
-                    GroupModel::Prospect->value => 'Prospects',
-                ])
+                ->options(collect($availablePopulationTypes)->mapWithKeys(fn (GroupModel $type) => [
+                    $type->value => $type === GroupModel::Student ? 'Students' : 'Prospects',
+                ]))
                 ->inline()
                 ->live()
                 ->dehydrated(false)
                 ->required()
-                ->afterStateHydrated(fn (Set $set, ?Campaign $record) => $set('population_type', ($record?->group->model ?? GroupModel::default())->value))
+                ->hidden(count($availablePopulationTypes) <= 1)
+                ->afterStateHydrated(fn (Set $set, ?Campaign $record) => $set(
+                    'population_type',
+                    self::defaultPopulationType($record, $availablePopulationTypes),
+                ))
                 ->afterStateUpdated(fn (Set $set) => $set('segment_id', null)),
             ToggleButtons::make('group_ownership')
                 ->label('Group Ownership')
                 ->options(function () {
                     $options = [
-                        self::OWNERSHIP_MINE => 'My Groups',
-                        self::OWNERSHIP_DEPARTMENT => "My Department's Groups",
+                        GroupOwnership::Mine->value => GroupOwnership::Mine->getLabel(),
+                        GroupOwnership::Department->value => GroupOwnership::Department->getLabel(),
                     ];
 
                     if (auth()->user()->canAny(['group.view-any', 'group.*.view'])) {
-                        $options[self::OWNERSHIP_ALL] = 'All Groups';
+                        $options[GroupOwnership::All->value] = GroupOwnership::All->getLabel();
                     }
 
                     return $options;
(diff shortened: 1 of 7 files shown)


app-modules/campaign/src/Filament/Resources/Campaigns/Pages/CreateCampaign.php
```diff
@@ -107,11 +107,15 @@ protected function getSteps(): array
                     Select::make('segment_id')
                         ->label('Population Group')
                         ->options(function () {
-                            return Group::query()
-                                ->whereHas('user', function (EloquentBuilder $query) {
+                            $query = Group::query();
+
+                            if (! auth()->user()->canAny(['group.view-any', 'group.*.view'])) {
+                                $query->whereHas('user', function (EloquentBuilder $query) {
```
payal-canyon: Could you please review the ticket and implement the changes based on the complete requirements? At the moment, I can’t see the requested enhancements, such as the Population Type and Group Ownership filters, reflected in this PR. Thank you!

payal-canyon (changes requested):

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,149 @@
...
+                $groupOwnership = $get('group_ownership');
+
+                if (blank($populationType) || blank($groupOwnership)) {
+                    return [];
+                }
+
+                $query = Group::query()->where('model', $populationType);
+
+                match ($groupOwnership) {
+                    self::OWNERSHIP_MINE => $query->where('user_id', auth()->id()),
```
payal-canyon: `All Groups` can be selected without the required permission.
The permission check currently only controls whether the `All Groups` button is rendered. Since `group_ownership` is browser-controlled Livewire state and default => `null` leaves the query unscoped, a user could set it to `all` (or another unexpected value) and see all groups they are licensed to access.

Could we enforce this server-side and fail closed? I think a `GroupOwnership` enum with `->enum(GroupOwnership::class)` would make this cleaner, then explicitly check the permission for All and reject invalid values in the query.

This would prevent the query from relying solely on client-controlled state and also remove the repeated ownership string literals.

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,149 @@
...
+        if ($afterSegmentIdUpdated) {
+            $segmentIdField->afterStateUpdated($afterSegmentIdUpdated);
+        }
+
+        return [
+            ToggleButtons::make('population_type')
+                ->label('Population Type')
+                ->options([
+                    GroupModel::Student->value => 'Students',
+                    GroupModel::Prospect->value => 'Prospects',
```
payal-canyon: Population Type currently shows both options and defaults to `Student`. For a Prospect-only user, they land on Students with an empty Population Group list and no indication that they need to switch the type.

Could we filter the options based on the user's licences, default to a population type they have access to, and hide the toggle when only one option is available? This would follow the same pattern used in `CreateGroup.php`.

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/CreateCampaignTest.php
```diff
@@ -102,3 +111,138 @@
...
+            function (Select $field) use ($otherUsersGroup) {
+                expect($field->getOptions())->not->toHaveKey($otherUsersGroup->getKey());
+
+                return true;
+            },
+        );
+});
+
+it('hides the All Groups ownership option for users without the group.*.view or group.view-any permission', function () {
+    $user = User::factory()->licensed(LicenseType::cases())->create();
```
payal-canyon, line 146: The current test only checks that the all option is hidden, but it doesn't verify that the query also rejects a manually supplied `group_ownership = 'all'`. Could we add a test to ensure another user's group is not returned when that value is set directly?

Also, the `only offers the user's own population groups by default` test currently only asserts that other groups are absent. Could we add a group owned by the current user and assert that it is present as well? Otherwise, the test would also pass with an empty options list.

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -54,21 +52,13 @@ class EditCampaign extends EditRecord
...
-                                $query->whereKey(auth()->id())->orWhereRelation('department.users', 'id', auth()->id());
-                            })
-                            ->pluck('name', 'id');
-                    })
-                    ->searchable()
-                    ->required(),
-                Toggle::make('enabled'),
+                ...PopulationGroupSelector::make(),
+                Toggle::make('enabled')
+                    ->helperText('Toggle this off to set your campaign to a draft state.'),
```
payal-canyon: `Draft state` is misleading here.
The helper text appears to be copied from the create wizard, but an existing campaign doesn't have a draft state. Here, disabling the campaign pauses its execution, and the archive action also refers to this as disabling the campaign.
Could we update the helper text to better reflect its actual behaviour?
```php
Toggle::make('enabled')
    ->helperText('Toggle this off to pause the campaign; no further journey steps will run.');
```

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
```diff
@@ -54,21 +52,13 @@ class EditCampaign extends EditRecord
...
-                Select::make('segment_id')
-                    ->label('Population Group')
-                    ->options(function () {
-                        return Group::query()
-                            ->whereHas('user', function (Builder $query) {
-                                $query->whereKey(auth()->id())->orWhereRelation('department.users', 'id', auth()->id());
-                            })
-                            ->pluck('name', 'id');
-                    })
-                    ->searchable()
```
payal-canyon, line 69: Hey @Orrison , I’d like to get your opinion on this one:
### Changing Population Type can invalidate existing journey steps

**What happens today**

A campaign created for a **Student** group can be changed to a **Prospect** group on Edit without any warning. The existing journey steps are not updated, so the population-specific steps - **Interaction, Care Team, and Tags** - can retain values from the previous population type.

The Journey Steps tab may then show those fields as blank, while execution can still use the stored values against the wrong population.

**Why I'm raising it now**

This is pre-existing behaviour - we already allowed switching between population types. However, This ticket introduces an explicit **Population Type [Students] [Prospects]** toggle, making the change much easier to trigger. I think we should define the expected behaviour before shipping it.

**Options**

* **A - Block the change:** If the campaign has journey steps, prevent changing population type and ask the user to remove the steps first. Safest and non-destructive.
* **B - Confirm and clear:** Warn the user that changing population type will remove the existing journey steps, then clear them on confirmation. This matches the Create flow but is destructive.
* **C - Leave as-is:** it can silently produce incorrect records.

So, my question is: do we need to fix this? If yes, which approach would you prefer?
Let me know your thoughts. Thanks!

payal-canyon (changes requested):

app-modules/campaign/src/Filament/Resources/Campaigns/Pages/EditCampaign.php
Orrison: Ah! Hmm, let's actually go with making it so you CANNOT change the population type during an edit.

You can change other things. But you are right, changing the population type when it has steps would cause issues and I think the easiest thing to do right now is to just stop that from happening.

If users need it for some reason later we can find a safer way to handle it then.

And let's take care of that in this ticket.

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,187 @@
...
+                ->label('Population Type')
+                ->options(collect($availablePopulationTypes)->mapWithKeys(fn (GroupModel $type) => [
+                    $type->value => $type === GroupModel::Student ? 'Students' : 'Prospects',
+                ]))
+                ->inline()
+                ->live()
+                ->dehydrated(false)
+                ->required()
+                ->hidden(count($availablePopulationTypes) <= 1)
+                ->disabled(fn (?Campaign $record) => $record !== null)
```
payal-canyon, line 126: Kevin’s decision was that the population type cannot be changed during an edit. While `->disabled()` correctly locks the field in the UI, it doesn’t appear to prevent the value from being changed through Livewire state.

This means `population_type` can still be tampered with, causing `segment_id` to load groups from the other population type and potentially allowing an invalid combination to be saved.

Could we derive the population type from the existing campaign record when editing, rather than trusting the form state? This would ensure the `segment_id` options and validation always use the campaign’s original population type.

Also, could we add a test that attempts to change `population_type` and `segment_id` during an `edit` and verifies that the save is rejected? The current test only checks that the field is disabled, which wouldn’t catch this bypass.

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,187 @@
...
+                if (($groupOwnership === GroupOwnership::All) && auth()->user()->cannot('group.view-any') && auth()->user()->cannot('group.*.view')) {
+                    return [];
+                }
+
+                $query = Group::query()->where('model', $populationType);
+
+                match ($groupOwnership) {
+                    GroupOwnership::Mine => $query->where('user_id', auth()->id()),
+                    GroupOwnership::Department => $query->whereHas('user', function (Builder $query) {
+                        $query->whereRelation('department.users', 'id', auth()->id());
```
payal-canyon: The department filter uses a nested `users -> departments -> users` `EXISTS` chain, while `:181` uses a direct `team_id` comparison for the same relationship. Could we use the direct `team_id` check here as well?

This is simpler, uses a single subquery, and keeps both sides of the ticket’s “map left to right” rule consistent.

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,187 @@
...
+                    }),
+                    GroupOwnership::All => null,
+                };
+
+                return $query->pluck('name', 'id');
```
payal-canyon: There’s no `orderBy('name')`, so the full-tenant “All Groups” list may be returned in an inconsistent order. Could we add `orderBy('name')` here to keep the options consistently sorted?

app-modules/campaign/src/Filament/Forms/Components/PopulationGroupSelector.php
```diff
@@ -0,0 +1,187 @@
...
+
+        if ($afterSegmentIdUpdated) {
+            $segmentIdField->afterStateUpdated($afterSegmentIdUpdated);
+        }
+
+        return [
+            ToggleButtons::make('population_type')
+                ->label('Population Type')
+                ->options(collect($availablePopulationTypes)->mapWithKeys(fn (GroupModel $type) => [
+                    $type->value => $type === GroupModel::Student ? 'Students' : 'Prospects',
```
payal-canyon: Could we use `GroupModel::getPluralLabel()` here instead of the ternary? The current approach would incorrectly label any future `GroupModel` case as “Prospects.” We can also follow the existing `Str::ucfirst($type->getPluralLabel())` pattern from `ListCampaigns.php:80`.

app-modules/campaign/tests/Tenant/Filament/Resources/Campaigns/Pages/CreateCampaignTest.php
```diff
@@ -102,3 +111,192 @@
...
+    $user->givePermissionTo('campaign.create');
+
+    app(PermissionRegistrar::class)->forgetCachedPermissions();
+
+    actingAs($user);
+
+    $otherUsersGroup = Group::factory()->student()->create();
+
+    livewire(CreateCampaign::class)
+        ->fillForm(['group_ownership' => 'all'])
```
payal-canyon: A few tests still hardcode ownership strings in this file and `EditCampaignTest.php`. Could we use `GroupOwnership::All->value`, etc., instead? This helps keep the tests aligned with the enum if the values change.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2783, opened 2026-08-26
[ADVAPP-2900]: Some Settings pages are inaccessible to users who have permission

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2900

### Technical Description

Ensure product_admin permissions are fully replaced with settings permissions; add tests

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/campaign/src/Filament/Pages/ManageCampaignSettings.php
+++ b/app-modules/campaign/src/Filament/Pages/ManageCampaignSettings.php
@@ -65,7 +65,7 @@ class ManageCampaignSettings extends SettingsPage
         /** @var User $user */
         $user = auth()->user();
 
-        return $user->can(['product_admin.view-any', 'product_admin.*.view', 'product_admin.*.update']);
+        return $user->can(['settings.view-any', 'settings.*.view', 'settings.*.update']);
     }
 
     public function form(Schema $schema): Schema
--- a/app-modules/campaign/tests/Tenant/Filament/Pages/ManageCampaignSettingsTest.php
+++ b/app-modules/campaign/tests/Tenant/Filament/Pages/ManageCampaignSettingsTest.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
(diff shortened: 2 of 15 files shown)


mxm1070 (changes requested):
**Missing test coverage: `ManageStudentInformationSystemSettings`**

Six other affected pages and actions have new access tests: `ManageDisplaySettings`, `ManageInstitutionDetailsSettings`, `ManageCollegeBrandingSettings`, `ManageProspectConversionSettings`, `ManageCampaignSettings`, and `ListStudents`. `ManageStudentInformationSystemSettings` has no test. The ticket lists this page as one of the five pages that need a test.

The fix to this page is correct. But the method `canAccess()` on this page checks only `$user->isSuperAdmin()`. This check existed before this PR and is not related to it. The other four pages check a `settings.*` permission instead.

`AuthServiceProvider` also sets a `Gate::after` rule. This rule gives every super admin every ability. As a result, the `settings.*.update` checks in `form()`, `save()`, and `getFormActions()` are unreachable. A user without super admin status cannot reach the page. A super admin always passes the permission check, no matter what permissions we assign. So the test pattern that denies access without permission, used on the other four pages, does not work here.

This repository already has a test for this same case. `AmazonS3Test.php` also gates `canAccess()` on `isSuperAdmin()` only:

```php
it('prevents access to the Amazon S3 Settings when you do not have the necessary permissions', function () {
    $user = User::factory()->create();
    actingAs($user);
    get(AmazonS3::getUrl())->assertForbidden();
});

it('allows access to the Amazon S3 Settings when you do have the necessary permissions', function () {
    $user = User::factory()->create();
    asSuperAdmin($user);
    get(AmazonS3::getUrl())->assertOk();
});
```

Can we add the same pair of tests for `ManageStudentInformationSystemSettings`?

mxm1070 (approved):

PR #2788, opened 2026-08-26
[ADVAPP-2916]: Stop errors about Azure SSO users not having profile pictures in Azure from appearing in Sentry

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2916

### Technical Description

Changes the Azure profile picture retrieval process to NOT send 404 exceptions to Sentry. This is not something we can do anything about if they don't have a profile picture in Azure, so it is just noise.

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/authorization/src/Http/Controllers/SocialiteController.php
+++ b/app-modules/authorization/src/Http/Controllers/SocialiteController.php
@@ -101,21 +101,21 @@ class SocialiteController extends Controller
         }
 
         if ($provider === SocialiteProvider::Azure) {
-            // Retry transient Azure failures (connection errors, rate limits, server errors) but not client errors.
-            $response = Http::withToken($socialiteUser->token)
-                ->dontTruncateExceptions()
-                ->retry(3, 500, when: function (?Throwable $exception): bool {
-                    if ($exception instanceof ConnectionException) {
-                        return true;
-                    }
+            try {
+                // Retry transient Azure failures (connection errors, rate limits, server errors) but not client errors.
+                $response = Http::withToken($socialiteUser->token)
+                    ->dontTruncateExceptions()
+                    ->retry(3, 500, when: function (?Throwable $exception): bool {
+                        if ($exception instanceof ConnectionException) {
+                            return true;
+                        }
 
-                    return $exception instanceof RequestException
-                        && ($exception->response->serverError() || $exception->response->status() === 429);
-                }, throw: false)
-                ->get('https://graph.microsoft.com/v1.0/me/photo/$value');
+                        return $exception instanceof RequestException
+                            && ($exception->response->serverError() || $exception->response->status() === 429);
+                    }, throw: false)
+                    ->get('https://graph.microsoft.com/v1.0/me/photo/$value');
 
-            if ($response->successful()) {
-                try {
+                if ($response->successful()) {
                     $mimeType = $response->header('Content-Type');
 
                     if (in_array($mimeType, ['image/png', 'image/jpeg', 'image/webp', 'image/jpg', 'image/svg+xml'])) {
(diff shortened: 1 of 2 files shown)


app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -101,37 +101,45 @@ public function callback(SocialiteProvider $provider)
...
+            $response = Http::withToken($socialiteUser->token)
+                ->dontTruncateExceptions()
+                ->retry(3, 500, when: function (?Throwable $exception): bool {
+                    if ($exception instanceof ConnectionException) {
+                        return true;
+                    }
+
+                    return $exception instanceof RequestException
+                        && ($exception->response->serverError() || $exception->response->status() === 429);
+                }, throw: false)
```
mxm1070: ### Must-fix: uncaught `ConnectionException` can break the entire Azure SSO login flow

The `Http::...->retry(3, 500, when: ..., throw: false)->get(...)` call is no longer wrapped in a try/catch — only the post-success processing below it is. `throw: false` only suppresses `Response::throw()` for HTTP-status failures; it does **not** stop a `ConnectionException` from propagating. `marshalConnectionException()` and the global `retry()` helper both re-throw it unconditionally once all 3 attempts are exhausted.

Concretely: if fetching the Azure profile photo hits a transient connection failure (DNS blip, network timeout, etc.), the exception now bubbles all the way out of `callback()` uncaught, and the whole request 500s instead of completing login. Before this PR, the entire block was wrapped in one try/catch, so this same failure was reported and the user still logged in successfully.

Reproduced with `Http::fake(fn () => throw new ConnectionException('simulated'))` — the callback route returns a 500 instead of redirecting.

Suggested fix — wrap the `Http::...->get(...)` call itself in the try/catch, not just the response-processing branch:

```php
if ($provider === SocialiteProvider::Azure) {
    try {
        // Retry transient Azure failures (connection errors, rate limits, server errors) but not client errors.
        $response = Http::withToken($socialiteUser->token)
            ->dontTruncateExceptions()
            ->retry(3, 500, when: function (?Throwable $exception): bool {
                if ($exception instanceof ConnectionException) {
                    return true;
                }

                return $exception instanceof RequestException
                    && ($exception->response->serverError() || $exception->response->status() === 429);
            }, throw: false)
            ->get('https://graph.microsoft.com/v1.0/me/photo/$value');

        if ($response->successful()) {
            $mimeType = $response->header('Content-Type');

            if (in_array($mimeType, ['image/png', 'image/jpeg', 'image/webp', 'image/jpg', 'image/svg+xml'])) {
                $extension = (new MimeTypes())->getExtensions($mimeType)[0] ?? null;

                $body = $response->body();

                if ($extension && $body) {
                    $media = $user->addMediaFromString($body)->usingFileName(Str::uuid() . '.' . $extension)->toMediaCollection('avatar');

                    if (is_null($media->created_by_id)) {
                        $media->createdBy()->associate($user);
                        $media->saveQuietly();
                    }
                }
            } else {
                throw new InvalidUserAvatarMimeType($mimeType, $user);
            }
        } elseif ($response->failed() && $response->status() !== 404) {
            // A 404 means the user has no Azure profile photo, which is expected and safe to ignore.
            report($response->toException());
        }
    } catch (Throwable $exception) {
        report($exception);
    }
}
```

Worth also adding a regression test similar to the existing ones, faking a thrown `ConnectionException` and asserting the callback still redirects/logs the user in.

mxm1070 (changes requested):
One must-fix issue below — see inline comment.

app-modules/authorization/src/Http/Controllers/SocialiteController.php
Orrison: Good catch, fixed in 4e2e44036. The entire Azure photo block (including the `Http::...->get()` call) is now wrapped in a single try/catch, so an exhausted-retry `ConnectionException` is reported and login still completes. Added a regression test that fakes a thrown `ConnectionException` and asserts the callback still redirects and authenticates the user.

app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -101,37 +101,45 @@ public function callback(SocialiteProvider $provider)
...
-                if (! str_contains($exception->getMessage(), 'Microsoft.Fast.Profile.Core.Exception.ImageNotFoundException')) {
+                } catch (Throwable $exception) {
                     report($exception);
                 }
-            } catch (Throwable $exception) {
-                report($exception);
+            } elseif ($response->status() !== 404) {
+                // A 404 means the user has no Azure profile photo, which is expected and safe to ignore.
+                report($response->toException());
             }
```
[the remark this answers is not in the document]
Orrison: Addressed in 9ce381021 — reporting is now guarded with `$response->failed() && $response->status() !== 404`, so `toException()` is never passed `null` to `report()`. The retry `when` closure was also made null-safe (`?Throwable`).

mxm1070 (approved):

PR #2798, opened 2026-08-31
[ADVAPP-2945]: Introduce the ability to archive students rather than deleting them

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2945

### Technical Description

> Students can now be archived instead of deleted.

> Deleting a student removed them from records they were still attached to - form submissions, engagements, interactions all lost the person's name.

> Archiving keeps that history intact while taking the student out of circulation: they no longer appear in student lists, global search, population groups, selects or reports, and they can no longer sign in to the Resource Hub portal or the customer advisor widget. They still resolve everywhere they are already referenced.

> No

### Cleanup Tasks

> Yes. `.cleanup-tasks/2026_08_31_student_archiving_feature.md`

---


Orrison (changes requested):
Moved this comment to Jira.

app-modules/report/src/Filament/Widgets/StudentCumulativeCountLineChart.php
```diff
@@ -119,32 +130,52 @@ protected function getStudentRunningTotalData(?Carbon $startDate = null, ?Carbon
 
         $months = $this->getMonthRange($startDate, $endDate);
 
-        $monthlyData = Student::query()
+        // Both aggregates run over the same population, so a student can only ever be
+        // subtracted from a bucket if they were added to an earlier one. Archiving a
+        // student created before this window would otherwise take the running total
+        // below the students actually counted in it.
+        $population = fn (): Builder => Student::query()
```
Orrison, line 137: Non-blocking note, and pre-existing (not introduced by this branch) - please create a backlog ticket rather than changing anything here.

The "Students (Cumulative)" chart starts its running total at 0 and only counts students created inside the window. With no date filter it defaults to a trailing 12-month window (`subMonths(11)` to end of this month, in `getStudentRunningTotalData()`), so students created more than 12 months ago are never included, in both the default and date-filtered views. It is effectively a rolling "students added in this window" line rather than a true cumulative population (baseline at window start plus growth).

This branch's archiving math is consistent with that existing behaviour: archives are subtracted only for the same in-window population (this `$population` closure), deliberately, so the total cannot dip below what it counted. Raising it here because the archiving work naturally draws attention to it.

Could you create a backlog ticket to confirm the intended semantics with product? If the chart should show a true running population, the fix would need to: (1) seed the running total with a baseline of students created on or before the window start and not archived as of that point, and (2) widen the archive subtraction beyond in-window creations, so a pre-window student archived mid-window is also subtracted. Both the baseline gap and the archive-subtraction follow-on should be captured in that ticket.

Orrison (commented):
Two follow-up items from a deeper review of the archiving behaviour. One is a should-fix gap, the other is a non-blocking note that mostly needs a backlog ticket. Neither blocks the core intent, which the branch meets well. The non-blocking note is left as an inline comment on `StudentCumulativeCountLineChart`.

### Change request: archived students are not cut off from the Resource Hub portal on ongoing requests

`app-modules/portal/src/Http/Middleware/AuthenticateIfRequiredByPortalDefinition.php` (the `$educatable = $accessToken?->tokenable;` / `abort(Response::HTTP_FORBIDDEN)` block, around lines 59 to 63). This file is not part of this PR's diff, so I could not attach an inline thread to it.

This middleware re-resolves the token's `tokenable` on every portal request but does not check `isArchived()`, so an archived student who already holds a valid `resource-hub-portal` token keeps full portal access until that token expires. The customer advisor equivalent (`CustomerAdvisorAuthorization`) was updated to reject archived students in this PR, and the portal login / refresh / confirm paths were too, but this ongoing per-request gate was missed.

Suggested fix: reject when `$educatable instanceof Student && $educatable->isArchived()`, returning a 403 alongside the existing failure path.

On approach: enforcing this in middleware (rather than firing an event from an archived observer and trying to invalidate sessions) is the right call and worth keeping. `isArchived()` reads `archived_at` fresh from the DB on each request, so it also covers the SIS sync writing `archived_at` directly to the table. That path bypasses Eloquent model events entirely, so an observer/listener approach would not catch SIS-driven archives.

Orrison (approved):

PR #2802, opened 2026-09-01
[ADVAPP-2912]: Improve security by adding a rate-limit and retry limit to login by students and prospects via widget experiences

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2912

### Technical Description

> Improve security by adding a rate-limit and retry limit to login by students and prospects via widget experiences

> No

### Cleanup Tasks

> N/A

---


app-modules/ai/tests/Tenant/Feature/Http/Controllers/CustomerAdvisors/AuthenticationConfirmControllerTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+use function Pest\Laravel\postJson;
+
+/** @phpstan-ignore-next-line */
+beforeEach(function () {
+    /** @phpstan-ignore-next-line */
+    $this->advisor = CustomerAdvisor::factory()->create(['is_embed_enabled' => true]);
+
+    /** @phpstan-ignore-next-line */
+    $this->withHeader('Origin', config('app.url'));
```
Orrison: There should be pest functions (like the `postJson` you imported above) that should allow you to do this without calling `$this`

app-modules/ai/tests/Tenant/Feature/Http/Controllers/CustomerAdvisors/AuthenticationConfirmControllerTest.php
```diff
@@ -0,0 +1,121 @@
...
+use App\Support\AuthenticationCodeRateLimiter;
+use Illuminate\Support\Facades\Hash;
+use Illuminate\Support\Facades\URL;
+
+use function Pest\Laravel\postJson;
+
+/** @phpstan-ignore-next-line */
+beforeEach(function () {
+    /** @phpstan-ignore-next-line */
+    $this->advisor = CustomerAdvisor::factory()->create(['is_embed_enabled' => true]);
```
Orrison: It would probably be better to have a local dataset instead of calling `$this` here

app-modules/ai/tests/Tenant/Feature/Http/Controllers/CustomerAdvisors/RequestAuthenticationControllerTest.php
```diff
@@ -0,0 +1,154 @@
...
+use function Pest\Laravel\postJson;
+
+/** @phpstan-ignore-next-line */
+beforeEach(function () {
+    /** @phpstan-ignore-next-line */
+    $this->advisor = CustomerAdvisor::factory()->create(['is_embed_enabled' => true]);
+
+    /** @phpstan-ignore-next-line */
+    $this->withHeader('Origin', config('app.url'));
+});
```
Orrison: Same comments as the other file

app/Support/AuthenticationCodeRateLimiter.php
```diff
@@ -88,6 +89,15 @@ public function recordCodeRequest(Model $author, string $scope): void
         RateLimiter::hit($this->codeRequestKey($author, $scope), self::CODE_REQUEST_DECAY_SECONDS);
     }
 
+    public function attemptCodeRequest(Model $author, string $scope): void
```
Orrison, line 102: This `attemptCodeRequest` diverges from the primitive the ticket asked us to reuse and from the five sibling widgets (application, form, meeting-center, survey, portal resource-hub), which all call `ensureCanRequestCode()` before minting and `recordCodeRequest()` after. The atomic `Cache::add` is genuinely nicer here — it closes the check-then-hit race the automated reviewer flagged on the two-step pattern — but as written it introduces three issues:

1. A third code-request primitive now sits alongside the two it effectively replaces, used by a single caller, with no comment explaining why the AI flow differs.
2. There is no unit test for `attemptCodeRequest` — `AuthenticationCodeRateLimiterTest` only covers `ensureCanRequestCode`/`recordCodeRequest`.
3. The feature tests are coupled to this implementation: they call `Cache::forget(codeRequestKey(...))` to bypass the cooldown, which only works because we write the key directly via `Cache::add`. Switching to the standard `recordCodeRequest` (`RateLimiter::hit` stores an extra `{key}:timer`) would silently break those tests.

Recommendation: either (a) use the established `ensureCanRequestCode`/`recordCodeRequest` pair for consistency with the ticket, or (b) if we intentionally prefer the atomic approach, unify all six flows onto it in a follow-up and add a dedicated unit test for `attemptCodeRequest`. Minor: the cooldown is recorded before mint/send here, so a failed send still consumes the 60s window (siblings record only after a successful send).

app-modules/ai/src/Http/Controllers/CustomerAdvisors/AuthenticationConfirmController.php
```diff
@@ -54,9 +54,9 @@ public function __invoke(AuthenticationConfirmRequest $request, CustomerAdvisor
             abort(403, 'Authentication code is expired.');
         }
 
-        if (! Hash::check($request->safe()['code'], $authentication->code)) {
-            abort(403, 'Authentication code is invalid.');
-        }
+        $request->validate([
```
Orrison, line 57: `AuthenticationConfirmRequest::rules()` already validates `code` as `['required', 'integer', 'digits:6']`, and this inline `validate()` repeats those same three rules before appending `ValidAuthenticationCode`. On valid input the format rules run twice. Consider moving the `ValidAuthenticationCode` rule into the FormRequest (it can reach the bound record via `$this->route('authentication')`), or dropping the format rules from one of the two places. The sibling controllers avoid this by taking a plain `Request` and doing a single inline `validate()`.

Orrison (changes requested):

app-modules/ai/tests/Tenant/Feature/Http/Controllers/CustomerAdvisors/AuthenticationConfirmControllerTest.php
```diff
@@ -43,17 +43,19 @@
 use Illuminate\Support\Facades\URL;
 
 use function Pest\Laravel\postJson;
+use function Pest\Laravel\withHeader;
 
 /** @phpstan-ignore-next-line */
 beforeEach(function () {
     /** @phpstan-ignore-next-line */
```
Orrison: I don't think these ignores are needed anymore

app-modules/ai/tests/Tenant/Feature/Http/Controllers/CustomerAdvisors/RequestAuthenticationControllerTest.php
```diff
@@ -46,17 +46,19 @@
 use Illuminate\Support\Facades\URL;
 
 use function Pest\Laravel\postJson;
+use function Pest\Laravel\withHeader;
 
 /** @phpstan-ignore-next-line */
 beforeEach(function () {
     /** @phpstan-ignore-next-line */
```
Orrison: I don't think these ignores are needed anymore

Orrison (changes requested):
If you can review anywhere that `@phpstan-ignore-next-line` was added in this increment and remove it if it is no longer needed. Also actually fixing the linting errors in some places added if possible (like providing a `@var` definition of the array structure.

If there are any violations left that specifically NEED to be ignored, then we should never use @phpstan-ignore-next-line, we should use the ignore structure that specifically targets the exact rule failure with a description as to why it must be ignored.

Orrison (approved):

PR #2804, opened 2026-09-01
[ADVAPP-2899]: Apply notification branding consistently across all system emails

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2899

### Technical Description

Ensure that notification settings are used across all outbound emails. Store Brand logo publicly. Remove name and description fields from Notification Settings.

No

### Cleanup Tasks

.cleanup-tasks/2026_09_01_theme_logo_public_disk.md

---

### DIFF
--- a/.cleanup-tasks/2026_09_01_theme_logo_public_disk.md
+++ b/.cleanup-tasks/2026_09_01_theme_logo_public_disk.md
@@ -1,0 +1,14 @@
+---
+title: Theme Logo Public Disk Cleanup
+created: 2026-09-01
+---
+
+## Feature Flags
+
+- App\Features\ThemeLogoPublicDiskFeature
+
+## Temporary Migrations
+
+- database/migrations/2026_09_01_000000_tmp_move_theme_logos_to_public_disk.php
+
+## Additional Cleanup
--- a/app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
+++ b/app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
@@ -108,9 +108,9 @@ class ManageBrandConfigurationSettings extends SettingsPage
                     ->aside()
                     ->schema([
                         SpatieMediaLibraryFileUpload::make('logo')
-                            ->disk('s3')
+                            ->disk('s3-public')
                             ->collection('logo')
-                            ->visibility('private')
+                            ->visibility('public')
                             ->image()
                             ->model(
                                 ThemeSettings::getSettingsPropertyModel('theme.is_logo_active'),
--- a/app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
+++ b/app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
@@ -38,6 +38,7 @@ namespace AdvisingApp\Theme\Filament\Pages;
 
 use AdvisingApp\Theme\Settings\ThemeSettings;
 use App\Enums\NavigationGroup;
+use App\Features\ThemeLogoPublicDiskFeature;
 use App\Models\Tenant;
 use App\Models\User;
 use App\Multitenancy\DataTransferObjects\TenantConfig;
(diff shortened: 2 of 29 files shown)


database/migrations/2026_09_01_000000_tmp_move_theme_logos_to_public_disk.php
```diff
@@ -0,0 +1,80 @@
...
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function (): void {
+            $this->moveThemeLogos(fromDisk: 's3', toDisk: 's3-public');
+
+            ThemeLogoPublicDiskFeature::activate();
+        });
```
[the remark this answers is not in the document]
rebecca-canyon: Because this will only move one file (at most), then if there's an error the feature flag shouldn't be activated.

.cleanup-tasks/2026_09_01_theme_logo_public_disk.md
Orrison, line 1: We delete `name` and `description` in `database/migrations/2026_09_01_171712_drop_name_and_description_from_notification_settings.php` here, but those remain as parameters in `app/Settings/NotificationSettings.php`.

If there is nowhere they are used or mentioned left in the codebase in this increment then I think we should delete them. But if there is a reason that they still need to be around temporarily, then we should add a note to additional cleanup here to remove them.

app-modules/theme/src/Filament/Pages/ManageBrandConfigurationSettings.php
```diff
@@ -108,9 +108,9 @@ public function form(Schema $schema): Schema
                     ->aside()
                     ->schema([
                         SpatieMediaLibraryFileUpload::make('logo')
-                            ->disk('s3')
+                            ->disk('s3-public')
                             ->collection('logo')
-                            ->visibility('private')
+                            ->visibility('public')
```
Orrison: We may need to have these return `s3` and `private` while the feature flag is inactive, as the logo will not have moved yet.

database/migrations/2026_09_01_000000_tmp_move_theme_logos_to_public_disk.php
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        $logo = $settingsProperty->getFirstMedia('logo');
+
+        if (is_null($logo) || $logo->disk !== $fromDisk) {
+            return;
+        }
+
+        $logo->move($settingsProperty, 'logo', $toDisk, $logo->file_name);
```
Orrison, line 75: Just checking, do we know if this will make sure it moves conversions as well? I do think the package allows you to define the disk that the conversions go onto, so you may need to confirm that those get moved as well

resources/views/vendor/filament-panels/components/logo.blade.php
```diff
@@ -44,9 +44,9 @@
 
 @if ($themeSettings->is_logo_active && $logo)
     @php
-        $logoUrl = $logo->getTemporaryUrl(
-            expiration: now()->addMinutes(5),
-            conversionName: $logo->hasGeneratedConversion('logo-height-250px') ? 'logo-height-250px' : '',
+        $logoUrl = $settingsProperty->getFirstMediaUrl(
+            'logo',
+            $logo->hasGeneratedConversion('logo-height-250px') ? 'logo-height-250px' : '',
```
Orrison: My conversion question from the migration stands here, but also, I think we may need to check the feature flag here and return the old way if it is inactive, as I think there will be some time that this fails because the migration moving it to a public disk has not happened yet, so `getFirstMediaUrl` will return a private asset

Orrison (changes requested):

database/migrations/2026_09_01_000000_tmp_move_theme_logos_to_public_disk.php
rebecca-canyon: I've confirmed that the conversions are moved as well

Orrison (approved):

PR #2820, opened 2026-09-03
[ADVAPP-2923]: Improve prompt and prompt type titles by preventing unintentional duplication

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2923

### Technical Description

> RImprove prompt and prompt type titles by preventing unintentional duplication

> No

### Cleanup Tasks

> .cleanup-tasks/2026_09_03_rename_prompts_and_prompt_types_duplicate_title_feature.md

---


danharrin (changes requested):
Do we need to handle creating duplicates (unique validation?) in `SyncTenantSmartPrompts` / `SyncTenantRequest`?

Just checking, is it deliberate that you have conditions for allowing the same title across different prompt types? Is this something that product requested or is there another reason? I thought it was probably going to be global (consistent with the example in the ticket)?

jayushi-canyon:
> is it deliberate that you have conditions for allowing the same title across different prompt types



> Do we need to handle creating duplicates (unique validation?) in `SyncTenantSmartPrompts` / `SyncTenantRequest`?
> 
> Just checking, is it deliberate that you have conditions for allowing the same title across different prompt types? Is this something that product requested or is there another reason? I thought it was probably going to be global (consistent with the example in the ticket)?

@danharrin Actually no it was not asked by product I thought that it would be okay to allow to have same prompts title across different prompt types. If that is contradicting with our approach of having unique title I can make the required changes.

danharrin (changes requested):
@jayushi-canyon please can you ask product in Jira for the correct approach?

danharrin (changes requested):
For the sync smart prompts request, the validation catches duplicate titles within the submitted category, but not conflicts with existing active records. For example, an incoming smart prompt can collide with an existing custom prompt, or with a stale smart prompt that is not deleted until the end of the sync. The database then throws and the endpoint returns a generic 500.

Also, the prompt migration’s `down()` recreates the new (`type_id`, `title`) index rather than restoring the original title-only index. It should restore the pre-migration schema and handle cross-type duplicates created after `up()`.

app/Http/Requests/Tenants/SyncTenantRequest.php
danharrin, line 1: This validation queries tenant models before the tenant has been made current. This request runs on the landlord-api middleware group, and `$tenant->execute()` is not called until the controller action, after FormRequest validation. As a result, a sync containing smart-prompt categories attempts to use an unconfigured tenant connection and throws before reaching the controller.

Could we move the database-backed conflict check into tenant context? Please also ensure the resulting duplicate is returned as a friendly validation error rather than being caught and converted to the controller’s generic 500. This is required by the ticket’s “other creation processes show a friendly validation error” criterion.

app-modules/ai/src/Filament/Resources/Prompts/Pages/CreatePrompt.php
danharrin, line 1: This changes prompt uniqueness from global to per-type before the corresponding migration has necessarily run. Since application code is deployed before per-tenant migrations, this validation can accept the same title in another type while the existing global database index still rejects it.

Per the project’s zero-downtime rules, can we gate the new per-type validation behind a temporary feature flag activated by the migration? Before activation, validation should retain the existing global uniqueness behavior.

danharrin (changes requested):

danharrin (approved):

PR #2825, opened 2026-09-07
[ADVAPP-2957]: Transition tabbing in the holistic profile to the tabbing approach

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2957

### Technical Description

> Transition tabbing in the holistic profile to the tabbing approach

> NO

### Cleanup Tasks

> N/A

---


app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,79 @@
...
+            ->columnSpanFull()
+            ->tabs(
+                $managers
+                    ->map(fn (string $manager, string $key): Tab => Tab::make($manager::getTitle($record, $pageClass))
+                        ->schema([
+                            Livewire::make($manager, [
+                                'ownerRecord' => $record,
+                                'pageClass' => $pageClass,
+                            ])
+                                ->lazy($key !== $firstKey)
```
payal-canyon: `->lazy(true)` can’t reproduce the previous `'on-load'` behaviour. Since inactive tabs are hidden, the relation managers are now deferred until their tab is opened, causing a loading skeleton on the first tab switch.

Could we pass lazy through the component data instead and remove `->lazy(...)`? Using `'lazy' => $key === $firstKey ? false : 'on-load'` would be a 1:1 match for the previous Blade behaviour and follows the existing `'on-load'` convention in `view-student.blade.php`.

app-modules/prospect/src/Filament/Resources/Prospects/Pages/ViewProspect.php
```diff
@@ -59,4 +66,26 @@ public function profile(Schema $schema): Schema
     {
         return ProspectProfileInfolist::configure($schema->record($this->getRecord()));
     }
+
+    public function relationManagerTabs(Schema $schema): Schema
+    {
+        return $schema->components(array_filter([
+            EducatableRelationManagerTabs::make([
```
payal-canyon: `ViewProspect` now owns the relation-manager lists in PHP, but the tests don’t cover the conditional managers.

Could we add the same permission/feature-gated cases as `ViewStudentTest` for the six relation managers? Without these, an incorrect or missing entry in the prospect list could go unnoticed.

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,79 @@
...
+            return null;
+        }
+
+        $firstKey = $managers->keys()->first();
+
+        return Tabs::make()
+            ->columnSpanFull()
+            ->tabs(
+                $managers
+                    ->map(fn (string $manager, string $key): Tab => Tab::make($manager::getTitle($record, $pageClass))
```
payal-canyon: Could we add `->key($key)` to `Tab::make(...)`? The current key is derived from the manager title, so renaming a title or having two managers with the same title could silently change or collide tab identities.

Using the existing array key would give each tab a stable identity without changing the current rendering behaviour.

payal-canyon: Could we drop the `$manager` type hint from the `map()` closure as well? The `filter()` closure already does this, and keeping `string $manager` here widens the `class-string<RelationManager>` type from the method’s `param` annotation.

This would keep both closures consistent while retaining the accurate string `$key` hint.

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,79 @@
...
+        $managers = Collection::make($managers)
+            ->filter(fn ($manager): bool => $manager::canViewForRecord($record, $pageClass));
+
+        if ($managers->isEmpty()) {
+            return null;
+        }
+
+        $firstKey = $managers->keys()->first();
+
+        return Tabs::make()
```
payal-canyon, line 58: Could we give both `Tabs::make()` instances explicit keys, e.g. `relation-manager-tabs` and `additional-relation-manager-tabs`? Without a label, both currently have no key/ID, which makes the rendered `wire:key` ambiguous and would also prevent features like `persistTab()` from working later.

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,79 @@
...
+use Filament\Schemas\Components\Tabs\Tab;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Collection;
+
+class EducatableRelationManagerTabs
+{
+    /**
+     * @param  array<string, class-string<RelationManager>>  $managers
+     */
+    public static function make(array $managers, Model $record, string $pageClass): ?Tabs
```
payal-canyon: Could we always return `Tabs` and hide it when there are no managers, instead of returning `?Tabs`?

This would remove the need for `array_filter()` at all four call sites and prevent a caller from accidentally passing `null` to `->components()`. It renders the same when there are no visible children.

payal-canyon (changes requested):

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,77 @@
...
+     * @param  array<string, class-string<RelationManager>>  $managers
+     */
+    public static function make(string $label, array $managers, Model $record, string $pageClass): Tabs
+    {
+        $managers = Collection::make($managers)
+            ->filter(fn ($manager): bool => $manager::canViewForRecord($record, $pageClass));
+
+        $firstKey = $managers->keys()->first();
+
+        return Tabs::make($label)
```
payal-canyon: Could we use `Tabs::make()` without a label and set the key explicitly with `->key($key)`? Passing the slug as the label causes it to be exposed as the tablist’s `aria-label`, which isn’t user-friendly.

This would keep the tabs uniquely keyed without adding a developer-facing accessible label.

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
```diff
@@ -0,0 +1,77 @@
...
+            ->columnSpanFull()
+            ->hidden($managers->isEmpty())
+            ->tabs(
+                $managers
+                    ->map(fn (string $manager, string $key): Tab => Tab::make($manager::getTitle($record, $pageClass))
```
payal-canyon: Could we add `->key($key)` to `Tab::make(...)`? The current tab key is derived from the manager title, so renaming a title changes the key, and duplicate titles could cause tabs/panels to behave incorrectly.

Using the stable array key would avoid both issues.

app-modules/student-data-model/src/Filament/Resources/Educatables/Schemas/EducatableRelationManagerTabs.php
payal-canyon: It looks like this one was missed. Could you please check and fix it?

app-modules/prospect/tests/Tenant/Prospect/ViewProspectTest.php
```diff
@@ -252,3 +260,174 @@
         ManageProspectTasks::class,
         ManageProspectCareTeam::class,
     ]);
+
+test('renders the EngagementsRelationManager based on proper access', function () {
```
payal-canyon, line 264: `EngagementsRelationManager::canViewForRecord()` allows access with either `engagement.view-any` or `engagement_response.view-any`, but the this test only covers the first.

Could we mirror the `ViewStudentTest` coverage by revoking `engagement.view-any` and then granting `engagement_response.view-any`? This would pin the OR behaviour.

payal-canyon (changes requested):

payal-canyon (approved):

PR #2837, opened 2026-09-14
[ADVAPP-2973]: Transition form request from a select field to a select table

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2973

### Technical Description

> Transition form request from a select field to a select table

> NO

### Cleanup Tasks

> N/A

---


app-modules/form/src/Filament/Tables/RequestableFormsTable.php
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Table configuration used by the "Request" {@see TableSelect}
+ * on the form submission request action.
+ */
+class RequestableFormsTable
+{
+    public static function configure(Table $table): Table
+    {
+        return $table
+            ->query(fn (): Builder => Form::query()->where('is_authenticated', true))
```
Orrison: Thought this was not a part of the original select, I think we should also add here that this should be `withoutArchived` and make sure we have a test asserting that as well.

app-modules/form/src/Filament/Actions/RequestFormSubmission.php
```diff
@@ -57,27 +58,13 @@ protected function setUp(): void
...
-                            ->all())
-                        ->getOptionLabelUsing(fn (string | int | null $value): ?string => filled($value)
-                            ? Form::query()
-                                ->where('is_authenticated', true)
-                                ->whereKey($value)
-                                ->value('name')
-                            : null)
-                        ->searchable()
+                        ->rules([
+                            Rule::exists(Form::class, 'id')->where('is_authenticated', true),
```
Orrison: Especially given my other comment to expand limiting on the table query I worry that this is super easy to have drift from the table query itself, could we have the `RequestableFormsTable` define a static method returning the query scoped and use that here AND in the `->query(...)` call in `RequestableFormsTable` so we make sure they remain the same?

Orrison (changes requested):

Orrison (approved):

PR #2839, opened 2026-09-14
[ADVAPP-2965]: Some users have reported that files uploaded through forms fail to download when their names contain special characters

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2965

### Technical Description

- Fixed broken download links for form-uploaded files with non-ISO-8859-1 names (e.g. macOS screenshots using U+202F before AM/PM), which previously failed to download.
- Added `attachmentContentDisposition` to App\Models\Media, building an ISO-8859-1-safe Content-Disposition (ASCII fallback + RFC 5987 filename*).
- Routed both upload blocks through a new signed `SubmissionMediaDownloadController` so the rendered link stays clean ASCII and survives the submission HTML sanitizer.
- Applies to forms, applications, and event registrations, which share these upload blocks.
Added regression tests for the U+202F disposition and the download controller/route, including a sanitizer-survival assertion.

No

### Cleanup Tasks

No

---


mxm1070 (changes requested):
The approach here is right and the fix is verified working. Requesting three changes before merge — two test additions and a route rename, none of which changes how the download behaves.

## The signed-route indirection is doing more than it looks like it is

The media library keys objects as `{id}/{file_name}`, so the U+202F is in the object key, not only in the `Content-Disposition` header. The presigned URL therefore carries `%E2%80%AF` in its path, and `HtmlSanitizerConfig` strips any href containing it — the anchor renders as `<a>download</a>` with no link at all. Keeping the raw storage URL and only fixing the disposition would still leave a dead link.

So there are two defects, and this PR fixes both.

## 1. The sanitizer assertion is weaker than it looks

`expect($sanitized)->toContain('href')` passes for any anchor that retains an `href` attribute, including one whose URL the sanitizer rewrote. One correction to the earlier automated comment on this, since it changes the fix: an *emptied* href would fail this assertion, so that case is already covered. The gap is a rewritten-but-permitted URL. Extracting and decoding the attribute closes it:

```php
expect($sanitized)->toMatch('/^<a href="[^"]+">download<\/a>$/');

preg_match('/href="([^"]*)"/', $sanitized, $matches);

// Decode first: the sanitizer re-encodes "=" as "&#61;", so a raw string
// comparison fails even when the URL survives intact.
expect(html_entity_decode($matches[1], ENT_QUOTES))->toBe($url);
```

## 2. Add a negative control for the behaviour above

The mechanism in the opening section is what justifies the indirection, and nothing in the suite currently protects it. Without a test, the controller reads like an avoidable layer and is an easy target for a later simplification.

This asserts the raw storage URL *is* stripped, alongside the existing assertion that the signed route survives:

```php
it('strips the raw storage url that the signed route replaces', function () {
    [, $media] = createSubmissionFieldWithMedia(nonLatin1FileName());

    // Derive the URL from the real object key rather than building one by hand: the media
    // library keys objects as "{id}/{file name}", so the non-ASCII character is in the URL
    // path itself, percent-encoded the way a storage presigner encodes a key.
    Storage::disk('s3')->buildTemporaryUrlsUsing(fn (string $path): string => 'https://s3.test/' . implode(
        '/',
        array_map(rawurlencode(...), explode('/', $path)),
    ));

    $rawUrl = $media->getTemporaryUrl(now()->addMinute());

    expect($rawUrl)->toContain(rawurlencode("\u{202F}"));

    $sanitized = (new HtmlSanitizer(app(HtmlSanitizerConfig::class)))->sanitize(
        '<a href="' . htmlspecialchars($rawUrl, ENT_QUOTES) . '">download</a>',
    );

    // This is the second defect the signed route fixes, and the reason an ASCII-safe
    // content disposition alone is not enough: the sanitizer deletes the link outright.
    expect($sanitized)->not->toContain('href');
});
```

## 3. The route name and path say `form-`, but the endpoint is not forms-only

It serves admissions applications and event registrations as well. The route file shows the tension already — the forms-only routes above sit inside the `Route::prefix('forms')->name('forms.')` group, and this one deliberately sits outside it, but kept the prefix in its name. The path is user-visible too, so an applicant downloading an admissions attachment gets a URL that reads `form-submission-media`.

Suggest dropping the prefix from both:

```php
Route::middleware(['web', 'auth', 'signed'])
    ->get('submission-media/{media}/download', SubmissionMediaDownloadController::class)
    ->name('submission-media.download');
```

That is four files including the tests, and no behaviour change.

To be clear about what should not move: the controller belongs in `app-modules/form/`, since that module already owns `Submissible`, `Submission` and `SubmissibleField` that the other modules build on, and `SubmissionMediaDownloadController` is the right class name. Only the route name and path carry the wrong scope.

Worth doing in this PR rather than later. The signature covers the whole URL, so renaming the path invalidates signed links that are already issued, and those live 24 hours. Right now nothing is issued, because the route is new here and has not shipped. After release the same rename means a day of broken links or a redirect kept indefinitely.

## One question on scope

Commit `711060ea` refactors `ProcessSesS3InboundEmail` in the engagement module — different module, not mentioned in the PR description, and unrelated to download links or filename encoding. Is it meant to be here? No objection either way; it just needs a line in the description if it stays.

Two small things if it does stay:

- The commit message reads as a pure refactor, but `media.name` now drops the file extension. Easy to skim past, so worth stating.
- `storeAttachmentMedia()` takes a `$collection` argument, but the inline branch hardcodes `inline_attachments` while the other branch uses the parameter. Every caller passes `attachments` today, so nothing is broken — dropping the argument or deriving both names from it would avoid a surprise later.

monali4-canyon:
### Answer on scope
------
 It's the ingestion-side half of the same defect this PR ports from AidingApp #1444. The root cause is media whose display name includes the file extension, and `ProcessSesS3InboundEmail` is what was writing that bad name in the first place; the download/display changes only mask the symptom without this fix. I'll add a line to the PR description to make that explicit.

**On the two points:**

1. The commit message is misleading. media.name now drops the file extension (via usingName(pathinfo($fileName, PATHINFO_FILENAME))), so it's a behavior change, not a pure refactor.

2. Fixed — I dropped the $collection argument. Both collection names are now handled inside the helper (inline_attachments for the inline branch, attachments otherwise), so there's no unused/misleading parameter. Lint and the attachment tests still pass.

mxm1070 (approved):

PR #2842, opened 2026-09-15
[ADVAPP-2963]: Improve the tabbing experience in the import export page

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2963

### Technical Description

- add import/export tabs and update related pages for better navigation
- add tests for import/export tab functionality and content rendering

No

### Cleanup Tasks

No

---


Orrison (commented):
I think this looks good, I am just going to request a second opinion from @danharrin about whether this is the current most native or efficient way to setup Filament tabs in this way.

app/Filament/Pages/ImportPage.php
```diff
@@ -85,13 +82,6 @@ public static function canAccess(): bool
         return $user->can('export_hub.view-any');
     }
 
-    public function content(Schema $schema): Schema
```
danharrin, line 88: Yes, there should be an easier way. We should avoid custom Blade views for pages if we can, and stick to using a schema like the page `content()` schema. This avoids maintaining two files for the same page, but also avoids introducing inconsistencies with spacing and custom CSS classes that get introduced by views.

We can add tabs to any schema using the `Tabs` schema component, and the default layout for this component is to have them "contained" like product asks for https://filamentphp.com/docs/4.x/schemas/tabs

danharrin (changes requested):

app-modules/student-data-model/src/Filament/Pages/ManageStudentSyncs.php
```diff
@@ -67,4 +66,15 @@ public static function canAccess(): bool
 
         return parent::canAccess();
     }
+
+    public function content(Schema $schema): Schema
+    {
+        return $schema->components([
+            View::make('filament.components.import-export-tabs')
```
danharrin: Sorry, this isn't what I was suggesting. We shouldn't need a Blade view here. We should be able to use the `Tabs` schema component I linked. Let me know if there is a reason we can't

danharrin (changes requested):

app-modules/student-data-model/src/Filament/Pages/ManageStudentSyncs.php
monali4-canyon: @danharrin 

I thought, native schema Tab toggles content client-side and has no href support, whereas these three tabs navigate between three separate Filament pages (each with its own URL, table, and permission). 

But I found it with different pages so I made changes

danharrin (changes requested):
Since the UI is now suggesting that these tabs are just different pieces of content within the same page, lets consolidate  all of these pages into one, and use the native Tabs component behaviour instead of custom HTML. The page title can follow what we already have as the sub navigation label for this page.

Use the `->livewireProperty('activeTab')` on the `Tabs` component and then define a `public string $activeTab = 'import'` with `#[Url(as: 'tab')]` on the page class so that it is synced into the URL. You can use this query parameter if you need to link to the page with a specific tab active.

monali4-canyon:
@danharrin I have applied changes as per your suggestions. Can you please review again and approve? So that I can apply similar changes in other PRs

danharrin (approved):

PR #2843, opened 2026-09-15
[ADVAPP-2962]: Improve the tabbing experience in the Unified Inbox

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/ADVAPP-2962

### Technical Description

- implement unified inbox tabs in the engagement module

No

### Cleanup Tasks

No

---


danharrin (changes requested):
Lets move to using a `content()` schema on both pages instead of a custom Blade view and use the `Tabs` component.

danharrin (changes requested):
Since the UI is now suggesting that these tabs are just different pieces of content within the same page, lets consolidate all of these pages into one, and use the native Tabs component behaviour instead of custom HTML. The page title can follow what we already have as the sub navigation label for this page.

Use the `->livewireProperty('activeTab')` on the `Tabs` component and then define a `public string $activeTab = 'inbox'` with `#[Url(as: 'tab')]` on the page class so that it is synced into the URL. You can use this query parameter if you need to link to the page with a specific tab active.

danharrin (approved):

